PR #78, opened 2022-03-03
Set msrv as 1.58

### LINKED ISSUE #58: Can't build
```bash
cargo build --release
    Updating crates.io index
error: failed to select a version for the requirement `reqwest = "=0.11.8"`
candidate versions found which didn't match: 0.11.6, 0.11.5, 0.11.4, ...
location searched: crates.io index
```

Dengjianping:
Close #58 

If the version of `rustc` is lower than 1.58, it will give a hint like this:
```
cannot be built because it requires rustc 1.58 or newer, while the currently active rustc version is 1.57.0
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ name = "lapce"
 version = "0.0.9"
 authors = ["Dongdong Zhou <dzhou121@gmail.com>"]
 edition = "2021"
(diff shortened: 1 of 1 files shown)


Michcioperz:
I would also like to see this merged, as I've run into the same problem when trying to package for myself for NixOS stable.

PR #97, opened 2022-03-04
fixed broken build

jorgehermo9:
When I tried to build sources, (`cargo build --release`), the following errors showed up: [screenshot](https://imgur.com/a/Wr0rNG4). 

The problem was using named variables on `format!` macro without naming them on the macro's arguments. The solution was simply specify the variables on macro's args, and use them on the order they're passed in.

This patch fixed the errors for me.

Greetings.

### DIFF
--- a/core/src/palette.rs
+++ b/core/src/palette.rs
@@ -333,7 +333,7 @@ impl PaletteItemContent {
             ),
             PaletteItemContent::SshHost(user, host) => (
                 None,
-                format!("{user}@{host}"),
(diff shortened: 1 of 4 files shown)


jorgehermo9:
I reviewed this, the problem was i didn't upgrade my rustc to 1.58 [update notes here](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html). However, feel free to close pull req if u dont want to support older versions of rustc.

PR #146, opened 2022-03-17
Add icon when file has not been saved

MinusGix:
This displays a circle icon when the file is not saved, making it easier to know what needs to be saved and what doesn't that you have open.  
image
Similar to vscode, if you hover over the tab the dot-icon is replaced with the close-icon.  
  
Note that this does not add any alert when closing a file that has not been saved, which is a feature that should also be added.


dzhou121:
Thank you so much for your PR. Looks good to me.

PR #152, opened 2022-03-17
Add more in-depth keypress from text conversion

MinusGix:
The current keypress handling is somewhat buggy, because it does a manual mapping of special keys to their enumeration version. A good example of this is that binding a key to `End` would have it be read from the keymapping file as `Character("End")` rather than the enum variant, which means that it would not match the keypress by the user.  
This replaces the conversion with the code from https://github.com/pyfisch/keyboard-types/blob/master/src/key.rs#L957, which is the library that Druid uses. However, it is converted to be lowercase to be consistent with the current lowercase handling.  
I also added the custom variants (like `left` being `ArrowLeft`) to the conversion function.  
Of note is that this slightly changes the behavior, so invalid keys are ignored (a warning is logged to the terminal, but it would be good to supply a visible notification to the user) rather than converted to characters (the keymap conversion code comes with a character check). I think this change of behavior is for the best since it makes it easier for the user to notice mistakes in their keymap.


lapce-data/src/keypress.rs
```diff
@@ -855,6 +855,318 @@ impl KeyPressData {
...
+            "eisu" => Eisu,
+            "hankaku" => Hankaku,
+            "hiragana" => Hiragana,
+            "hiraganakatakana" => HiraganaKatakana,
+            "kanamode" => KanaMode,
+            "kanjimode" => KanjiMode,
+            "katakana" => Katakana,
+            "romaji" => Romaji,
+            "zenkaku" => Zenkaku,
+            "zenkakuhankaku" => ZenkakuHankaku,
```
jayy-lmao, line 962: TIL

lapce-data/src/keypress.rs
```diff
@@ -864,26 +1176,18 @@ impl KeyPressData {
...
-                _ => druid::KbKey::Character(parts[parts.len() - 1].to_string()),
+
+            let last_part = parts[parts.len() - 1];
+            let key =
+                KeyPressData::map_str_to_key(last_part.to_lowercase().as_str());
+            let key = if let Some(key) = key {
+                key
+            } else {
+                // Skip past unrecognized key definitions
+                eprintln!("Unrecognized key: {}", last_part);
```
jayy-lmao, line 1187: 👍 
Nice for debugging.
Do you think its worth ticketing up introducing a logging library, or is standard prints probably enough for an IDE?

MinusGix: Optimally, this would be a notification in the corner like VSCode has. That way we'd inform the user that there was an issue, but I don't think Lapce has an existing way to make those so I didn't do that here.    
`tracing` could be added without dependency cost, since we already depend on druid which uses tracing itself. I think it could be useful, especially with settings/environment variables to tell it to log certain things (so then we don't have to wait out the long rebuild times after we add a few prints).   
Overall, I think it could provide some benefit but isn't too major.

dzhou121: You can use ```log::warn!() ```

And change the log level here https://github.com/lapce/lapce/blob/master/lapce-ui/src/app.rs#L19

which will be logged to a file in the config folder.

MinusGix:
I switched the code to use `log::warn!` instead.

dzhou121:
Thanks very much for the PR!

PR #156, opened 2022-03-18
feat: delete to beginning of line and end of line

jayy-lmao:


lapce-data/src/editor.rs
```diff
@@ -3562,10 +3562,23 @@ impl KeyPressFocus for LapceEditorBufferData {
...
+                            &selection,
+                            1,
+                            &Movement::StartOfLine,
+                            Mode::Insert,
+                            true,
+                            self.editor.code_lens,
+                            self.editor.compare.clone(),
+                            &self.config,
+                        )
+                    }
```
jayy-lmao, line 3581: Don't know if this changes any critical behaviour. Might be worth a check by someone who knows what the original code was for

lapce-data/src/editor.rs
```diff
@@ -3784,7 +3797,47 @@ impl KeyPressFocus for LapceEditorBufferData {
                 self.set_cursor_after_change(selection);
                 self.update_completion(ctx);
             }
-            LapceCommand::DeleteForeward => {
+            LapceCommand::DeleteToEndOfLine => {
```
jayy-lmao, line 3800: Allows `d $` to delete until end of line

dzhou121:
I was thinking when you press d or y, you enter a special mode, so that you can reuse the movement key presses. And if you have changed the key binding for movement, that would work.

jayy-lmao:
> I was thinking when you press d or y, you enter a special mode, so that you can reuse the movement key presses. And if you have changed the key binding for movement, that would work.

I don't know if I will have the time in the immediate future to develop a new mode. I've fixed the conflicts for this in case you want it to merge, otherwise I'll come back to it once some of the core mode behavior etc has changed :)

PR #161, opened 2022-03-18
add Github Action to do a build on push

SingingBush:
please consider running a CI build on every push to reduce chances of merging changes that break the build

### DIFF
--- a/.github/workflows/cargo.yml
+++ b/.github/workflows/cargo.yml
@@ -1,22 +1,31 @@
-on: [push]
+on: 
+  push:
+  schedule:
+    - cron: '30 2 5 * *'
 
 name: CI
 
 jobs:
-  build_and_test:
-    name: Rust project
-    runs-on: ubuntu-latest
+  build:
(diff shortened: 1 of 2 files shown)


dzhou121:
I feel like we also need to do a build for Windows and macOS?

SingingBush:
> I feel like we also need to do a build for Windows and macOS?

I totally agree. Do you want me up update this PR or merge this for now and do a follow up?

dzhou121:
Can you please update this PR?

dzhou121:
Oh and no need for --release to make the build faster.

SingingBush:
I've made the required changes so that it now builds on all three platforms. It's also fairly simple to additionally perform the builds with both _stable_ and _nightly_ toolchains. Note that the Ubuntu builds are faster by far so building on all three systems takes about 20 minutes.

I also added a badge to the README so that the build status will be shown on the landing page of the repo once actions are running from your repo.

https://github.com/SingingBush/lapce/actions/runs/2005576765

dzhou121:
Thanks.

PR #174, opened 2022-03-18
Implement ability to load themes from folder

MinusGix:
This adds the ability to load `.toml` themes from a folder (the same folder that `settings.toml` resides in) named `themes`. The name of the theme is the same as its filename but with out the `.toml` extension, note that this means it is case-sensitive on platforms where case-sensitivity matters.  
Example (bad) theme I threw together:  
image

I added a `Themes` structure because that felt more reasonable for a `load_theme` function to be on, and allows adding more later on rather than putting them on `Config`. I also added a type alias `Theme` to make it more clear what the hashmap is in places that use themes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2060,6 +2060,7 @@ dependencies = [
  "structdesc",
  "strum",
  "strum_macros",
+ "thiserror",
  "threadpool",
  "tinyfiledialogs",
  "toml",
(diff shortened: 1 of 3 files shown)


dzhou121:
Can you please rebase

MinusGix:
Note that I changed the behavior of `get_color_unchecked` to work better with cases where user-themes are missing properties (because, likely new properties will be added in the future and panicking on old themes is undesirable) or they're invalid (such as misspelled).  
Though, I do somewhat dislike that this means that `get_color` is only looking at the current theme, while `get_color_unchecked` falls through to the base theme. Should I also change `get_color` to behave in the same way, where it falls through? Or a different name? Or ignore it?

PR #179, opened 2022-03-19
Only save settings strings when focus is lost

bugadani:
While it's extremely cool that the UI updates in real-time as I change the Font Family or other settings, writing a file for each key press seems more wasteful than necessary.

This PR cleans up storing the settings file and only stores settings when the input field loses focus.

A few notes:
 - Realtime UI update could be restored by decoupling settings storage from the internal state.
 - Settings store (and consequently, UI update) can also happen after a small inactivity delay. I am planning to implement this as it is fairly simple.


bugadani:
Rebased

Issue #185, opened 2022-03-19
Confusion between uppercase and lowercase keyboard modifiers


rbran:
Some keymaps don't work out of the box on Arch Linux. Such the `goto_definition` with keymap `ctrl+n`.

To fix this problem I had to add the following config:
```
# Remove the invalid key
[[keymaps]]
command = "-goto_definition"
mode = "n"
key = "Ctrl+n"

[[keymaps]]
command = "goto_definition"
mode = "n"
key = "ctrl+n"
```

I think this happen because lapce compare `Ctrl` and `ctrl` case-sensitive in some place.

bugadani:
Your "Ctrl+n" command is `command = "-goto_definition"` (note the extra -). Is this intentional?

rbran:
@bugadani Yes, the first one remove the uppercase `Ctrl` , with is invalid.

bugadani:
Thanks! So far, I didn't see anything obviously wrong with the code but I'll keep checking. Keypress parts are converted to lowercase strings internally, so I'm not sure if this is the issue. What happens if you unbind and re-bind the same key combination, with capical `C`?

rbran:
It works. The config seems to be case insensitive, and doesn't matter the Ctrl upper or lower case.

But if you only add
```
[[keymaps]]
command = "goto_definition"
mode = "n"
key = "Ctrl+n"
```
The `settings` window will show the config as duplicated.

Issue #116, opened 2022-03-13
Keybinds don't work


CatDevz:
Typing in the document itself works as expected, however no keybinds works. There is no context in which keybinds worked. I tested from both the latest release version and built from source.

SoulDrivenDev:
Perhaps relevant: I loaded lapce for the first time and got Ctrl+shift+P working for the command palette, but then no keyboard shortcuts worked after that.

bugadani:
Hi and thanks for reporting this issue!

There have been a number of changes made to the key bindings. Could you please test if this issue has been resolved or not?

SoulDrivenDev:
Latest release works wonderfully for me :)

bugadani:
Glad to hear it! :)

PR #191, opened 2022-03-19
Keymap parsing and keypress handling improvements

bugadani:
This PR:
 - Adds an empty common keymap file, to allow deduplicating existing default keymaps
 - Cleans up the keymap parsing function to use simpler code and avoid several unnecessary allocations.

These changes came up while trying to figure out the cause of #116 and #185 - I didn't make any progress on that front, but these changes might make the parser code slightly easier to understand. Maybe, hopefully :)

### DIFF
--- a/defaults/keymaps-common.toml
+++ b/defaults/keymaps-common.toml
@@ -1,0 +1,157 @@
+# --------------------------------- General --------------------------------------------
+
+[[keymaps]]
+key = "F1"
+command = "palette.command"
+
+[[keymaps]]
+key = ":"
+command = "palette.command"
+mode = "n"
+
+[[keymaps]]
+key = "space"
+command = "toggle_code_lens"
+mode = "nv"
+
+# --------------------------------- Basic editing ---------------------------------------
+
+[[keymaps]]
+key = "alt+up"
+command = "move_line_up"
+mode = "i"
+
+[[keymaps]]
+key = "alt+down"
+command = "move_line_down"
+mode = "i"
+
+# ------------------------------------ Multi cursor -------------------------------------
+
+[[keymaps]]
+key = "alt+I"
+command = "insert_cursor_end_of_line"
+mode = "i"
+
+# ------------------------------------ ------------ -------------------------------------
+
+[[keymaps]]
+key = "ctrl+m"
+command = "list.select"
+when = "list_focus"
+
+[[keymaps]]
+key = "enter"
+command = "list.select"
+when = "list_focus"
+
+[[keymaps]]
+key = "ctrl+p"
+command = "list.previous"
+when = "list_focus"
+
+[[keymaps]]
+key = "up"
+command = "list.previous"
+when = "list_focus"
+
+[[keymaps]]
+key = "ctrl+n"
+command = "list.next"
+when = "list_focus"
+
+[[keymaps]]
+key = "down"
+command = "list.next"
+when = "list_focus"
+
+[[keymaps]]
+key = "o"
+command = "list.expand"
+when = "list_focus"
+mode = "n"
+
+[[keymaps]]
+key = "/"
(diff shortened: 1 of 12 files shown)


defaults/keymaps-common.toml
```diff
@@ -0,0 +1,157 @@
...
+command = "right"
+
+[[keymaps]]
+key = "left"
+command = "left"
+
+[[keymaps]]
+key = "up"
+command = "up"
+when = "!list_focus"
```
bugadani, line 110: `up` and `down` only had the `when = "!list_focus"` in the macOS keymap. I don't know which version of this is the intended, so I picked the macOS one and added the `when`.

bugadani:
I put this back to draft because I want to continue refactoring to avoid parsing the TOML strings twice.

lapce-ui/src/app.rs
```diff
@@ -26,8 +26,8 @@ pub fn lanuch() {
                 message
             ))
         })
-        .level(log::LevelFilter::Off)
-        .level_for("piet_wgpu", log::LevelFilter::Info);
+        .level(log::LevelFilter::Info)
```
bugadani, line 29: Note to self: Don't forget to revert this change

dzhou121:
Can I suggest to keep a PR small so that it’s easier to merge?

bugadani:
I can split it up and submit it in parts once I'm done. Parts of it are small (like the `Modes`) change, but splitting up into multiple files will naturally stay rather large.

lapce-data/src/keypress/mod.rs
```diff
@@ -370,17 +370,14 @@ impl KeyPressData {
 
         if mode != Mode::Insert
```
bugadani, line 371: Checking `mode` here looks redundant. `handle_count` already checks these conditions, but here the code is checking the original focus, not the potentially changed one.

@dzhou121 if the focus changes, why does this code check the original focus' mode?

lapce-data/src/keypress/mod.rs
```diff
@@ -352,10 +352,8 @@ impl KeyPressData {
             KeymapMatch::None => {
                 self.pending_keypress.clear();
                 if focus.get_mode() == Mode::Insert {
-                    let mut keypress = keypress.clone();
-                    keypress.mods.set(Modifiers::SHIFT, false);
                     if let KeymapMatch::Full(command) =
-                        self.match_keymap(&[keypress], focus)
+                        self.match_keymap(std::slice::from_ref(&keypress), focus)
```
bugadani, line 356: Here I am either
 - introducing a bug
 - or I'm allowing key combinations like `shift+Delete` which I don't think were possible before

bugadani:
I've cleaned up the line ending changes that inflated this PR.

bugadani:
I have created a commit history that is hopefully easy to review. One downside is, that `keypress/mod.rs` still uses CRLF as line endings, to reduce the diff size.

Note that [Refactor keymap loader to avoid deserializing TOMLs twice](https://github.com/lapce/lapce/pull/191/commits/da13814138ca7a8a56228ca931dea60414a03510) may look like it has a weird code structure - impl blocks are generally mixed up. This was done to reduce the diff. The impl blocks are merged in 
[Split up into files](https://github.com/lapce/lapce/pull/191/commits/125794060f140e0ae0ef1941974dd5f3c2ea0dd0)

[Split up into files](https://github.com/lapce/lapce/pull/191/commits/125794060f140e0ae0ef1941974dd5f3c2ea0dd0) is basically a bunch of cut-paste operations, and fixing up the imports. The commit should not contain any code changes other than that, but don't trust me on this.

bugadani:
Rebased

dzhou121:
Thank you so much for the PR.

PR #205, opened 2022-03-21
Avoid heap allocations in `check_condition` and simplify

bugadani:


bugadani:
Sure I can do that

bugadani:
Well that was slightly embarassing. I have added tests and fixed a few bugs with the implementation.

bugadani:
Can we move this forward? If for nothing else, maybe for the added tests? :)

PR #215, opened 2022-03-22
Round widget coordinates in Settings

bugadani:
This fixes a weird rendering issue, as shown in the screen shot below.

Without this patch, depending on the size and position of the window, sometimes I got a settings window like this (worst case):

image


dzhou121:
Can you also take a screenshot after this patch is applied?

bugadani:
image

PR #220, opened 2022-03-22
Fix click-to-toggle panels

### LINKED ISSUE #218: UX: only the source control side panel can be closed, the others can't.

bugadani:
Closes #218

### DIFF
--- a/lapce-data/src/command.rs
+++ b/lapce-data/src/command.rs
@@ -192,26 +192,51 @@ pub enum LapceWorkbenchCommand {
     #[strum(serialize = "show_panel")]
     ShowPanel,
 
-    #[strum(serialize = "toggle_panel")]
-    TogglePanel,
+    /// Toggles the panel passed in parameter.
+    #[strum(serialize = "toggle_panel_focus")]
+    TogglePanelFocus,
 
-    #[strum(serialize = "toggle_terminal")]
-    ToggleTerminal,
+    /// Toggles the panel passed in parameter.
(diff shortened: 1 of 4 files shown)


bugadani:
The special-casing causes a different behaviour than VSCode and makes the implementation slightly more complex. I personally wouldn't keep it, but it was how the SourceControl panel worked before, so I left it in.

dzhou121:
We might be a toggle_panel_visual and toggle_panel_focus

toggle_panel_visual is for mouse clicks, and toggle_panel_focus is for key presses.

dzhou121:
We might be a toggle_panel_visual and toggle_panel_focus

toggle_panel_visual is for mouse clicks, and toggle_panel_focus is for key presses.

bugadani:
What would be the difference in behavior?

dzhou121:
E.g., when the focus is in the editor and the terminal is open, toggle_panel_focus will jump the focus to the terminal, while toggle_panel_visual will close the terminal. 

If the focus is on the terminal, both toggle_panel_focus and toggle_panel_visual will close the terminal panel.

bugadani:
I hope I understood you correctly. Clicks now work as focus commands, but they also fall back to visibility-based toggling for the panels that don't focus.

The other option I had was, to not define `Focus` variants for those panels, but I think this is a bit more simple to extend in the future.

bugadani:
Note that CI isn't green on this one (yet), but github has issues and a run has not been scheduled.

PR #223, opened 2022-03-22
Fix bug in hex_to_color and remove unnecessary allocations

PhilipDaniels:
Adds some unit tests too.


dzhou121:
Thanks for the PR!

PR #224, opened 2022-03-22
Add gruvbox theme

Kyllingene:
It's an imperfect, opinionated theme, but it's better than nothing; adds a `gruvbox-theme.toml` to `defaults` and adds `Lapce Gruvbox` to the theme list.

Note: I originally changed my default-dark, but I reverted the changes to make a PR. Nothing is actually changed in the default theme.


dzhou121:
Thanks for the PR. I'm not sure of this should be a theme plugin. I'll leave it for now until we've got a better idea how to download custom themes.

dzhou121:
Thanks for the PR. I'm not sure of this should be a theme plugin. I'll leave it for now until we've got a better idea how to download custom themes.

Kyllingene:
Ok, cool; it's great to see such active development, keep up the good work!

PR #227, opened 2022-03-23
Fix some places where completion would remain when it should not

MinusGix:
This fixes minor issues where completion would remain when clicking in the editor, and when cutting/pasting text.  
I'm sure more of the commands have this issue, and should cancel/update completion within them, but this PR does it for a few common commands.

### DIFF
--- a/lapce-data/src/editor.rs
+++ b/lapce-data/src/editor.rs
@@ -537,7 +537,7 @@ impl LapceEditorBufferData {
         Ok(())
     }
 
-    fn cancel_completion(&mut self) {
+    pub fn cancel_completion(&mut self) {
         let completion = Arc::make_mut(&mut self.completion);
         completion.cancel();
     }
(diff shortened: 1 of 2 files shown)


lapce-data/src/editor.rs
```diff
@@ -2702,6 +2702,7 @@ impl LapceEditorBufferData {
             mouse_event.mods.shift(),
             mouse_event.mods.alt(),
         ));
+        self.cancel_completion();
```
dzhou121, line 2705: Probably do this on mouse_down

dzhou121, line 2705: Probably do this on mouse_down

lapce-data/src/editor.rs
```diff
@@ -3675,11 +3681,13 @@ impl KeyPressFocus for LapceEditorBufferData {
                     };
                     let data = RegisterData { content: s, mode };
                     self.paste(ctx, &data);
+                    self.update_completion(ctx);
```
dzhou121, line 3684: probably move this into ```self.paste(ctx, &data)```

PR #228, opened 2022-03-23
Implement more general closing of modals

MinusGix:
This adds support for hiding the completion by pressing esc, similar to vscode. I often rely on this since I navigate with the arrow keys, and use esc to quickly hide completion if I decide to move down/up. This is also useful if there is a bug in which the completion menu remains when it should not (such as #227 ).

### DIFF
--- a/defaults/keymaps-common.toml
+++ b/defaults/keymaps-common.toml
@@ -146,13 +146,8 @@ mode = "n"
 
 [[keymaps]]
 key = "esc"
-command = "palette.cancel"
-when = "palette_focus"
+command = "modal.close"
 
-[[keymaps]]
-key = "esc"
-command = "code_actions.cancel"
-when = "code_actions_focus"
 
 [[keymaps]]
 key = "ctrl+b"
--- a/lapce-data/src/code_action.rs
+++ b/lapce-data/src/code_action.rs
@@ -30,7 +30,10 @@ impl KeyPressFocus for CodeActionData {
     }
 
     fn check_condition(&self, condition: &str) -> bool {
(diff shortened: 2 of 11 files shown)


dzhou121:
Probably make it more generic for esc if a 'modal' is active. A modal can be the settings menu, the context menu(right click), completion, command palette, etc.

MinusGix:
I'll rework this PR to do that. Have a `when`-kind for `modal-open` that they all cause to be active.

MinusGix:
Need to implement it for settings, forgot about that.

MinusGix:
This mostly works. I got it to work for completion, code-actions, context menu, and the palette menu. However, I only got it barely working for settings.  
Currently it seems the settings modal doesn't take focus properly. An easy way to illustrate this is to open lapce, click on the editor, then open settings through the settings button. You can still type in the main editor. This means that the settings modal only has focus once you click within one of its text boxes, in which case my esc code works. However, that is obviously not quite what we want. I imagine I'm missing some easy way to make it take focus (I tried adding `ctx.request_focus()` in the handling of `LapceUICommand::ShowSettings`, but that didn't work). So, if you have any advice on how to fix this I'd appreciate it, or it can be merged and improved later.  
  
A second thing is that the change to how `HideMenu` is handled has it return focus to the editor (if there is one), but that's a bit iffy. I'd prefer it to return focus to whatever it was on before hand (so if you had a context menu on the file-explorer pane, it would return focus to the file-explorer), but I was unsure of a good way to do that. Probably isn't too notable of an issue right now, but something to improve in the future.

PR #231, opened 2022-03-23
Implement command + keybind to jump to start/end of document

MinusGix:
This implements the Ctrl+Home and Ctrl+End keybinds and commands which jump to the start/end of the file respectively.


defaults/keymaps-common.toml
```diff
@@ -34,6 +34,14 @@ command = "line_start_non_blank"
 key = "End"
 command = "line_end"
 
+[[keymaps]]
```
dzhou121, line 37: I would follow VSCode for the default keymaps.

For mac, it's ⌘↑ / ⌘↓

MinusGix:
I think this fixes the Mac vs non-mac keybindings?  
(The closing was me accidentally pushing without committing and github apparently autocloses when the branch is equal to master)

lapce-data/src/buffer.rs
```diff
@@ -1459,6 +1459,12 @@ impl Buffer {
                     self.offset_of_line_col(line, col, config.editor.tab_width);
                 (new_offset, horiz)
             }
+            Movement::DocumentStart => {
+                (0, ColPosition::Start)
+            }
+            Movement::DocumentEnd => {
+                (self.len(), ColPosition::End)
```
dzhou121, line 1466: For normal mode, you need the offset to be self.len() -1

dzhou121: Actually it's a bit more complicated than that. For Vim, if the document ends with \n, then it will be self.len(), otherwises, move to previous grapheme boundary of self.len()

MinusGix:
I believe this change makes it work correctly?

PR #233, opened 2022-03-23
Fix crash caused by overflow

bugadani:
This PR fixes the following crash:
image

I don't know why this happened, though, so this might not be the root cause of the issue.

### DIFF
--- a/lapce-ui/src/tab.rs
+++ b/lapce-ui/src/tab.rs
@@ -803,7 +803,10 @@ impl Widget<LapceTabData> for LapceTabNew {
                                                 editor.cursor =
                                                     movement::Cursor::new(
                                                         CursorMode::Normal(
-                                                            buffer.len() - 1,
+                                                            buffer.offset_line_end(
+                                                                buffer.len(),
+                                                                false,
+                                                            ),
                                                         ),
                                                         None,
                                                     );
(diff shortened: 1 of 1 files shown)


MinusGix:
I imagine the buffer was empty (or if it wasn't, it was still loading somehow?) which is why the subtraction would underflow.

bugadani:
>  it was still loading

Maybe, the crash was right on startup

lapce-ui/src/tab.rs
```diff
@@ -733,11 +733,13 @@ impl Widget<LapceTabData> for LapceTabNew {
...
                                         {
                                             let editor = Arc::make_mut(editor);
+                                            let new_position =
+                                                buffer.len().saturating_sub(1);
                                             if data.config.lapce.modal {
                                                 editor.cursor =
                                                     movement::Cursor::new(
                                                         CursorMode::Normal(
-                                                            buffer.len() - 1,
+                                                            new_position,
```
dzhou121, line 742: use this ```buffer.offset_line_end(buffer.len(), false)```

lapce-ui/src/tab.rs
```diff
@@ -746,7 +748,7 @@ impl Widget<LapceTabData> for LapceTabNew {
                                                     movement::Cursor::new(
                                                         CursorMode::Insert(
                                                             Selection::caret(
-                                                                buffer.len() - 1,
+                                                                new_position,
```
dzhou121, line 751: use this ```buffer.offset_line_end(buffer.len(), true)```

bugadani:
I've changed the code to the suggestions and rebased.

PR #246, opened 2022-03-24
Add PHP syntax highlighting

Jellyfrog:
Seems to work fine :)
image

The code lens I don't understand enough to implement proper support for, it replaces the entire file like this atm:
image


dzhou121:
Thanks!

PR #273, opened 2022-03-25
Add C# tree-sitter language

Tiggilyboo:
Created accompanying tree sitter support issue (Encompasses all languages) should anyone else want to tick some of these off.

- Requires tree-sitter-c-sharp PR: https://github.com/tree-sitter/tree-sitter-c-sharp/pull/221 to go upstream before merging this. Currently will fail build unless my fork is used instead (But I would encourage waiting on upstream...)

Let me know if I should have approached this differently. 

Cheers


bugadani:
Hmm, I took a look on this again, what happened to the `C` package?

bugadani:
From the experience we had with these kind of PRs, I think it would be better to add tree-sitter packages in batches. Each one of these causes at least one merge conflict that needs to be manually resolved, which is far from optimal.

bugadani:
Whoops, sorry for the ping, I didn't notice this doesn't compile because of the upstream crate.

ghishadow:
Looks like upstream has merged PR, though this PR might need rebase

bugadani:
Closing because #1104 was merged

PR #275, opened 2022-03-25
Fix and simplify `Movement::update_index`

bugadani:
Some of the new test cases failed with the old code. Namely, `Movement::Up.update_index(0, 1, 1, true);` previously gave `1` which resulted in the crash worked around in https://github.com/lapce/lapce/pull/269

This issue is visible when the language server offers a single code assist. In this case, when the assist list is visible, the up key allows switching selection between the actual code assist, and an invalid item that causes a visual glitch. Selecting this invalid item previously (before #269) crashed Lapce.

A demonstration of the issue:

https://user-images.githubusercontent.com/977627/160213054-e6315242-e04e-47e2-8d71-d7bc1a018db1.mp4

With this PR applied, this behaviour does not occur, and pressing the up/down arrows has no visual effect (because of this, I did not record a video, that would look like a single unchanging frame).

### DIFF
--- a/lapce-data/src/movement.rs
+++ b/lapce-data/src/movement.rs
@@ -897,9 +897,13 @@ impl Movement {
 
 pub fn remove_n_at<T>(v: &mut Vec<T>, index: usize, n: usize) {
     match n.cmp(&1) {
-        Ordering::Equal => {v.remove(index);},
-        Ordering::Greater => {v.splice(index..index + n, std::iter::empty());},
-        _ => ()
+        Ordering::Equal => {
+            v.remove(index);
+        }
+        Ordering::Greater => {
+            v.drain(index..index + n);
+        }
+        _ => (),
     };
 }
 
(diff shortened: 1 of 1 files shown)


lapce-data/src/movement.rs
```diff
@@ -897,9 +897,13 @@ impl Movement {
...
 pub fn remove_n_at<T>(v: &mut Vec<T>, index: usize, n: usize) {
     match n.cmp(&1) {
-        Ordering::Equal => {v.remove(index);},
-        Ordering::Greater => {v.splice(index..index + n, std::iter::empty());},
-        _ => ()
+        Ordering::Equal => {
+            v.remove(index);
+        }
+        Ordering::Greater => {
+            v.drain(index..index + n);
```
bugadani, line 908: While this change is unrelated, rustfmt literally wouldn't let me save the file without making it. I could have chosen not to comit it, but I prefer well-formatted code.
And while I was at it, I replaced `splice` with `drain`, which it uses internally, so this code should be equivalent, but strictly simpler.

bugadani:
@dzhou121 I've added a bit more explanation to this issue and a video demonstrating a visible effect of the bug this patch fixes.

Issue #272, opened 2022-03-25
Tree Sitter Language Support


Tiggilyboo:
Having looked through the majority of the issues, I would like to collectively address the state of the tree sitter language support currently in Lapce.

Here follows current support:

- [ ] [Agda](https://github.com/tree-sitter/tree-sitter-agda)
- [ ] [Bash](https://github.com/tree-sitter/tree-sitter-bash) PR: https://github.com/lapce/lapce/pull/282
- [X] [C](https://github.com/tree-sitter/tree-sitter-c) PR: https://github.com/lapce/lapce/pull/274
- [X] [C++](https://github.com/tree-sitter/tree-sitter-cpp) PR: https://github.com/lapce/lapce/pull/276
- [ ] [C#](https://github.com/tree-sitter/tree-sitter-c-sharp) PR: https://github.com/lapce/lapce/pull/273
- [ ] [CSS](https://github.com/tree-sitter/tree-sitter-css) PR: https://github.com/lapce/lapce/pull/315
- [x] [Elixir](https://github.com/elixir-lang/tree-sitter-elixir) PR: https://github.com/lapce/lapce/pull/317
- [x] [Elm](https://github.com/elm-tooling/tree-sitter-elm) PR: #625 
- [ ] [Fluent](https://github.com/tree-sitter/tree-sitter-fluent)
- [ ] [Glimmer](https://github.com/alexlafroscia/tree-sitter-glimmer)
- [X] [Go](https://github.com/tree-sitter/tree-sitter-go)
- [x] [Haskell](https://github.com/tree-sitter/tree-sitter-haskell) PR: #636
- [ ] [Haxe](https://github.com/vantreeseba/tree-sitter-haxe)
- [ ] [HCL (Terraform)](https://github.com/mitchellh/tree-sitter-hcl)
- [x] [HTML](https://github.com/tree-sitter/tree-sitter-html) PR: #428
- [x] [Java](https://github.com/tree-sitter/tree-sitter-java) PRs: #266 #429
- [X] [Javascript](https://github.com/tree-sitter/tree-sitter-javascript)
- [X] [Typescript](https://github.com/tree-sitter/tree-sitter-typescript)
- [ ] Julia
- [ ] [JSON](https://github.com/tree-sitter/tree-sitter-json) PR: https://github.com/lapce/lapce/pull/280
- [x] [Markdown](https://github.com/MDeiml/tree-sitter-markdown) PR: https://github.com/lapce/lapce/pull/320
[...]

PR #276, opened 2022-03-25
Add Cpp tree-sitter support

Tiggilyboo:
Add tree-sitter Cpp support to cpp, cxx, cc, hpp, hxx file name extensions.

Tree Sitter Support Issue: https://github.com/lapce/lapce/issues/272

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1863,6 +1863,7 @@ dependencies = [
  "thiserror",
  "tree-sitter",
  "tree-sitter-c",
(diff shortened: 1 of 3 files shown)


dzhou121:
Thanks for the PR. Any idea there's no change on ```cargo.lock``` ?

Tiggilyboo:
> Thanks for the PR. Any idea there's no change on `cargo.lock` ?

Missed it in the commit, have corrected this and fast forwarded from master.

PR #282, opened 2022-03-25
Add bash tree-sitter support

Tiggilyboo:
- Requires MR in upstream: https://github.com/tree-sitter/tree-sitter-bash/pull/120

Part of the tree-sitter support issue: https://github.com/lapce/lapce/issues/272


panekj:
Fixed in https://github.com/lapce/lapce/pull/690

PR #284, opened 2022-03-25
Cancel code actions when cancelling completion

### LINKED ISSUE #283: Clicking when code assist list is open does not close the list
When the available code assists list is visible, clicking with the mouse does not close the list. Afterwards, the cursor can be moved with the arrow keys, and new assists are shown in the same area. Escape or clicking no longer closes the list, but moving the scroll position does.

Demonstration:

https://user-images.githubusercontent.com/977627/160213709-d31065c0-a64d-4616-8b3d-2b8b1507101e.mp4

bugadani:
Fixes #283

### DIFF
--- a/lapce-ui/src/code_action.rs
+++ b/lapce-ui/src/code_action.rs
@@ -253,11 +253,18 @@ impl Widget<LapceTabData> for CodeAction {
 
     fn lifecycle(
         &mut self,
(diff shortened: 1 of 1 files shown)


dzhou121:
I think a better approach is to cancel the code actions when focus is lost on code action widget.

bugadani:
Right, that's simpler, too.

PR #293, opened 2022-03-26
fix insert_new_line can add CR chacator

dzhou121:


Cody-Duncan:
I've tested this locally. No matter what I do, I cannot get the CR and LF to split up in ways that broke text rendering like before.
- Save in different editors with a different line ending: CR/LF vs LF
- Add newlines in one, save, and then add text on the newline in the other. Repeat in reverse.

+ In an editor showing line endings, CR/LF stay together as expected
+ text in lapce renders correctly no matter the combination of line endings around the text

Consider the issue with mixed line-endings fixed as far as I can tell with basic testing.

PR #295, opened 2022-03-26
Move some painting related code out of lapce-data

bugadani:
This is nowhere near complete, but I don't want to reorganize the whole world in one go, that would be a synchronization nightmare.

My intent is to separate the responsibilities of the crates a bit better - for example, lapce-data should not contain code related to the UI. This will hopefully help later to decouple structs from one another, where possible.

Another goal is to split up huge files into smaller, more-or-less self-contained modules. My hope is that this helps in reading and understanding the code.

This PR also cleans up an unused parameter.


bugadani:
I need to fix line endings still...

PR #299, opened 2022-03-26
Add support for WSL

### LINKED ISSUE #198: Support for WSL (Windows Subsystem for Linux)
It would be fine to support directly connection to WSL like VSCode and other IDE/editors do, rather than using ssh. You can start from auto detecting or generating ssh configs for WSL hosts.

hasali19:
This PR adds support for remote workspaces using WSL on Windows. This works similarly to the ssh functionality but uses the `wsl` command to interact with the remote instead of ssh.

Closes #198

### DIFF
--- a/lapce-data/Cargo.toml
+++ b/lapce-data/Cargo.toml
@@ -56,6 +56,7 @@ structdesc = { git = "https://github.com/lapce/structdesc" }
 lapce-core = { path = "../lapce-core" }
 lapce-rpc = { path = "../lapce-rpc" }
 lapce-proxy = { path = "../lapce-proxy" }
+bytemuck = "1.8.0"
 
 [build-dependencies]
 cc = "1"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -258,9 +258,9 @@ checksum = "72feb31ffc86498dacdbd0fcebb56138e7177a8cc5cea4516031d15ae85a742e"
 
 [[package]]
 name = "bytemuck"
-version = "1.7.2"
+version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "72957246c41db82b8ef88a5486143830adeb8227ef9837740bdec67724cf2c5b"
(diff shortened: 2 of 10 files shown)


dzhou121:
Thanks for the PR!

@bugadani would you mind testing this?

bugadani:
Sure, can we have some instructions, please? :)

hasali19:
Click the cloud icon in the top left corner and select "Connect to WSL". It should connect, then you can open a folder from WSL.

hasali19:
Btw, it seems there's currently no way to disconnect from a remote session (including SSH) unless I'm missing something? Should I that?

bugadani:
Hmm, works nicely. In fact, work better than local, because the "Open recent workspace" button works in WSL, but not normally :D I am able to connect, I am able to switch to a local workspace, and using the command palette, I'm able to switch back to a previous WSL workspace.

bugadani:
One question: what happens if I have multiple WSL installations?

hasali19:
It will use your default. Might be worth adding a picker but I figured this is a good start.

hasali19:
I've got rid of the hardcoded distro name by parsing the output of `wsl -l -v` to find the default distro. The output format isn't ideal to parse but there doesn't seem to be a better way to find the default distro.

lapce-data/Cargo.toml
```diff
@@ -56,6 +56,7 @@ structdesc = { git = "https://github.com/lapce/structdesc" }
 lapce-core = { path = "../lapce-core" }
 lapce-rpc = { path = "../lapce-rpc" }
 lapce-proxy = { path = "../lapce-proxy" }
+bytemuck = "1.8.0"
```
bugadani, line 59: Could you please mark this as a platform-specific dependency?

bugadani: Eh, nevermind, piet-wgpu already uses bytemuck so it wouldn't do anything positive right now.

dzhou121:
Thanks for the PR!

PR #100, opened 2022-03-06
Close the tab by clicking middle mouse button

SunDoge:
Add support for closing tab by clicking middle mouse button. This is the default behavior of vscode, firefox and chrome. But feel free to reject it, since it's just a small tweak.

PR #300, opened 2022-03-26
Only react to left clicks on editor tabs headers

bugadani:
Clicking/dragging with right or middle mouse button is confusing, and https://github.com/lapce/lapce/pull/100 would use the middle button to close a tab, so let's limit the "normal" interactions to the left mouse button.


dzhou121:
Why does the change has the check_condition in it?

bugadani:
I'm not sure...

PR #305, opened 2022-03-26
Implement hover support

MinusGix:
This implements hover support for language servers.  
Currently this is limited in some ways. (These could be opened as issues)  
 - No markdown support. RA doesn't produce good plaintext output as well. This would not be too challenging to add.
 - Could look better
 - Would be useful to be able to select text inside them
 - Able to show hover with a keybind
 - Hide when mouse moves from the target word in document, similar to vscode

### DIFF
--- a/lapce-data/src/config.rs
+++ b/lapce-data/src/config.rs
@@ -72,6 +72,8 @@ impl LapceTheme {
     pub const COMPLETION_BACKGROUND: &'static str = "completion.background";
     pub const COMPLETION_CURRENT: &'static str = "completion.current";
 
+    pub const HOVER_BACKGROUND: &'static str = "hover.background";
+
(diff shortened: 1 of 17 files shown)


lapce-data/src/config.rs
```diff
@@ -128,12 +130,21 @@ pub struct EditorConfig {
     pub tab_width: usize,
     #[field_names(desc = "If opened editors are shown in a tab")]
     pub show_tab: bool,
+    #[field_names(
+        desc = "How long (in ms) it should take before the hover information appears"
+    )]
+    #[serde(default = "EditorConfig::default_hover_delay")]
```
dzhou121, line 136: I guess this is to save the need to put the field in the default settings file. But I would prefer to remove this and put it in the settings file because that default file is a kind of documentation that lists all the settings.

PR #314, opened 2022-03-28
Don't allocate strings for syntax highlight lookups

bugadani:
This PR tries to clean up syntax highlighting by not creating temporary strings just to look up in a hashmap. Besides that, this PR extracts some Theme related operations from Config.


bugadani:
I absolutely dislike how the fallback works. Normally, we shouldn't allow loading incomplete themes. Otherwise, we might want to merge the missing keys into the loaded theme at load time. I also don't really like the fact that we store all the unused themes, but at least that has basically no effect.

PR #317, opened 2022-03-28
Add elixir language support

VitorTrin:
closes https://github.com/lapce/lapce/issues/316


lapce-core/Cargo.toml
```diff
@@ -17,6 +17,7 @@ tree-sitter-javascript = "0.20.0"
 tree-sitter-typescript = "0.20.0"
 tree-sitter-python = "0.19.1"
 tree-sitter-toml = "0.20.0"
+tree-sitter-elixir = { git = "https://github.com/VitorTrin/tree-sitter-elixir.git"}
```
Tiggilyboo, line 20: Instead of using the fork, can you make a PR upstream at https://github.com/VitorTrin/tree-sitter-elixir with your changes?

Also, a `version` requirement (In this case I think the current release is  in the tree-sitter-elixer block couldn't hurt :) Otherwise we may explode the next time main / master gets pushed...

Cheers

VitorTrin: Done

dzhou121:
Thanks!

zoedsoupe:
treesitter-elixir also support embedded elixir too, didn’t it? for this case those extensions would be ` eex` (extended elixir), `heex` (html + `eex`) and `leex` (deprecated)

PR #318, opened 2022-03-28
Add tree-sitter JSON support

dakata1337:


Tiggilyboo:
I've added this to the parent list: https://github.com/lapce/lapce/issues/272
Thanks!

dzhou121:
Thanks!

Tiggilyboo:
The tree-sitter dependency is 0.19 being added in this. There's an upstream to pop this over to 0.20. For the time being there will be duplicate crate versions.

PR #320, opened 2022-03-29
Add markdown tree-sitter language

Tiggilyboo:
Requires exposing of highlights.scm, PR: https://github.com/MDeiml/tree-sitter-markdown/pull/35


lapce-core/Cargo.toml
```diff
@@ -21,5 +21,6 @@ tree-sitter-php = { git = "https://github.com/tree-sitter/tree-sitter-php.git",
 tree-sitter-c = "0.20.1"
 tree-sitter-cpp = "0.20.0"
 tree-sitter-json = "0.19.0"
+tree-sitter-markdown = { git = "https://github.com/MDeiml/tree-sitter-markdown.git", version = "0.0.1" }
```
Tiggilyboo, line 24: Despite not being a part of the official tree-sitter github, this is also used in neovim's syntax highlighting so I thought it adequate.

MDeiml:
I should probably warn you that my parser is still a bit experimental. But people have tried it with neovim and it seems to mostly work.

PR #326, opened 2022-03-29
Refactor towards testability, step 4: add separate `register` field

bugadani:
This PR adds `register` so commands don't need to access the whole `main_split` field. Based on #297


bugadani:
Closed in favour of #350

PR #208, opened 2022-03-21
Refactor towards testability, step 5: Remove some unnecessary clones

bugadani:
The testability related change in this is passing proxy by reference, instead of a clone of its `Arc`. The rest is just general cleanup.

PR #327, opened 2022-03-29
Refactor towards testability, step 6: make buffer edit operations more generic

bugadani:
Simple edit tests won't ever touch non-local buffers, so let's allow them to pass in no-op operations. Based on #208


bugadani:
Closed in favour of #350

PR #333, opened 2022-03-30
Remove unnecessary field and function

bugadani:


bugadani:
I've dropped the commit that cleans up `notify_special` as it didn't really belong here.

PR #346, opened 2022-04-01
Pass unhandled key events through the default handler

### LINKED ISSUE #344: Out of Screen focus prevent command palette opening
## Issue

Trying to open the command palette does not work unless the focus is on an input field, that is:
- an actual opened file
- The commit message field of the source control menu
- The integrated terminal

In these fields, using the command CTRL+p works as intended. When the focused input field goes out of the screen, the command palette does not work anymore.

My very naive thinking would be that input Keyboard shortcuts handling is only enabled on text fields.

## Steps to reproduce

- Launch Lapce
- Notice opening the command palette works as the focus is on the integrated terminal
- select the source control menu. The focus moves to the commit message field
- select another menu, such as the plugin menu. The commit message field goes out of the screen but the focus is still on it; we cannot write anything anymore
- Opening the command palette does not work anymore

## Software information

- Os: ArchLinux, kernel 5.16.15-arch1-1
- DE: SwayWM, version 1.7, Wayland
- Lapce version: using latest commit 303b7e6acb2df479fce5d1d1598bae2d24664f67, release mode, but confirmed it's also a problem on latest release 0.0.12

bugadani:
Resolves #344

@dzhou121 ~do you have a better idea? This PR is pretty bad, in that it's not the widgets that pass focus back to the main split (I couldn't get that to work, only for FileExplorer). This means, if one of the panels will support any kind of keyboard input in the future, the implementor needs to know about this special casing.~

### DIFF
--- a/lapce-ui/src/tab.rs
+++ b/lapce-ui/src/tab.rs
@@ -23,7 +23,7 @@ use lapce_data::{
     },
     editor::EditorLocationNew,
     hover::HoverStatus,
-    keypress::KeyPressData,
+    keypress::{DefaultKeyPressHandler, KeyPressData},
     movement::{self, CursorMode, Selection},
     palette::PaletteStatus,
     panel::{PanelPosition, PanelResizePosition},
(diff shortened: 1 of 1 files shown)


dzhou121:
I think a better idea is to catch unhandled KeyDown in LapceTab.

bugadani:
Like this?

PR #347, opened 2022-04-01
Don't open Palette when settings is open

bugadani:
This PR:
 - adds an editor kind to each editor view
 - adds the editor kind to check_condition
 - adds a condition to the palette actions to disallow opening behind the settings window
 - removes unused code


bugadani:
I'm not entirely certain what happened to the keymaps, but I think they should be using `\n` now consistently. Maybe.

dzhou121:
I think it's probably easier to check if settings is active in the palette open code.

bugadani:
I would like to avoid special-casing more things. The palette should not assume anything about settings. We might introduce other dialogs in the future that should also disable palette opening, and we might also introduce new palette-style controls as well. This way, we don't need to touch the code except to introduce new editor types when needed.

dzhou121:
What if the focus is not on the settings editor when the settings is open, and the key press was captured by the catch all? 

I think the setting, palette, file picker, context menu, code actions etc should be popups, and only one popup can be active at a time.

PR #350, opened 2022-04-02
Refactor buffer editing to allow testing

bugadani:
This PR splits up Buffer into three parts:
 - `BufferData`, which contains everything strictly necessary for edits and undo/redo.
 - `BufferDecoration` which holds the special sauce provided by Buffer, e.g. syntax highlighting. This is split off so it can be wrapped into a decorator type for the fancy edit operations that e.g. trigger commands.
 - the rest - some of the `Buffer` data the other two groups didn't need (yet).

This PR introduces the `BufferDataListener` trait to observe and control the edit apply process, `EditableBufferData` that exposes the actual low-level edit operations, combined with a listener object, and `BufferEditListener` that is used by `Buffer` to glue things together.

Names are bad and are subject to change.

### DIFF
--- a/lapce-data/src/buffer.rs
+++ b/lapce-data/src/buffer.rs
@@ -14,7 +14,7 @@ use lsp_types::SemanticTokensServerCapabilities;
 use lsp_types::{CodeActionResponse, Position};
 use serde::{Deserialize, Serialize};
 use std::cell::RefCell;
-use std::cmp::{self, Ordering};
+use std::cmp;
 use std::collections::HashMap;
 use std::ops::Range;
 use std::rc::Rc;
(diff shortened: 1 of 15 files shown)


bugadani:
I think I need to add the selection registry to the mock editor. I think that would solve the failing undo test.

bugadani:
Why did CI even run for this? 🤔

bugadani:
Follow up work:
 - porting as many commands as possible
 - adding single keypress edits (typing) as a command
 - tests, tests, tests

bugadani:
I've removed the test framework and the first few tests and submitted them as a separate pr: #380

PR #357, opened 2022-04-03
Remove duplicate svg module

PhilipDaniels:
The svg.rs module in lapce-data and lapce-ui were virtually identical except for one tiny fix that the UI had. Since it is needed in lapce-data, remove the UI one and use the data one everywhere.

Perhaps review by commit.


lapce-data/src/svg.rs
```diff
@@ -122,8 +122,6 @@ pub fn completion_svg(
 
     Some((
         get_svg(&format!("symbol-{}.svg", kind_str))?,
-        config
-            .get_color(&("style.".to_string() + theme_str))
-            .cloned(),
+        config.get_style_color(theme_str).cloned(),
```
bugadani, line 125: Good find 💖 I thought I didn't miss any of these

bugadani (approved):
> Since it is needed in lapce-data

Let's make a note somewhere that this is the case. I would like to have all UI related code moved out of lapce-data.

PR #360, opened 2022-04-03
Create CmdLineArgs.rs

ailyas19776:
Creating new command line options --help and --version using 'getopt' library

### DIFF
--- a/lapce-ui/src/CmdLineArgs.rs
+++ b/lapce-ui/src/CmdLineArgs.rs
@@ -1,0 +1,166 @@
+extern crate getopts;
+use getopts::Options;
+use std::env;
+
+pub fn version(executable: &str){
+   	format!("Usage: {} [options]", executable);
+    println!("Lightning-fast and Powerful Code Editor: Check details at https://lapce.dev");
+}
+
+pub fn help(executable: &str, opts: Options) {
+        let error = format!("Usage: {} [options]", executable);
+        println!("{}", opts.usage(&error));
+        println!("If you are having trouble with Lapce setup, please visit https://docs.lapce.dev");
+        println!("If you wish to contribute, please go to https://github.com/lapce/lapce.git");
+}
+
+
+pub fn start_cmd_line() -> bool {
+    let args: Vec<String> = env::args().collect();//get arguments
+    let exec = args[0].clone();//executable
+    let mut opts = Options::new();
+    opts.optflag("v", "version", "check Lapce version");
+    opts.optflag("h", "help", "help");
+    //additional new options go here
+    let matches = match opts.parse(&args[1..]) {
+        Ok(m) => { m }//accepts from Ok()
+        Err(f) => {panic!("{}. Use --help to list recognized commands", f.to_string()); return false;}//rejects from Err()
+    };
+    if matches.opt_present("h") {
+        help(&exec, opts);
+        return true;
+    }
+    if matches.opt_present("v"){
+        version(&exec);
+        return true;
+    }
+    return true;
+}
+
+
+#[cfg(test)]
+mod tests {
+    // Note this useful idiom: importing names from outer (for mod tests) scope.
+    use super::*;
+
+    #[test]
+    fn test_stub(){}
+    fn test_general() {
+		let mut a = cmdLine::new();
+        let mut b = cmdLine::new();
+        let mut c = cmdLine::new();
+        a=b;
+        b=c;
+        c=a;
+        assert_eq!(a, b);
+        assert_eq!(a,c);
+        assert_eq!(b,c);     
+   }
+    fn test_works_not_version(){
+    	let args: Vec<String> = env::args().collect();//get arguments
+    	let exec = args[0].clone();//executable
+    	//let mut arguments = cmdLine::new();
+    	let mut opts = Options::new();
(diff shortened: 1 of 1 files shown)


lapce-ui/src/CmdLineArgs.rs
```diff
@@ -0,0 +1,166 @@
...
+
+
+#[cfg(test)]
+mod tests {
+    // Note this useful idiom: importing names from outer (for mod tests) scope.
+    use super::*;
+
+    #[test]
+    fn test_stub(){}
+    fn test_general() {
```
bugadani, line 48: The test framework will only run functions marked with the `#[test]` attribute, and here only the first (empty) function is marked as such.

Anyway, what's the purpose of this function? It doesn't seem to test anything useful.

bugadani (commented):
Thanks for your PR! I have a few nits to pick:
 - please use `snake_case` filenames
 - please format your code by running `cargo fmt`

lapce-ui/src/CmdLineArgs.rs
ailyas19776: Added your changed! 

Yreah, you are right, it does not do anything, I was just checking to see if the test cases were actually working, forgot to delete the stub. I deleted it in the new commit!

bugadani: Thanks! I would like to point out that still only one function is a test, the others will not run.

lapce-ui/src/cmd_line_args.rs
```diff
@@ -0,0 +1,181 @@
...
+        let mut opts = Options::new();
+        opts.optflag("v", "version", "check Lapce version");
+        opts.optflag("h", "help", "help");
+        let matches = match opts.parse(&args[1..]) {
+            Ok(m) => m, //accepts from Ok()
+            Err(f) => {
+                if (f.to_string() != "version".to_string) {
+                    panic!("{}. Use --help to list recognized commands", f.to_string());
+                } //rejects from Err()
+                assert!(true)
```
bugadani, line 83: `assert!(true)` does nothing.

lapce-ui/src/cmd_line_args.rs
```diff
@@ -0,0 +1,181 @@
...
+        let args: Vec<String> = env::args().collect(); //get arguments
+        let exec = args[0].clone(); //executable
+                                    //let mut arguments = cmdLine::new();
+        let mut opts = Options::new();
+        opts.optflag("v", "version", "check Lapce version");
+        opts.optflag("h", "help", "help");
+        let matches = match opts.parse(&args[1..]) {
+            Ok(m) => m, //accepts from Ok()
+            Err(f) => {
+                if (f.to_string() != "version".to_string) {
```
bugadani, line 80: I'm fairly certain `"version".to_string` shouldn't compile. I think you can compare a `String` and a `&str` directly, anyway, so `.to_string()` on the literal is most likely not necessary.

lapce-ui/src/cmd_line_args.rs
```diff
@@ -0,0 +1,181 @@
...
+        a = b;
+        b = c;
+        c = a;
+        assert_eq!(a, b);
+        assert_eq!(a, c);
+        assert_eq!(b, c);
+    }
+    fn test_works_not_version() {
+        let args: Vec<String> = env::args().collect(); //get arguments
+        let exec = args[0].clone(); //executable
```
bugadani, line 72:
```suggestion
        let exec = &args[0]; //executable
```

Why the clone?

ailyas19776: Fixed!

lapce-ui/src/cmd_line_args.rs
ailyas19776: Fixed!

lapce-ui/src/cmd_line_args.rs
ailyas19776: Fixed!

lapce-ui/src/cmd_line_args.rs
ailyas19776: Test cases should be passing as well

bugadani: While your tests may not "fail", I think you are not testing what you want to test. In essence, you are not calling your code anywhere, only third party code which isn't very meaningful.

What you might want to do instead is the following:

- extract the argument processing into a function that takes `&[String]` and returns the `matches`, or the error. Something like `fn handle_command_line_arguments(&[String]) -> anyhow::Result<bool> {...}` but I'm not really familiar with `anyhow` so you may need to write slightly different code.
- `start_cmd_line` should probably return an `anyhow::Result<bool>` or similar, where `Ok(false)` could stop program execution (in case either the version, or the help argument is present).
- in your tests, call the `handle_command_line_arguments` function with different inputs
  - one that verifies that the options are not set, if neither are present in the input
  - one that tests that your function returns an error if it can't process the input arguments
  - one test case that verifies the two help cases, and one that verifies the two version cases. In your case this means checking if your function returns `Ok(false)`, for example with `Result::expect`.

I think this would be closer to what you intend to test.

ailyas19776: Working on it!

Issue #359, opened 2022-04-03
`lapce-data` requires code related to icons


bugadani:
cc https://github.com/lapce/lapce/pull/357

Icon related code really should belong in the `ui` crate, in my opinion.

PhilipDaniels:
This comes down to this function, which is only used in one place and is trivial to remove

https://github.com/lapce/lapce/blob/202955c823023b7414fd5ed3249ff39fbc6029bf/lapce-data/src/data.rs#L329-L331

and this one

https://github.com/lapce/lapce/blob/202955c823023b7414fd5ed3249ff39fbc6029bf/lapce-data/src/palette.rs#L228-L235

which is also only used in one place (palette.rs in the UI crate) but might be more work to move out. I will try it and see.

PR #363, opened 2022-04-03
Move svg.rs from the data crate to the UI crate

PhilipDaniels:
Fixes https://github.com/lapce/lapce/issues/359

After moving the two functions mentioned in my comment on that issue, it was just a question  of fixing up all the imports.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1929,7 +1929,6 @@ dependencies = [
  "unicode-segmentation",
  "unicode-width",
  "ureq",
- "usvg",
  "uuid",
  "xi-rope",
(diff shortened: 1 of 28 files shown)


bugadani:
Nice! Could you please check if the usvg dependency is still necessary in lapce-data?

PhilipDaniels:
> Nice! Could you please check if the usvg dependency is still necessary in lapce-data?

Good spot, it can be removed. In fact it can be removed from the UI crate too as it isn't explicitly needed. However, it is a dependency of druid and piet-wgpu.

bugadani (approved):

PR #364, opened 2022-04-03
Adds alternative keymap to normal_mode

sl4m:
This adds an alternative keymap to `normal_mode` when modal editing is enabled.


sl4m:
Fixed a merge conflict

PR #148, opened 2022-03-17
feat: go to definition vim command

jayy-lmao:
This adds "go to definition" VIM command.

Tested locally and it works a treat.

PR #366, opened 2022-04-04
fix: typos

jayy-lmao:
Here are the small typo changes from my original PR:  https://github.com/lapce/lapce/pull/148

### DIFF
--- a/defaults/keymaps-common.toml
+++ b/defaults/keymaps-common.toml
@@ -226,12 +226,12 @@ mode = "n"
 
 [[keymaps]]
 key = "G"
-command = "go_to_line_deault_last"
+command = "go_to_line_default_last"
 mode = "nv"
 
 [[keymaps]]
 key = "g g"
-command = "go_to_line_deault_first"
+command = "go_to_line_default_first"
 mode = "nv"
 
 [[keymaps]]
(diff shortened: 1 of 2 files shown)


bugadani (approved):
I love newlines. With all the sarcasm of my heart 😂 Thanks for the PR, we'll need to figure out how to prevent messing up line endings once more. Especially because it looks like some lines use on, others use a different one, in the same file which I have no idea how it came to be.

PhilipDaniels:
@bugadani I always found the best way was to use a `.gitattributes` file to control line endings in the repo rather than relying on individual user's config. The situation is confused by git trying by default to normalize line endings depending on your OS. Best to turn that off! These days most editors can handle Unix-style line endings so that's what I go for. I used to do this when I shared files between Unix, Windows and Cygwin - .bashrc files don't work unless they are Unix-style line endings!

https://www.philipdaniels.com/gitcheatsheet/examplegitattributes.txt

dzhou121:
@PhilipDaniels Would you like to send a PR in to add that file?

PhilipDaniels:
@dzhou121  I'll look into it. It's been a while since I had to worry about this...not sure what will happen when applied to an existing repo.

bugadani:
Hopefully not much, most of the line ending discrepancy has been fixed recently. This is a regression from that state I think

PR #378, opened 2022-04-05
Improve text measurement

PhilipDaniels:
Make text measurement more efficient and regularise the calls to these methods.

cc @bugadani 

There are still some direct calls to `new_text_layout` - I will tackle them in another PR if this is OK. The diff is a bit confused, you might want to look at `config.rs` by itself.

### DIFF
--- a/lapce-data/src/config.rs
+++ b/lapce-data/src/config.rs
@@ -539,9 +539,15 @@ impl Config {
     }
 
     /// Calculate the width of the character "W" (being the widest character)
-    /// at the specified font size in the editor's current font family.
+    /// in the editor's current font family at the specified font size.
     pub fn char_width(&self, text: &mut PietText, font_size: f64) -> f64 {
         Self::editor_text_size_internal(self.editor.font_family(), font_size, text, "W").width
+    }
+
+    /// Calculate the width of the character "W" (being the widest character)
+    /// in the editor's current font family and current font size.
+    pub fn editor_char_width(&self, text: &mut PietText) -> f64 {
+        Self::editor_text_size_internal(self.editor.font_family(), self.editor.font_size as f64, text, "W").width
     }
 
     /// Calculate the width of `text_to_measure` in the editor's current font family and font size.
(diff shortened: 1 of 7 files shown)


lapce-data/src/config.rs
```diff
@@ -538,30 +538,91 @@ impl Config {
...
+        // `new_text_layout` can also work by taking ownership of an owned type such
+        // as String, Rc, or Arc. But they all require allocation. We want to measure
+        // `strs` with arbitrary lifetimes. If we 'cheat' by extending the lifetime of the
+        // `text_to_measure` (in this function only) then we can safely call `new_text_layout`
+        // because the `text_layout` value that is produced is local to this function and hence
+        // always dropped inside this function, and hence its lifetime is always stricly less
+        // than the lifetime of `text_to_measure`, irrespective of whether `text_to_measure`
+        // is actually static or not.
+        let static_str: &'static str =
+            unsafe { std::mem::transmute(text_to_measure) };
```
bugadani, line 618: We are also relying on implementation details here. Namely, that `new_text_layout` doesn't store our not-really-static string reference in a static/global/shared state. Which technically might change at any time in the future, so this transmute is really **very** unsafe and it would be a lot better if `new_text_layout` could just take any `&str` instead.

lapce-data/src/config.rs
```diff
@@ -538,30 +538,91 @@ impl Config {
...
-        let text_layout = text
-            .new_text_layout(c.to_string())
-            .font(self.editor.font_family(), self.editor.font_size as f64)
-            .build()
-            .unwrap();
-        text_layout.size().width
+    /// Calculate the width of the character "W" (being the widest character)
+    /// in the editor's current font family and current font size.
+    pub fn editor_char_width(&self, text: &mut PietText) -> f64 {
+        Self::editor_text_size_internal(
```
bugadani, line 556: Subjective: This call really is `char_width(text, self.editor.font_size as f64)`, or `editor_text_width`.

lapce-data/src/config.rs
```diff
@@ -538,30 +538,91 @@ impl Config {
...
+        .width
+    }
+
+    /// Calculate the width of `text_to_measure` in the editor's current font family and font size.
+    pub fn editor_text_width(
+        &self,
+        text: &mut PietText,
+        text_to_measure: &str,
+    ) -> f64 {
+        Self::editor_text_size_internal(
```
bugadani, line 571: Subjective: This could call `editor_text_size` directly instead of duplicating the font size arg.

lapce-data/src/config.rs
```diff
@@ -538,30 +538,91 @@ impl Config {
         self.themes.style_color(name)
     }
 
+    /// Calculate the width of the character "W" (being the widest character)
```
bugadani, line 541: I almost commented on the "W" argument before I saw this.

Since much of the editor rendering relies on fonts being monospaced, is "W" really the widest character? Relying on it certainly prevents us from supporting variable width fonts...

bugadani (approved):
I don't have problems with the implementation except a few stylistical ones (feel free to ignore) and a few phylosophical, which aren't in scope of this PR.

lapce-data/src/config.rs
bugadani: On the other hand, I kinda dislike seeing "W" in two places.

lapce-data/src/config.rs
PhilipDaniels: I think it generally is, but it all depends on the font metrics of course. And yes, this entire functionality will have to be replaced with something else to support variable width fonts, which funnily enough is what kicked off my interest in contributing to Lapce. Another one is we need separate font settings for the editor and the terminal.

lapce-data/src/config.rs
PhilipDaniels: Could replace with a call to `char_width`...I generally dislike stacking calls of wrapper methods but it will cut down the duplication. I'll look at it tomorrow.

bugadani: It's fine either way

PR #379, opened 2022-04-06
Turn typing into an edit command

bugadani:
May be extremely broken.

This PR, besides extracting character insertion, refactors the algorithm to deal with individual selection regions separately, because each region may have different behavior requirements.

### DIFF
--- a/lapce-data/src/editor/commands/insert_chars.rs
+++ b/lapce-data/src/editor/commands/insert_chars.rs
@@ -7,7 +7,7 @@ use crate::{
         get_word_property, matching_char, matching_pair_direction, EditType,
         WordProperty,
     },
-    movement::{Cursor, CursorMode, InsertDrift, Selection},
+    movement::{ColPosition, Cursor, CursorMode, InsertDrift, SelRegion, Selection},
 };
 
 pub struct InsertCharsCommand<'a> {
@@ -22,6 +22,13 @@ impl<'a> InsertCharsCommand<'a> {
         self,
         mut buffer: EditableBufferData<'a, L>,
     ) -> Option<RopeDelta> {
+        fn region_to_selection(region: SelRegion) -> Selection {
+            let mut current_selection = Selection::new();
+            current_selection.add_region(region);
+
+            current_selection
+        }
+
         let Self {
             cursor,
             tab_width,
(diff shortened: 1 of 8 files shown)


bugadani:
One of the new tests are failing because multiple cursors don't correctly inserts matching pairs.

lapce-data/src/editor/commands/insert_chars.rs
```diff
@@ -0,0 +1,291 @@
...
+            .collect::<Vec<_>>();
+
+        let edits_after = edits_after
+            .iter()
+            .map(|(selection, content)| (selection, content.as_str()))
+            .collect::<Vec<_>>();
+
+        buffer.edit_multiple(&edits_after, EditType::InsertChars);
+
+        // Adjust selection according to previous late edits
```
bugadani, line 155: @dzhou121 while this function got more complex, this (L155-175) is the part that was missing previously.

lapce-data/src/editor/commands/insert_chars.rs
```diff
@@ -0,0 +1,291 @@
...
+            .collect::<Vec<_>>();
+
+        let delta = buffer.edit_multiple(&edits, EditType::InsertChars);
+
+        // Update selection
+        let mut selection =
+            selection.apply_delta(&delta, true, InsertDrift::Default);
+
+        // Apply late edits
+        let edits_after = edits_after
```
bugadani, line 138: Note to self: It would be better to allow edits of `SelRegion`s, not just `Selection`s. That way, we wouldn't have to create two vectors.

bugadani:
I'm a bit worried that this implementation might be a bit smarter than necessary. VSCode, if it has multiple cursors, selects a single, consistent edit mode instead of independently deciding for each selection region.

This different behavor can be seen when inserting parentheses, one inside a word and one outside of one. Lapce inserts the matching pair ourside, which might be surprising.

bugadani:
@dzhou121 I'm marking this as ready but I would like to know your take on some of the changes, specifically [this comment](https://github.com/lapce/lapce/pull/379#issuecomment-1093375600) and I think changes made to the selection history and/or the additional wrapper code executed in `edit_with_command` might have some effect on modal editing that I'm not experienced enough to say.

I guess unintended changes will turn up eventually but I would appreciate if you could provide some feedback.

PR #381, opened 2022-04-06
Changed mode indicator color to match theme

dakata1337:


bugadani (commented):
Thanks, this is a nice change! I have a single suggestion to avoid some duplication, which may or may not be worth some consideration.

bugadani (approved):

PR #384, opened 2022-04-06
Remove some unnecessary calls of to_string when calling new_text_layout

PhilipDaniels:
Such calls are not necessary when passing a static string.
I also included a few comments from my previous PR that I intended to do but it was merged before I could get round to it :-)

### DIFF
--- a/lapce-ui/src/settings.rs
+++ b/lapce-ui/src/settings.rs
@@ -357,15 +357,15 @@ impl Widget<LapceTabData> for LapceSettingsPanel {
                     .get_color_unchecked(LapceTheme::EDITOR_BACKGROUND),
             );
 
-            static SETTINGS_SECTIONS: [&str; 3] = ["Core Settings", "Editor Settings", "Keybindings"];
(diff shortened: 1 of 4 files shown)


lapce-ui/src/settings.rs
```diff
@@ -357,13 +357,15 @@ impl Widget<LapceTabData> for LapceSettingsPanel {
...
+            static SETTINGS_SECTIONS: [&str; 3] = ["Core Settings", "Editor Settings", "Keybindings"];
+
+            for (i, text) in SETTINGS_SECTIONS
                 .iter()
                 .enumerate()
             {
                 let text_layout = ctx
                     .text()
-                    .new_text_layout(text.to_string())
+                    .new_text_layout(*text)
```
bugadani, line 368: I might be wrong, but I think calling `into_iter()` on the array (which should be a `const`, really) would allow removing this dereference.

PhilipDaniels: Done.

bugadani (approved):

Issue #383, opened 2022-04-06
File in file explorer should be highlighted on tab change


eddi0815:
Hey,

great project. I'd like the feature that the current file is highlighted in the file explorer. This should be updated on tab change. At the moment it stays on the last file selected from the file explorer.

flukejones:
Looks like this can be closed?

PR #386, opened 2022-04-07
Highlight active file on tab change, pt 1

bugadani:
Part 1 of #383

This PR adds an event that is fired when the user clicka file in file explorer, or switches tabs. This event is handled by the file explorer file list, and sets the active highlighted file's path.

This PR does not open a folder structure if the file is in a closed folder.

### DIFF
--- a/lapce-ui/src/picker.rs
+++ b/lapce-ui/src/picker.rs
@@ -13,16 +13,16 @@ use druid::{
 };
 use lapce_data::{
     command::{LapceUICommand, LAPCE_UI_COMMAND},
-    config::LapceTheme,
+    config::{Config, LapceTheme},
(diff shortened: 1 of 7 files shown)


lapce-ui/src/picker.rs
```diff
@@ -705,6 +705,137 @@ impl Widget<LapceTabData> for FilePickerExplorer {
     }
 }
 
+#[allow(clippy::too_many_arguments)]
+pub fn paint_file_node_item_by_index(
```
bugadani, line 709: What even is FilePicker?

dzhou121: That's the file dialog to open folder when you connect to remote.

PR #392, opened 2022-04-07
Don't panic if terminal does not exist

### LINKED ISSUE #390: Opening a recent workspace in a new window causes panic
Steps to reproduce:
 - open a new window using the command palette
 - in the new window, open a recent workspace

The following message is printed to console:
```thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', lapce-proxy\src\dispatch.rs:345:50```

bugadani:
Fixes #390


bugadani:
It would probably be better if we knew the root cause, but it doesn't seem things are broken with multiple windows...

PR #405, opened 2022-04-09
Add support for more C++ extensions

ccassise:
Lapce should recognize more C++ file extensions, this PR adds those missing extensions.

This is the most "official" list I could find for C++ extensions: https://gcc.gnu.org/onlinedocs/gcc-4.4.1/gcc/Overall-Options.html#index-file-name-suffix-71 and I've basically only added the extensions that are recognized by other popular editors (specifically VSCode).

### DIFF
--- a/lapce-core/src/language.rs
+++ b/lapce-core/src/language.rs
@@ -52,8 +52,8 @@ pub enum LapceLanguage {
 
 impl LapceLanguage {
     pub fn from_path(path: &Path) -> Option<LapceLanguage> {
-        let extension = path.extension()?.to_str()?;
-        Some(match extension {
+        let extension = path.extension()?.to_str()?.to_lowercase();
+        Some(match extension.as_str() {
             "rs" => LapceLanguage::Rust,
             "js" => LapceLanguage::Javascript,
             "jsx" => LapceLanguage::Jsx,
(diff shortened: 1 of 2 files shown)


lapce-core/src/language.rs
```diff
@@ -65,7 +65,8 @@ impl LapceLanguage {
             "php" => LapceLanguage::Php,
             "ex" | "exs" => LapceLanguage::Elixir,
             "c" | "h" => LapceLanguage::C,
-            "cpp" | "cxx" | "cc" | "hpp" | "hxx" => LapceLanguage::Cpp,
+            "cpp" | "cxx" | "cc" | "c++" | "C" | "CPP" | "hpp" | "hxx" | "hh"
```
bugadani, line 68: I think we would want to highlight "C" (and maybe "H", which is missing?) as C files, not C++

bugadani (changes requested):
Thanks for this!

lapce-core/src/language.rs
superlou: Are extensions ever case-sensitive, e.g. should xYz be a different `LapceLanguage` than XyZ?

ccassise: I do agree with @superlou that extensions should not be case sensitive. Looking at VSCode and vim they do not treat extensions as case sensitive. At least, not for language extensions I tested. I've gone ahead and made this change which includes the changes @bugadani suggested.

If you don't want extensions to be case sensitive then I'll do just the original suggested changes.

bugadani (approved):
TIL that in GCC, `C` is case sensitive and means C++. This hurts my brain soooo bad.

Thanks for the changes, I think case-insensitive matching is just fine here.

PR #412, opened 2022-04-11
Refactor indent and outdent line edits

bugadani:


bugadani:
@dzhou121 I hope I don't mess up anything here, because now some additional code runs after in/outdenting lines, that's usually done for other commands (like diagnostics update, and cursor yank / `add_delete` which I still don't understand what it does).

PR #414, opened 2022-04-11
Fix offset computation for multi width characters

nheuillet:
Separation of a fix made in #385 

The offset computation for lines containing multi-width characters (such as \t) was incorrectly done. This PR adds a fix as well as a unit test to show the fix is working.

This PR is important for languages such as Golang, where \t is used instead of space. Without the fix, the Edit position, as well as the diagnostic Position, is incorrectly computed, and:
- For diagnostics, the error wave is shifted to the left of the error
- For Completion / edits, the previous text is overwritten (eg: `fmt.Println` becomes `fPrintn`)


nheuillet:
CC @bugadani : is this enough to separate this fix from the other PR?

bugadani (commented):
I don't think `PluginConfiguration` should be a part of this PR and I would like the merge commits  dropped, if you don't mind.

nheuillet:
Ah, this is a mistake on my master branch.. I will remove it tomorrow, sorry for the trouble!

bugadani:
> sorry for the trouble!

No trouble at all :)

nheuillet:
This good for a review again @bugadani

bugadani (approved):
LGTM, thanks. Should be squash-merged as 5 commits for this is a bit excessive, but we can do that.

dzhou121:
I think we'll need to leave the current ```offset_of_line_col``` as is, because these are to calculate the visual col offset which will be affected by the ```\t``` character.

And you will just simply do a ```offset_of_line(line) + col``` for the position returned by the LSP.

nheuillet:
@dzhou121 I'm not sure I'm following what you are saying sorry. Using the unpatched version of `offset_of_line_col`, I'm having visual errors on the diagnostic waves: the offset is incorrectly calculated with `\t` and the waves are shifted to the left of the error. `offset_of_line(line) + col` might work though, I need to check on that

dzhou121:
What I meant is we still need the original version of ```offset_of_line_col```, which is needed when you move cursors up and down.

bugadani:
Interesting point and I'll need to keep this in mind once I get to testing cursor movement commands. `offset_of_line(line) + col` is a bit more low level than what would be elegant, in my opinion, but my opinion isn't very strong here.

bugadani:
Apparently, this fixes #458 
@dzhou121 are you sure this code is correct as it is?

PR #421, opened 2022-04-14
fix: Check for value instead of unwrap on None (case when terminal is closed)

Liberatys:


bugadani (approved):
Thanks, good catch!

Liberatys:
Always a pleasure to contribute to a cool project :D

Issue #423, opened 2022-04-14
Palette opened on / in Terminal


Liberatys:
When typing "/" inside of the terminal (in modal mode) the terminal is not acting on the command as the mode for palette.line is normal. But the split of the terminal is acting on the / as they have a different mode.

Reproduce:
1. Open the application in modal mode
2. Go into the terminal
3. Type "/"
4. Result => The dialogue opens the palette and the focus is switched.

I'm already looking into a possible solution.

PR #424, opened 2022-04-14
fix: Terminal should swallow interactions

Liberatys:
As the terminal current only has one mode to consider it should handle all interactions that are addressed to it.
Because the terminal is not handling all interactions some commands that should not be called (such as palette.line) are being handled by another dialogue and thus executed even if the mode is not supported.

Fixes: #423

### DIFF
--- a/lapce-ui/src/terminal.rs
+++ b/lapce-ui/src/terminal.rs
@@ -609,6 +609,7 @@ impl Widget<LapceTabData> for LapceTerminal {
                         term_data.receive_char(ctx, &s);
                     }
                 }
+                ctx.set_handled();
                 data.keypress = keypress.clone();
             }
             Event::Command(cmd) if cmd.is(LAPCE_UI_COMMAND) => {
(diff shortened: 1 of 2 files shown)


lapce-ui/src/terminal.rs
```diff
@@ -609,6 +609,7 @@ impl Widget<LapceTabData> for LapceTerminal {
                         term_data.receive_char(ctx, &s);
                     }
                 }
+                ctx.set_handled();
```
Liberatys, line 612: This sits at the end of the "KeyPress" matcher and should be removed once the Terminal has a concept of "Input" and "Normal". For now, the terminal should just eat all the inputs. (Maybe? :D What do you want to do? This seemed like a good solution for the moment)

bugadani (approved):
Thanks, this looks straight forward to me.

lapce-ui/src/terminal.rs
bugadani: The "Terminal" mode feels like a bit of a hack to me, but I'm not familiar with the underlying design decisions.

Issue #387, opened 2022-04-07
There is a command in the Command Palette to display the command palette


PhilipDaniels:
Minor nit, but kinda pointless since you are already in the command palette.

PR #427, opened 2022-04-14
feat: Add exclusion list to command palette to remove command palette open directive

Liberatys:
Fixes: #387

### DIFF
--- a/lapce-data/src/palette.rs
+++ b/lapce-data/src/palette.rs
@@ -761,12 +761,20 @@ impl PaletteViewData {
 
     #[allow(unused_variables)]
     fn get_commands(&mut self, ctx: &mut EventCtx) {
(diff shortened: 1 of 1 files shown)


lapce-data/src/palette.rs
```diff
@@ -761,12 +761,20 @@ impl PaletteViewData {
 
     #[allow(unused_variables)]
     fn get_commands(&mut self, ctx: &mut EventCtx) {
+        let excluded_items: Vec<String> = vec![
```
bugadani, line 764: Could this be a `const` slice of `&'static str` instead? I don't think it's necessary to run through the allocator twice here.

bugadani (changes requested):

bugadani (approved):
Thanks!

PR #430, opened 2022-04-15
[TreeSitter] Add ruby language support

Liberatys:
Add the tree-sitter crate for ruby and integrate into the language setup.

Currently against a fork of 'tree-sitter-ruby' till a pr to bump the pinned tree-sitter version.

### DIFF
--- a/lapce-core/Cargo.toml
+++ b/lapce-core/Cargo.toml
@@ -19,6 +19,7 @@ tree-sitter-python = "0.19.1"
 tree-sitter-toml = "0.20.0"
 tree-sitter-elixir = { git = "https://github.com/elixir-lang/tree-sitter-elixir.git", version = "0.19.0" }
 tree-sitter-php = { git = "https://github.com/tree-sitter/tree-sitter-php.git", version = "0.19.1" }
+tree-sitter-ruby = { git = "https://github.com/Liberatys/tree-sitter-ruby.git", branch = "chore/allow-range-of-tree-sitter" }
 tree-sitter-c = "0.20.1"
 tree-sitter-cpp = "0.20.0"
 tree-sitter-json = "0.19.0"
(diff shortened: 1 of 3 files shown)


lapce-core/Cargo.toml
```diff
@@ -19,6 +19,7 @@ tree-sitter-python = "0.19.1"
 tree-sitter-toml = "0.20.0"
 tree-sitter-elixir = { git = "https://github.com/elixir-lang/tree-sitter-elixir.git", version = "0.19.0" }
 tree-sitter-php = { git = "https://github.com/tree-sitter/tree-sitter-php.git", version = "0.19.1" }
+tree-sitter-ruby = { git = "https://github.com/Liberatys/tree-sitter-ruby.git", branch = "chore/allow-range-of-tree-sitter" }
```
Liberatys, line 22: Have to wait for the to merge the pr so that we can use tree-sitter 0.19 < .

bugadani: `tree-sitter = "0.19"` implicitly allows >= 0.19 (as it's a sugar of `^0.19` so I think the PR might work regardless. But this is just the theory, as shown by the java package...

Liberatys: If I understand the version pinning/ranges correctly then "0.19" will allow us to use versions from 0.19 to < 0.20.
Because the version is ranging to the first non zero leftmost digit? So this doesn't work for us, as tree-sitter is currently set to 0.20.6?

bugadani: Oh damn you are right, never mind me then :)

bugadani: In that case, I'm not against merging your own fork.

Liberatys: Find by me ^^ I'd create a follow-up pr to use the main repo once the pr is merged.

Liberatys: I'd assume that your Java Pr suffers from the same problem ^^

bugadani: Yes, the html crate mislead me on this front...

PR #432, opened 2022-04-16
Fix: remove minimum window size constraint

### LINKED ISSUE #86: Cannot properly tile window to ¼ the monitor size, because minimum window size is too big.
Title says it all. Ideally there would be no explicit minimum window size limit and it’ll squish as small as the window contents allow. Monitor is 1080p on the machine i’m on now, but i also have a 720p Laptop i tile windows to ¼ on during a normal dev session. So if there must be an explicit limit, don’t know how Druid works, then i suggest making it smaller than a quarter of 720p.

Domerpig:
Fixes #86

I have completely removed the minimum window size.


bugadani (approved):
I feel like limiting the window size is unnecessary. There are a few caveats (panel sizes, settings, maybe the hover card won't work nice) but they are limitations that can be lifted later.

bugadani:
@dzhou121 ping

dzhou121:
Thanks for the PR. You'll need to remove the restriction from here as well

https://github.com/lapce/lapce/blob/master/lapce-ui/src/app.rs#L48

Domerpig:
@dzhou121 This is the line I removed in my PR. When I removed it, it was line 46.

bugadani:
There's at least one more: https://github.com/lapce/lapce/blob/0a1eb8f8229977f0c08cc8e19f2f9d2ff9f7bd4c/lapce-ui/src/app.rs#L135

PR #433, opened 2022-04-17
Indentation code refactor

zhaojiangbin:
Reduced code duplication in IndentLine, OutdentLine and InsertTab.  Also eliminated a few str::to_string() calls.


bugadani:
Thanks! I'm not against these changes, but I don't know when I can get to reviewing them properly.

lapce-core/src/indent.rs
```diff
@@ -11,18 +11,19 @@ pub enum IndentStyle {
     Spaces(u8),
 }
 
+pub const DEFAULT_INDENT: IndentStyle = IndentStyle::Spaces(4);
```
bugadani, line 14: This would be better as a Default implementation or a constant of `IndentStyle`

lapce-core/src/indent.rs
```diff
@@ -11,18 +11,19 @@ pub enum IndentStyle {
...
     /// For example, passing `"    "` (four spaces) will create `IndentStyle::Spaces(4)`.
     #[allow(clippy::should_implement_trait)]
     #[inline]
     pub fn from_str(indent: &str) -> Self {
-        // XXX: do we care about validating the input more than this?  Probably not...?
-        debug_assert!(!indent.is_empty() && indent.len() <= 8);
-
         if indent.starts_with(' ') {
-            IndentStyle::Spaces(indent.len() as u8)
+            IndentStyle::Spaces((indent.len() % Self::LONGEST_INDENT.len()) as u8)
```
bugadani, line 26: Why the remainder? I think I would prefer either the previous panicking, or a `min` call, or both.

Also, shouldn't this function be fallible, and check if the `indent` is a string of spaces? Or count the actual number of spaces? I don't know all the usages off the top of my head, but just looking at this function in isolation, the implementation has a lot of assumptions.

lapce-core/src/indent.rs
```diff
@@ -32,21 +33,10 @@ impl IndentStyle {
...
-            IndentStyle::Spaces(1) => " ",
-            IndentStyle::Spaces(2) => "  ",
-            IndentStyle::Spaces(3) => "   ",
-            IndentStyle::Spaces(4) => "    ",
-            IndentStyle::Spaces(5) => "     ",
-            IndentStyle::Spaces(6) => "      ",
-            IndentStyle::Spaces(7) => "       ",
-            IndentStyle::Spaces(8) => "        ",
-
+            IndentStyle::Spaces(x) if x <= Self::LONGEST_INDENT.len() as u8 => Self::LONGEST_INDENT.split_at(x.into()).0,
```
bugadani, line 36: If we are validating the length here, we might not need to do that in `from_str`.

lapce-core/src/indent.rs
```diff
@@ -32,21 +33,10 @@ impl IndentStyle {
...
             // Unsupported indentation style.  This should never happen,
-            // but just in case fall back to two spaces.
-            IndentStyle::Spaces(n) => {
-                debug_assert!(n > 0 && n <= 8); // Always triggers. `debug_panic!()` wanted.
-                "  "
-            }
+            // but just in case fall back to the default of 4 spaces
+            _ => "    "
```
bugadani, line 39: Any reason why this was changed? Panicking in debug builds is a good way to catch unintended behaviour. Also, since this branch is supposed to be unreachable, why change from 2 spaces to 4?

lapce-data/src/editor/commands/indent_line.rs
```diff
@@ -19,62 +16,29 @@ pub struct IndentLineCommand<'a> {
...
-        let edits = edits
-            .iter()
-            .map(|(selection, s)| (selection, s.as_str()))
-            .collect::<Vec<(&Selection, &str)>>();
-
-        let delta = buffer.edit_multiple(&edits, EditType::InsertChars);
-        cursor.apply_delta(&delta);
-
-        Some(delta)
+        Some(indentation::create_multi_edits(
```
bugadani, line 27: I like the strategy here, but this makes reading the code a lot more difficult. For one, the implementation is now syntactically identical to outdenting, which is just confusing. I think this could be remedied by renaming `edit_one_line` to `indent/outdent_one_line`

lapce-data/src/editor/commands/indentation.rs
```diff
@@ -0,0 +1,83 @@
...
+            if line_content == "\n" || line_content == "\r\n" {
+                continue;
+            }
+            let nonblank = buffer.first_non_blank_character_on_line(line);
+            if let Some(edit) = edit_one_line(&buffer, nonblank, indent, tab_width) {
+                edits.push(edit)
+            }
+        }
+    }
+    apply_edits(buffer, cursor, edits)
```
bugadani, line 82: If you call the function `create_`, it probably shouldn't also apply the edit. Especially since it does a lot more than `create_edit` which has a suspiciously similar name.

bugadani (commented):
I have a few preliminary comments. The code definitely looks to be in a better shape than before, but in my opinion there are still room for improvement.

lapce-core/src/indent.rs
zhaojiangbin: Will do.

lapce-core/src/indent.rs
zhaojiangbin: > Why the remainder? I think I would prefer either the previous panicking, or a `min` call, or both.

I thought reminder is better than `debug_assert()` because it will not panic. But if assertion is preferred, I can revert this change.
 
> 
> Also, shouldn't this function be fallible, and check if the `indent` is a string of spaces? Or count the actual number of spaces? I don't know all the usages off the top of my head, but just looking at this function in isolation, the implementation has a lot of assumptions.

The original implementation did not do either of these so I followed suit. Perhaps we can make this function fallible in subsequent changes?

bugadani: I guess my main question was: "why pick remainder over anything else?". It will not panic, but as written, it is not correct and its overflow behaviour is surprising: if I call `from_str` with 8 spaces, the function returns `IndentStyle::Spaces(0)`. 9 spaces -> 1

I think saturating using `indent.len().min(...)` would work better. I think I would like to keep the panic, though, just so we can catch the intended-as-unreachable conditions.

> The original implementation did not do either

We can change a lot here to fix surprising/inconsistent/etc. behavior and we don't have to be quick about it because the code quality and its performance is not really critical. I leave it up to you whether to address things in this PR or later (or address them at all).

zhaojiangbin:
@bugadani I pushed two more commits to address your feedbacks. Is there anything else I need to revise?

bugadani:
No, thanks, this looks good assuming the tests are still green :) Thanks for the changes!

bugadani (approved):

zhaojiangbin:
@bugadani Thanks for your review and feedbacks.

Now a dumb question as I don't have much experience with the github PR workflow. Is my part done with this PR and it just waits for someone to merge? Or do I need to close the PR by clicking the "Close pull request" button?

bugadani:
We'll handle the rest. Or more precisely, @dzhou121 will.

"Close pull request" is more like a "Nevermind, I changed my mind" button, for when you intend to withdraw your request. Instead, we have a big green "Merge pull request" button that actually applies your changes.

PR #434, opened 2022-04-17
Fix clippy warnings

Qwanve:
I was just poking around and noticed some easy looking clippy warnings, so I went ahead and fixed them


bugadani (approved):
Thanks!

Issue #163, opened 2022-03-18
Input issues


VarLad:
On Fedora 35, GNOME, Wayland. AMD iGPU using Mesa. 

Running Lapce v0.0.11 binary with `WGPU_BACKEND=gl ./lapce`
Issues encountered:
~~1. The cursor doesn't move beyond one line up when I hold the Up key.~~
~~Inteterestingly, when I go into modal mode (Normal), and hold j button, it goes all the way down, but when I hold k button, it doesn't go all the way up.~~
~~2. Can't navigate the text in the terminal using the left and right keys.~~
3. It doesn't seem to scroll down (in disabled modal mode), or up, when I try selecting text with my mouse and drag the mouse down (or up)
~~4. Weird numbering in modal editing, normal and visual modes~~
~~5. `del` key doesn't delete forward~~

Abdillah:
> Weird numbering in modal editing, normal and visual modes

That's numbering is the offset to ease up jumping command, btw.

bugadani:
Thanks for reporting! I believe 1) should be fixed now.

bugadani:
1) should be working on current master
2) should be fixed by #437

bugadani:
3 has been fixed by 99b685a7086a11abb3c37e6f6f062078d02edf4a so I think we can close this collection now :)

VarLad:
@bugadani Just downloaded the latest release. 
Unfortunately, 2. still doesn't work

bugadani:
Please open a new one for that. I didn't notice it among the crossed-out lines 😅

PR #437, opened 2022-04-19
Fix cursor movement in non-vi mode terminal

### LINKED ISSUE #195: Arrow keys don't recall command history in terminal on windows
Pressing any of the arrow keys has no effect, the up arrow doesn't recall the last command and the left and right arrows don't move the cursor.

bugadani:
cc #163
Fixes #195

### DIFF
--- a/defaults/keymaps-common.toml
+++ b/defaults/keymaps-common.toml
@@ -29,10 +29,12 @@ mode = "i"
 [[keymaps]]
 key = "Home"
 command = "line_start_non_blank"
+mode = "inv"
 
 [[keymaps]]
 key = "End"
 command = "line_end"
+mode = "inv"
 
 [[keymaps]]
 key = "PageUp"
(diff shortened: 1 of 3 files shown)


dzhou121:
I think we shouldn't catch the keymap and send the keycode in the keydown event. Do you want me to do this?

bugadani:
No, I think I should do this, it should be good practice. I might give up, though :)

But I think it would be weird to handle arrows in the editor, and almost everything except arrows in terminal. Should I add `when = !terminal` (or `mode = <something>`?) to the arrows in the keymaps, so we can handle them like tab/enter/backspace?

dzhou121:
In keymaps file, mode = "inv" would do it.

bugadani:
On that note, do you intend keeping `Terminal` mode? It feels like a bit of a hack to me, since we can have `when` conditions in the keymap. Is there some other reason for the terminal mode?

dzhou121:
That's a good heritage from Neovim. You can go to normal mode and visual mode when you're focusing in Terminal.

bugadani:
> I think we shouldn't catch the keymap and send the keycode in the keydown event.

As far as I can see, alacritty (the app) captures the keys and sends ANSI sequences to the terminal, like [this](https://github.com/alacritty/alacritty/blob/master/alacritty/src/config/bindings.rs#L424) or [with modifiers](https://github.com/alacritty/alacritty/blob/master/alacritty/src/config/bindings.rs#L593).

I'm not sure I can send key codes directly, the best I can do is what `Backspace` does and send ANSI sequences, unless I'm missing something.

bugadani:
So this turned out a bit complex, but hopefully not too complex. I don't know how to "send a character code" with modifiers to the terminal, so I've implemented the most common ansi sequences instead.

PR #441, opened 2022-04-20
Create the fastdev profile

PhilipDaniels:
I noticed that recently the dev builds of lapce are running slower than they used to, perhaps because of the rendering engine change. It is possible to get the best of both worlds with cargo - a good debugging experience with fast builds for your own code, AND release-optimised code for your dependencies. I find this to be nice and speedy and yet still debuggable, this works well when you don't care about stepping into your dependencies.

To demonstrate, I have created a new `fastdev` profile in the root `Cargo.toml`. Alternatively, if nobody cares about stepping into the dependencies for debugging purposes, it is possible to configure the `dev` profile this way.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,8 +21,8 @@ path = "lapce-proxy/src/bin/lapce-proxy.rs"
 [workspace]
 members = ["lapce-ui", "lapce-proxy", "lapce-rpc", "lapce-data", "lapce-core"]
 
-# A profile which compiles all dependencies in release mode but lapce
-# code in dev mode. This gives a good debugging experience for your
+# A profile which compiles all (non-workspace) dependencies in release mode
+# but Lapce code in dev mode. This gives a good debugging experience for your
 # code and fast performance of other people's code. After the initial
 # build subsequent ones are as fast as dev mode builds.
 # See https://doc.rust-lang.org/cargo/reference/profiles.html
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,8 +32,5 @@ members = ["lapce-ui", "lapce-proxy", "lapce-rpc", "lapce-data", "lapce-core"]
 [profile.fastdev.package."*"]
 opt-level = 3
 
-[profile.fastdev.build-override]
-opt-level = 3
-
 [profile.fastdev]
 inherits = "dev"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,3 +20,20 @@ path = "lapce-proxy/src/bin/lapce-proxy.rs"
 
 [workspace]
 members = ["lapce-ui", "lapce-proxy", "lapce-rpc", "lapce-data", "lapce-core"]
+
+# A profile which compiles all dependencies in release mode but lapce
+# code in dev mode. This gives a good debugging experience for your
+# code and fast performance of other people's code. After the initial
+# build subsequent ones are as fast as dev mode builds.
+# See https://doc.rust-lang.org/cargo/reference/profiles.html
+# To use this profile:
+#   cargo build --profile fastdev
+#   cargo run --profile fastdev --bin lapce
+[profile.fastdev.package."*"]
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -20,3 +20,20 @@ path = "lapce-proxy/src/bin/lapce-proxy.rs"
...
+# code and fast performance of other people's code. After the initial
+# build subsequent ones are as fast as dev mode builds.
+# See https://doc.rust-lang.org/cargo/reference/profiles.html
+# To use this profile:
+#   cargo build --profile fastdev
+#   cargo run --profile fastdev --bin lapce
+[profile.fastdev.package."*"]
+opt-level = 3
+
+[profile.fastdev.build-override]
```
bugadani, line 35: According to the docs, this is "build dependencies (build scripts, proc macros, and their dependencies)". What do we expect from this, compile-time improvements?

bugadani (approved):
Interesting idea. I personally didn't really have many problems with the debug mode performance, but I have access to _relatively_ recent hardware. Still, it's an improvement that some might want to use and doesn't interfere otherwise.

Cargo.toml
bugadani: Also, should we set `build-override` for the release profile as well? According to the docs, build dependencies are never optimized.

PhilipDaniels: I'm not sure to be honest, it was new to me. I put it in because Lapce-UI and Data have `[build-dependencies]`, but on closer inspection I'm not sure why - I don't see any build scripts.

bugadani: I don't expect a lot here. The longest build scripts are I believe the 30s per piece compile times for tree-sitter packages, but I don't expect those to improve because of this. I think we should try and compare clean build times with and without this section.

PhilipDaniels: I'll have a look later this evening. It seems that the `[build_dependencies]` sections in UI and data can simply be removed, the project builds without them.

bugadani (changes requested):
I'll withdraw my hasty approval so we can dig a bit deeper into this.

dzhou121:
Can this be optional? I do want to keep everything in debug mode so that I can catch the slow parts and optimise .

bugadani:
If I understand it correctly, if you run Lapce with `cargo run --bin lapce` it will compile with the default `dev` profile just like it does now, so this should be optional as it is.

Cargo.toml
bugadani: I think we might want to emphasize that `[profile.fastdev.package."*"]` is only active for **non-workspace** packages, i.e. lapce-data and co. aren't affected. The docs are clear on this, and you already have "lapce code" in there, but it was a lot less obvious for me than it could have.

PhilipDaniels:
> Can this be optional? I do want to keep everything in debug mode so that I can catch the slow parts and optimise .

As @bugadani says, it's a completely new profile, so the existing `dev` profile works exactly as before. But `dev` is very slow on my machine, I need something that runs faster while being quick to build.

Cargo.toml
PhilipDaniels: > I think we might want to emphasize that `[profile.fastdev.package."*"]` is only active for **non-workspace** packages, i.e. lapce-data and co. aren't affected. The docs are clear on this, and you already have "lapce code" in there, but it was a lot less obvious for me than it could have.

Comment in Cargo.toml updated.

PhilipDaniels: > I'll have a look later this evening. It seems that the `[build_dependencies]` sections in UI and data can simply be removed, the project builds without them.

Removed these sections. They appear to make no difference to compile times, and hence I assume no difference to the generated code.

bugadani (approved):
Thanks for the changes!

bugadani:
@dzhou121 I'd like to ping you on this one, too. New profile is completely optional and incremental compilation should mitigate the effects of a partially optimized build.

PR #444, opened 2022-04-20
Build release versions using LTO

bugadani:
This PR enables link-time optimization for releases built with CI. This gives a ~5% binary size reduction and slightly improved performance at the cost of (on my machine) more than 2x the build time, so I only recommend this for the releases.


PhilipDaniels:
Have you considered settings [codegen-units](https://doc.rust-lang.org/rustc/codegen-options/index.html#codegen-units) to 1?

bugadani:
Not at first, but it also gives a slight bonus, so why not :)

bugadani:
@dzhou121 I'm pinging you once more because this should be a slight, but easy win in app size and performance.

PR #449, opened 2022-04-21
Fix menu positions, allocate less while rendering window

### LINKED ISSUE #445: Window tab UX
Currently, having multiple "window-level" tabs look like this:
image
image

I think this isn't the best we can do here:
 - File picker's drop shadow is displayed on top of the tabs, but it doesn't display when there are no tabs
 - The tab row is below the topmost row (with the remote icon, "Open folder" button, etc), but the contents of that row depend on the active tab
 - Tabs are unnecessarily wide, while editor tabs only take up as much space as they need.

bugadani:
- Title bar now displays the menu at the correct relative space, not assuming y=0
 - Commands are no longer reallocated liberally
 - Menu now uses point of origin in the equality check
 - File explorer drop shadow no longer draws outside of the editor tab area.
 - Some other cleanups

### DIFF
--- a/lapce-ui/src/editor.rs
+++ b/lapce-ui/src/editor.rs
@@ -113,6 +113,8 @@ pub struct LapceEditor {
     mouse_hover_timer: TimerToken,
 
     text_layouts: HashMap<usize, PietTextLayout>,
+
+    right_click_menu_items: Arc<Vec<MenuItem>>,
 }
 
 impl LapceEditor {
@@ -125,6 +127,30 @@ impl LapceEditor {
             mouse_pos: Point::ZERO,
             mouse_hover_timer: TimerToken::INVALID,
             text_layouts: HashMap::new(),
+
+            right_click_menu_items: Arc::new(vec![
+                MenuItem {
+                    text: LapceCommand::GotoDefinition
+                        .get_message()
+                        .unwrap()
+                        .to_string(),
+                    command: LapceCommandNew {
+                        cmd: LapceCommand::GotoDefinition.to_string(),
+                        palette_desc: None,
+                        data: None,
+                        target: CommandTarget::Focus,
+                    },
+                },
+                MenuItem {
+                    text: "Command Palette".to_string(),
+                    command: LapceCommandNew {
+                        cmd: LapceWorkbenchCommand::PaletteCommand.to_string(),
+                        palette_desc: None,
+                        data: None,
+                        target: CommandTarget::Workbench,
+                    },
+                },
+            ]),
         }
     }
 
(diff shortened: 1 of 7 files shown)


bugadani:
There is still quite the room for improvement around paint codes, but this PR should be good enough. One bug I might have caused is tab drag-and-drop doesn't properly finish on mouse button release.

bugadani:
I don't know how this looks on macos, I didn't touch the macos specific painting but I'm pretty sure tabs now overlap the window control buttons?

dzhou121:
The tab bar is below because the title bar position's drag is reserved for dragging the whole window if that makes sense. It's only on macOS right now but I'm planning to do the same on Windows and Linux. 

I commented in the original issue that we'd want to make the tabs consistent. Now looking at the screenshot, I kind of prefer them to be different so that I can easily tell them apart. So maybe let's leave them as is, and hopefully we could have a UX designer one day to solve the style of it :)

bugadani:
> The tab bar is below because the title bar position's drag is reserved for dragging the whole window

So I can do two things here:
 - keep the title bar on top on macos
 - offset the window tabs to be under the title bar
both of these should be easy enough to do.

How do browsers and other tab based apps deal with this reserved area on macos? If we could have our own window borders, and we could place tabs in this area, we would be able to always display the tabs even if only one exists, which would improve discoverability of the feature, if this makes sense.

I would still want to keep the controls of the current "Title" visually under the window tabs. Maybe a good compromise would be to merge it into the footer, or just move below the main area.

I'll revert the width change.

Thanks for the feedback!

bugadani:
I just realized I didn't do a very good job (none at all) with blending the window tabs into the actual tab content 😅 well, that's also a point that I can try.

bugadani:
I've made a few changes just so we can see how it looks.

This variation removes separator lines from the window tabs, and switches foreground/background colors.
I like this better than the first. VS Code also doesn't use separator lines, and we might want to try that as well. At this state, the visual inconsistency between editor and window tabs (drop shadow and borders) are not good, but editor tabs are not the focus of this version.

image

Points of possible improvement:
 - `[+]` button to open a new tab
 - A different color for the title bar and editor tabs
 - Tab separators could probably be lower contrast
 - Some of the spacing could be updated, I didn't bother for this screenshot.
 - Highlight tabs (and panel buttons) on mouse hover with a different background color

bugadani:
Current screenshot:
image

I've changed a lot:
 - new color for window tabs/title bar
 - removed borders
 - added padding to tab separators
 - added drop shadow to editor tabs
 - split and close buttons no longer disappear

It's not perfect (tab separators are rendered under drop shadows), and I want to add drop shadow to separate the editor from inactive tabs. I personally like how this looks, but I'm not attached to it. However, some of the code changes might be useful later (like the new colors, the cleanups, removal of assumptions).

bugadani:
@dzhou121 I've reverted most of the later changes. Here's the latest:
image

There is one open question: how should I fix the macOS layout? Should I offset the tab bar, so the window's title bar is empty, or should I place back the title bar, effectively making this PR nonvisual on macOS?

bugadani:
I've reverted the visual changes, except for one: the split buttons are now rendered always, not just when the mouse is inside 90% of the window area :)

Issue #251, opened 2022-03-24
Delete Key


Im1Bear:
Every time I want to edit some File, I have to press the Delete Key twice, is this intentional?

bugadani:
Hi and thanks for reporting this issue!

Could you please describe your issue in a bit more detail? Are you not able to type text to edit a file? Do you have modal editing enabled by accident?

panekj:
Closing since it's stale

PR #451, opened 2022-04-23
Create key_del.rs

ailyas19776:
Fixing Issue #251 : Delete key.

### DIFF
--- a/lapce-ui/src/key_del.rs
+++ b/lapce-ui/src/key_del.rs
@@ -1,0 +1,89 @@
+use unicode_segmentation::UnicodeSegmentation;//graphemes
+use termion::event::Key;//Key
+use termion::input::TermRead; //StdinLock
+use termion::raw::{RawTerminal};//Screen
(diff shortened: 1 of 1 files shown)


bugadani:
Thanks for the PR! As it is, while we appreciate your efforts, I don't think this PR is ready to be accepted.

I would like to ask you to describe your general approach, and maybe share some information about the issue if you can reproduce it.

My concerns:
 - The PR doesn't hook into any existing code, so how would this fix anything?
 - I don't have enough information about the issue (no clear description, no way to reproduce to ckeck any potential fixes), so I don't know what this fixes, not to mention the how.
 - Reading `stdin` in a GUI application is probably a bit unorthodox.
 - Lapce does not depend on `termion` so the PR most likely wouldn't compile. 
 - Least importantly, the code style is quite out of place. I understand this patch was written probably on Github's UI without access to rustfmt, but it would be nice if you could format it, and fix capitalization.

PR #452, opened 2022-04-23
support parsing the RUST_LOG env var

zhaojiangbin:


lapce-ui/src/logging.rs
```diff
@@ -0,0 +1,87 @@
...
+) -> Dispatch {
+    use LevelFilter::*;
+    match level {
+        "off" => apply(dispatch, module, Off),
+        "error" => apply(dispatch, module, Error),
+        "warn" => apply(dispatch, module, Warn),
+        "info" => apply(dispatch, module, Info),
+        "debug" => apply(dispatch, module, Debug),
+        "trace" => apply(dispatch, module, Trace),
+        val @ _ => {
```
bugadani, line 25: This is an interesting pattern. Why did you chose this over a bare identifier, like `other => ...`?

lapce-ui/src/logging.rs
```diff
@@ -0,0 +1,87 @@
...
+    // RUST_LOG=path::to::module1=trace
+    //
+    // This would set the thresold at Info for all: RUST_LOG=error,info
+    //
+    // This sets the threshold at Error for all modules but 'module1' and
+    // 'module2' which are at Info and Debug, respectively:
+    // RUST_LOG="error,path::to::module1=info,path::to::module2=debug"
+    for module in value.split(',').filter(|s| !s.is_empty()) {
+        let mut iter = module.split('=');
+        if let Some(val) = iter.next() {
```
bugadani, line 56: I think `split_once` could be a better choice here:
 - The implementation could have one fewer branches
 - It would reject `foo=bar=baz` style invalid input

lapce-ui/src/logging.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+    dispatch
+}
+
+pub(super) fn override_log_levels(dispatch: Dispatch) -> Dispatch {
+    match std::env::var("RUST_LOG") {
+        // Not an error if the env var does not exist.
+        Err(std::env::VarError::NotPresent) => dispatch,
+        Err(std::env::VarError::NotUnicode(val)) => {
+            // TODO: throw an error? logger is not configured yet
```
bugadani, line 80: I think a "noisy" ignore is fine here. Logging isn't entirely critical, and printing to console should be enough to notify someone.

If we would want to keep this error in the logs, we could change this function to initialize the logger, and then log this error, but I don't think we need the extra complexity.

bugadani (approved):
Thanks, this looks nice and useful!

lapce-ui/src/logging.rs
zhaojiangbin: > This is an interesting pattern. Why did you chose this over a bare identifier, like `other => ...`?

Because I forgot that a mere `other => ...` will bind anything else. :P

I will fix it.

lapce-ui/src/logging.rs
zhaojiangbin: > I think `split_once` could be a better choice here:
> 
>     * The implementation could have one fewer branches
> 
>     * It would reject `foo=bar=baz` style invalid input

Good point. Replaced it with `splice_once()`.

zhaojiangbin:
> Thanks, this looks nice and useful!
> 
> I've approved the PR but I have a few questions and one place where I think we can improve this, but none of that is really important.

Thanks for the review.

bugadani (approved):
Thanks for the changes!

PR #453, opened 2022-04-24
Implement Markdown support for Hover

### LINKED ISSUE #322: Hover Markdown Support
Currently, the hover feature (that uses LSP hover for its information) does not support Markdown. This is a common way of outputting text, which can provide more visually useful output.  
As well, Rust Analyzer appears to convert the markdown to plaintext in a poor manner, which can make the outputted text look bad.  
Code highlighting could probably be done through our existing treesitter syntax highlighting, and so most of it would be finding a good markdown parser library and parsing the text. Then rendering it appropriately.  
Also note that you should edit the `HoverClientCapabilities` in `lapce-proxy/src/lsp.rs` to specify the content format to be `vec![Markdown, PlainText]`, so that we specify that we prefer markdown (since if it is implemented then we support it!) but also support PlainText as a secondary.

MinusGix:
Closes #322.

This implements rendering markdown in the Hover view, if the LSP supports sending it. Otherwise it continues rendering plaintext.  
image
  
When trying to figure out how to implement it, I realized Druid provides a `TextLayout` (that uses the `WgpuTextLayout`) which provides useful features like getting the size of the text, which would have simplified my initial implementation of hover. This PR uses Druid's text layout rather than directly using Piet's.
The basic idea was primarily based around https://github.com/linebender/druid/blob/master/druid/examples/markdown_preview.rs (though, that uses a bit different code because of the differences between Lapce and mainline Druid).  
I increased the default size of the hover box by 100px, because it felt cramped with headers. Like the comment says, it would be nice if this was configurable.

Current questions:
 - Should link coloration be put as `editor.link`? Something like `hover.link` could work, but it feels like it should be a general color for anything that renders a normal hyperlink.

Current issues (I'll open separate issues for these):
 - Strikethrough support. Lapce's fork of Druid was made before this commit https://github.com/linebender/druid/commit/f2e78abf6e1268b5eafdafa76eb2b8f5f3abb3b3 was put into mainline Druid, so I was unable to add strikethrough support at the moment. This is not too major, but it would be good to support it.
 - Links. This could have been implemented without too much issues, but probably would have required bringing in a link opener crate and so can be done in a separate PR. We might want to support links opening files in the editor, and jumping around in the markdown view (since you can make links to headers) as well.
 - Code highlighting. I think this shouldn't be too hard to implement with `Syntax`.


lapce-data/src/hover.rs
```diff
@@ -152,41 +170,316 @@ impl Default for HoverData {
...
+        base_font_size,
+        link_color,
+    }: HoverTextStyle,
+) {
+    use pulldown_cmark::HeadingLevel;
+    match tag {
+        Tag::Heading(level, _, _) => {
+            // The size calculations are based on the em values given at
+            // https://drafts.csswg.org/css2/#html-stylesheet
+            let font_size = match level {
```
bugadani, line 445: This might be a bit nicer if you returned only the scale factor here, and calculate the font size separately.

bugadani (commented):
Haven't done a proper review, just a quick glance, but so far I really like this. Good job!

lapce-data/src/hover.rs
```diff
@@ -152,41 +170,317 @@ impl Default for HoverData {
...
-            HoverItem::Markdown(x) | HoverItem::PlainText(x) => x.as_str(),
+            HoverItem::Markdown(x) => x.as_str(),
+            HoverItem::PlainText(x) => x.as_str(),
         }
     }
-}
-impl From<MarkedString> for HoverItem {
-    fn from(text: MarkedString) -> Self {
+
+    fn from_marked_string(text: MarkedString, style: HoverTextStyle) -> Self {
```
bugadani, line 221: Could this work with `&HoverTextStyle` instead?

lapce-data/src/hover.rs
```diff
@@ -152,41 +170,317 @@ impl Default for HoverData {
...
+
+    fn links(&self) -> &[druid::text::Link] {
+        self.text.links()
+    }
+}
+
+// TODO: This could moved into a more general utility file, for use in things other than hover
+/// Parse the given markdown into renderable druid rich text
+/// `base_font_size` is the font size of normal text, and the size of headers is computed from it
+fn parse_markdown(text: &str, style: HoverTextStyle) -> MarkdownText {
```
bugadani, line 341: Could this work with `&HoverTextStyle` instead? I don't mind taking it by value, except cloning `HoverTextStyle` for `HoverContents::Array` looks a bit awkward to my eyes.

lapce-data/src/hover.rs
```diff
@@ -152,41 +170,317 @@ impl Default for HoverData {
...
+    }
+
+    fn links(&self) -> &[druid::text::Link] {
+        self.text.links()
+    }
+}
+
+// TODO: This could moved into a more general utility file, for use in things other than hover
+/// Parse the given markdown into renderable druid rich text
+/// `base_font_size` is the font size of normal text, and the size of headers is computed from it
```
bugadani, line 340: Mention of `base_font_size` here seems like an oversight - did you introduce `HoverTextStyle` in a later revision?

lapce-data/src/hover.rs
```diff
@@ -152,41 +170,317 @@ impl Default for HoverData {
...
+    fn env_update(&self, ctx: &druid::text::EnvUpdateCtx) -> bool {
+        self.text.env_update(ctx)
+    }
+
+    fn links(&self) -> &[druid::text::Link] {
+        self.text.links()
+    }
+}
+
+// TODO: This could moved into a more general utility file, for use in things other than hover
```
bugadani, line 338: Maybe most of the markdown-specific code could be moved out of hover. But looking at how flat Lapce's source code is in terms of modules, this might be done at a later stage. However, I think some todos could be added to direct some attention to this.

lapce-data/src/hover.rs
```diff
@@ -152,41 +170,317 @@ impl Default for HoverData {
...
+/// Parse the given markdown into renderable druid rich text
+/// `base_font_size` is the font size of normal text, and the size of headers is computed from it
+fn parse_markdown(text: &str, style: HoverTextStyle) -> MarkdownText {
+    use pulldown_cmark::{Event, Options, Parser};
+
+    let mut builder = RichTextBuilder::new();
+
+    // Our position within the text
+    let mut pos = 0;
```
bugadani, line 347: `pulldown_cmark` has an offset iterator option that provides offsets for the events. I think it could be useful here.

lapce-data/src/hover.rs
```diff
@@ -152,41 +170,317 @@ impl Default for HoverData {
...
+                HeadingLevel::H5 => 0.83,
+                HeadingLevel::H6 => 0.75,
+            };
+            let font_size = font_scale * base_font_size;
+            attrs.size(font_size).weight(FontWeight::BOLD);
+        }
+        Tag::BlockQuote => {
+            attrs.style(FontStyle::Italic).text_color(BLOCKQUOTE_COLOR);
+        }
+        // TODO: We could use the language paired with treesitter highlighting to highlight the code
```
bugadani, line 459:
```suggestion
        // TODO: We could use the language paired with treesitter to highlight the code
```

lapce-data/src/hover.rs
```diff
@@ -152,41 +170,317 @@ impl Default for HoverData {
...
+            }
             MarkupKind::PlainText => HoverItem::PlainText(content.value),
         }
     }
 }
+
+const BLOCKQUOTE_COLOR: Color = Color::grey8(0x88);
+
+#[derive(Data, Clone)]
+pub struct MarkdownText {
```
bugadani, line 251: This struct isn't (or at least, it doesn't look) really markdown-specific. Maybe a name like `StyledText` would be a better choice long term (but I'm really bad at naming, so maybe not).

lapce-ui/src/hover.rs
```diff
@@ -177,75 +179,24 @@ impl Widget<LapceTabData> for HoverContainer {
...
-    /// The final entry
-    Final { height: f64 },
-}
-
 #[derive(Default)]
-pub struct Hover {}
+pub struct Hover {
+    /// The active text layout to be rendered
+    /// Uses [`RichText`] to unify non-markdown and markdown text together
+    active_layout: TextLayout<MarkdownText>,
```
bugadani, line 186: ```Uses [`RichText`] to unify non-markdown and markdown text together``` seems incorrect here, just by looking at these two lines. Could you clarify that you mean this refers to inside `MarkdownText`?

lapce-data/src/hover.rs
```diff
@@ -152,41 +170,317 @@ impl Default for HoverData {
...
+    /// Rendered rich text
+    text: RichText,
+    /// Indices into the underlying text that code blocks start and begin at
+    /// This lets us calculate a rectangle surrounding the text to apply a background too, in order
+    /// to differentiate it from normal text more.
+    /// Note that this covers multiline and inline codeblocks
+    #[data(eq)]
+    code_block_indices: Vec<Range<usize>>,
```
bugadani, line 259: Longer term, maybe this list could also store the type of the block. I.e., whether it's a code block, a block quote, or something else. Could be useful if we end up drawing those with different block styles as well.

lapce-data/src/hover.rs
```diff
@@ -152,41 +170,317 @@ impl Default for HoverData {
...
+    }
+
+    /// Draw the markdown to the paint ctx, using the layout
+    pub fn draw(
+        &self,
+        ctx: &mut PaintCtx,
+        env: &Env,
+        origin: Point,
+        layout: &TextLayout<MarkdownText>,
+        code_background_color: Color,
```
bugadani, line 283: I think we should pass in the `Config` here and pick the background color in this function. We might end up needing it later.

lapce-data/src/hover.rs
```diff
@@ -152,41 +170,317 @@ impl Default for HoverData {
...
+                        style.clone(),
+                    );
+
+                    if should_add_newline_after_tag(&tag) {
+                        builder.push("\n");
+                        pos += 1;
+                    }
+                } else {
+                    // Very minor warning. We got a bit of malformed markdown (according to
+                    // pulldown_cmark, at least)
```
bugadani, line 393:
```suggestion
                    // Very minor warning. We got a bit of malformed markdown
```

No need for that, I think :)

lapce-data/src/hover.rs
```diff
@@ -152,41 +170,317 @@ impl Default for HoverData {
...
+                builder.push(&text);
+                pos += text.len();
+            }
+            Event::Code(text) => {
+                builder.push(&text).font_family(FontFamily::MONOSPACE);
+                code_block_indices.push(pos..(pos + text.len()));
+                pos += text.len();
+            }
+            // TODO: Some minimal 'parsing' of html could be useful here, since some things use
+            // basic html like `<code>text</code>`.
```
bugadani, line 407: I think we should be consistent with what rustdoc does here. But I understand this is the most significant simplification made in this PR, and for good reasons.

lapce-data/src/hover.rs
```diff
@@ -135,12 +150,15 @@ impl HoverData {
         let items = Arc::make_mut(&mut self.items);
         // Extract the items in the format that we want them to be in
         *items = match resp.contents {
-            HoverContents::Scalar(text) => vec![HoverItem::from(text)],
-            HoverContents::Array(entries) => {
-                entries.into_iter().map(HoverItem::from).collect()
+            HoverContents::Scalar(text) => {
+                vec![HoverItem::from_marked_string(text, style)]
```
bugadani, line 154: For some reason I keep coming back to this code and disliking the single element vectors. Note to self: this is absolutely fine, move on.

bugadani (commented):
Giving this a second go. I can't see anything I really want changed, but hopefully my comments aren't entirely misguided.

lapce-ui/src/hover.rs
MinusGix: Ah, I'll fix that. I originally was using `TextLayout<RichText>` instead of `TextLayout<MarkdownText>` but then I realized I needed to keep track of the code block indices so needed a custom time.  
(It is unifying in the sense that I don't have a `TextLayout<String>` and a `TextLayout<MarkdownText>` as an enum or something, I just treat plaintext as unstyled rich/markdown text)

lapce-data/src/hover.rs
MinusGix: I included that because some of the rust documentation actually uses `<code>stuff</code>`, for some reason.

lapce-data/src/hover.rs
MinusGix: I like using `SmallVec` (and would use it in several places here) in some of my other projects for that reason.

lapce-data/src/hover.rs
MinusGix: Yep. Initially only had base font size but then realized I wanted multiple different styling information and decided to make a structure.

lapce-data/src/hover.rs
MinusGix: I believe we need our own pos because the indices we use are indices into the generated richtext (where we insert newlines and such) - though I hadn't actually seen that function.

lapce-data/src/hover.rs
MinusGix: I think it is currently specific to markdown, even if it could be used more generally.

lapce-data/src/hover.rs
MinusGix: I just removed the entire comment. The pulldown remark was about an error potentially being specific to pulldown, which doesn't try to include all possible fancy markdown variant.

lapce-data/src/hover.rs
MinusGix: Agreed.

MinusGix:
Responded to the various comments and applied fixes/changes.  
Also,
- Moved blockquote to config rather than a constant. Forgot about this. `markdown.blockquote`, since I felt like it didn't fit on editor, but I am honestly unsure where it should go. I gave blockquote a special color since the $greys weren't the most readable.
- Forgot test it on light theme. Introduced a set text color to ensure it is the right color (I'm assuming Druid doesn't clear whatever cache it has when I change the text color to the same color).


Also, I forgot to note this
- No image support currently. Possible to implement but probably not too major, since most documentation does not use images.

lapce-data/src/hover.rs
MinusGix: Changed the comment a bit and moved it before most of the markdown parsing/rendering code. Though I'm unsure what you mean with adding more todos.

lapce-data/src/hover.rs
bugadani: It's used for markdown specifically, but the implementation/general purpose (holding even richer text than `RichText`) isn't related to markdown, I think. Anyway, it's fine.

lapce-data/src/hover.rs
bugadani: > I'm unsure what you mean

Basically what you did. I meant to mark all the relevant structs, but now it's plenty good, thanks.

bugadani (approved):
> No image support currently.

Maybe as a first pass, you could just display links to the images. But, let's push that out of scope of this PR. There are plenty of changes here that stand on their own.

MinusGix:
Think I need to rebase

MinusGix:
Rebased

PR #459, opened 2022-04-27
implement top_of_window and bottom_of_window

zhaojiangbin:
to complement center_of_window

### DIFF
--- a/lapce-ui/src/editor/view.rs
+++ b/lapce-ui/src/editor/view.rs
@@ -1,4 +1,4 @@
-use std::{iter::Iterator, sync::Arc};
+use std::{iter::Iterator, sync::Arc, ops::Sub};
 
 use druid::{
     piet::PietText, BoxConstraints, Command, Data, Env, Event, EventCtx, LayoutCtx,
@@ -11,7 +11,7 @@ use lapce_data::{
         CommandKind, EnsureVisiblePosition, LapceCommand, LapceUICommand,
         LAPCE_COMMAND, LAPCE_UI_COMMAND,
     },
-    config::LapceTheme,
+    config::{LapceTheme, EditorConfig},
     data::{
         EditorTabChild, EditorView, FocusArea, LapceTabData, PanelData, PanelKind,
     },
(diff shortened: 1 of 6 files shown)


zhaojiangbin:
> @zhaojiangbin Could you please rebase this?

It is now rebased.

bugadani (approved):

lapce-ui/src/editor/view.rs
```diff
@@ -259,19 +259,51 @@ impl LapceEditorView {
...
+                let h = if h > margin { h - margin } else { margin };
+                Point::new(cursor.x, cursor.y + h as f64)
+            },
+            EnsureVisiblePosition::BottomOfWindow => {
+                // Cursor line will be at the bottom edge, thus center of the
+                // view is half height up.
+                let h = height as usize;
+                let h = if h > margin { h - margin } else { margin };
+                let y = cursor.y as usize;
+                let y = if h > y { y } else { y - h };
```
bugadani, line 302: This pattern is repeated often enough to deserve its own function. This is basically `saturating_sub`, right? Also I think this code is complicated enough to warrant some tests.

zhaojiangbin:
@dzhou121, what do you think of this PR?

dzhou121:
I assume this is for ```zt``` and ```zb```, can you please add the default key bindings as well?

lapce-ui/src/editor/view.rs
zhaojiangbin: > This pattern is repeated often enough to deserve its own function. ~This is basically `saturating_sub`, right?~ No it's not. 

Indeed the calculation was incorrect. Fixed in the later commits.

> Anyway, I also think this code is complicated enough to warrant some tests.

I assume that you meant unit tests? I was trying to add some but could not figure out how to verify it without repeating the same calculation in the tests. I need to think about it.

zhaojiangbin:
> I assume this is for `zt` and `zb`, can you please add the default key bindings as well?

Done.

dzhou121:
Thanks!

PR #460, opened 2022-04-28
support count in word movement commands

zhaojiangbin:
word_forward, word_backward, word_end_forward

In other words, now you can do <kbd>3w</kbd>, <kbd>5b</kbd>, and <kbd>4e</kbd> in vi mode, for example.


lapce-data/src/buffer.rs
```diff
@@ -1184,6 +1184,20 @@ impl Buffer {
...
     }
+
+    // The folder function calculates offset of the next/previous word boundary
+    // for the given offset.
+    fn move_by_word<F>(
+        &self, offset: usize, count: usize, tab_width: usize, folder: F
+    ) -> (usize, ColPosition)
+    where
+        F: FnMut(usize, ()) -> usize {
+        let new_offset = std::iter::repeat(()).take(count).fold(offset, folder);
```
bugadani, line 1195: This is a bit cryptic to eyes not accustomed to functional primitives like fold. I think a comment might be helpful.

lapce-data/src/buffer.rs
```diff
@@ -1184,6 +1184,20 @@ impl Buffer {
             line
         }
     }
+
+    // The folder function calculates offset of the next/previous word boundary
+    // for the given offset.
```
bugadani, line 1189: Could you turn this into a proper doc comment, please?

bugadani (commented):
Thanks for this!

I'm not sure what the plan is right now. There's a `refactor` branch that seems to touch a lot of editing related code, so I think @dzhou121 would want to hold off merging this PR until work on that branch completes?

zhaojiangbin:
> Thanks for this!
> 
> I'm not sure what the plan is right now. There's a `refactor` branch that seems to touch a lot of editing related code, so I think @dzhou121 would want to hold off merging this PR until work on that branch completes?

I didn't know that. I don't mind waiting.

lapce-data/src/buffer.rs
zhaojiangbin: Will try.

zhaojiangbin:
@dzhou121, @bugadani, is this PR still being considered?

bugadani:
Is there a way to add tests to this? Something like this one? (where did my cute test framework go? 😢 ) https://github.com/lapce/lapce/blob/f0f3dab62ceb4eaf4eca19ac2976227d46f9395a/lapce-core/src/buffer.rs#L1099

bugadani (commented):
Thanks for the changes! I like the general direction (and I especially like how you placed the character offsets under the strings, that's very helpful!), but I still have a few readability concerns. Those are nothing serious, and I don't mind if you decide not to fix those.

zhaojiangbin:
> Thanks for the changes! I like the general direction (and I especially like how you placed the character offsets under the strings, that's very helpful!), but I still have a few readability concerns. Those are nothing serious, and I don't mind if you decide not to fix those.

All good points. Fixed as much as I can.

bugadani (approved):
Thanks :)

PR #470, opened 2022-05-06
allow to disable auto check for plugins

zhaojiangbin:
Also corrected a couple of spelling errors in the existing function names.

I am not sure if this single boolean setting deserves a group by its own. It could have been named as "auto-check-plugins" and added in the "lapce" (aka core) group.

### DIFF
--- a/lapce-data/src/config.rs
+++ b/lapce-data/src/config.rs
@@ -329,10 +329,18 @@ impl Themes {
     }
 }
 
+#[derive(FieldNames, Debug, Clone, Deserialize, Serialize, Default)]
+#[serde(rename_all = "kebab-case")]
+pub struct PluginsConfig {
+    #[field_names(desc = "Check for plugins automatically")]
+    pub auto_check: bool,
+}
+
 #[derive(Debug, Clone, Deserialize, Default)]
 pub struct Config {
     pub lapce: LapceConfig,
     pub editor: EditorConfig,
+    pub plugins: PluginsConfig,
     #[serde(skip)]
     pub themes: Themes,
     #[serde(skip)]
--- a/lapce-ui/src/settings.rs
+++ b/lapce-ui/src/settings.rs
@@ -21,7 +21,7 @@ use lapce_data::{
         CommandExecuted, CommandKind, LapceUICommand, LAPCE_COMMAND,
         LAPCE_UI_COMMAND,
     },
-    config::{EditorConfig, LapceConfig, LapceTheme},
+    config::{EditorConfig, PluginsConfig, LapceConfig, LapceTheme},
     data::{LapceEditorData, LapceTabData},
     document::Document,
     keypress::KeyPressFocus,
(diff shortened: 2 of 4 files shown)


lapce-ui/src/settings.rs
```diff
@@ -364,8 +369,8 @@ impl Widget<LapceTabData> for LapceSettingsPanel {
                     .get_color_unchecked(LapceTheme::EDITOR_BACKGROUND),
             );
 
-            const SETTINGS_SECTIONS: [&str; 3] =
-                ["Core Settings", "Editor Settings", "Keybindings"];
+            const SETTINGS_SECTIONS: [&str; 4] =
+                ["Core Settings", "Editor Settings", "Plugins Settings", "Keybindings"];
```
bugadani, line 373:
```suggestion
                ["Core Settings", "Editor Settings", "Plugin Settings", "Key Bindings"];
```

lapce-data/src/config.rs
```diff
@@ -329,10 +329,18 @@ impl Themes {
     }
 }
 
+#[derive(FieldNames, Debug, Clone, Deserialize, Serialize, Default)]
+#[serde(rename_all = "kebab-case")]
+pub struct PluginsConfig {
+    #[field_names(desc = "Check for plugins automatically")]
+    pub auto_check: bool,
```
bugadani, line 336: What happens when this is false? The plugin list in the plugins panel can't be updated? I think it would be nice to add a manual update option in this case.

zhaojiangbin: > What happens when this is false? The plugin list in the plugins panel can't be updated? I think it would be nice to add a manual update option in this case.

Yes I agree. There should be a "Check Now" button if this setting is false.

In retrospection, this PR is a bit premature. I was only thinking of having the option to not do the fetch on starting up.

I wouldn't mind if this PR is sent back to the drawing board. I need to learn druid to add the "Check Now" button.

bugadani: You can always press the "Convert to draft" button ;)

zhaojiangbin: > You can always press the "Convert to draft" button ;)

Yeah I will think about it.

By the way, I noticed that we are not using the druid widgets such as button, checkbox, etc. For example, I am looking at https://github.com/lapce/lapce/blob/6e449b1d5cc62dbb2453602668d3b7dab24df91b/lapce-ui/src/keymap.rs#L451-L550

IIUC, this code draws a popup with a "Cancel" button and a a "Save" button. Then clicking the buttons is handled by https://github.com/lapce/lapce/blob/6e449b1d5cc62dbb2453602668d3b7dab24df91b/lapce-ui/src/keymap.rs#L58-L73

What're the reasons that we are creating the UI this way?

bugadani: @dzhou121 knows this

hbina:
Can I pick this up if you don't mind @zhaojiangbin ?

panekj:
This PR became quite stale and we're moving around the UI stuff soon that will change a lot of things, so I'm closing this.

PR #474, opened 2022-05-07
Introduce `ReadOnlyString`

bugadani:
There are probably more places where this could be used, but for now, Panel headers only.

This PR introduces a read-only string type that is either a `&'static str` or a cheap to clone `Arc<str>`. This PR implements `TextStorage` for this type so it can be used in text layouts.

### DIFF
--- a/lapce-ui/src/explorer.rs
+++ b/lapce-ui/src/explorer.rs
@@ -223,7 +223,7 @@ impl FileExplorer {
             data.file_explorer.widget_id,
             split_id,
             SplitDirection::Vertical,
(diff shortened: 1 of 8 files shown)


bugadani:
This is a drop in the ocean at this point, but I'm hoping this can be used more widely to avoid allocating for otherwise static string.

zhaojiangbin:
Will `ReadOnlyString` work with `druid::LocalizedString` when we start localizing the UI?

bugadani:
I'm not familiar with that but the transition shouldn't be harder than from `String`

PR #476, opened 2022-05-07
Rework Problems lists to use Druid widgets

### LINKED ISSUE #467: Line height set for text editor not suitable for error display box
The lines in the text area appear squashed if a line height that looks good for main editor is used.

This is illustrated by the screenshots in this issue: https://github.com/lapce/lapce/issues/466

The line height setting being used is 18, and font is `Comic Mono`, same result with other fonts like `Cascadia Code`.

bugadani:
Fixes #467

This PR also implements a few useful helpers.

### DIFF
--- a/lapce-data/src/container.rs
+++ b/lapce-data/src/container.rs
@@ -1,4 +1,3 @@
-
 use druid::{Point, Size};
 
 pub struct ChildState {
--- a/lapce-data/src/data.rs
+++ b/lapce-data/src/data.rs
@@ -336,7 +336,7 @@ impl LapceWindowData {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, PartialEq)]
 pub struct EditorDiagnostic {
     pub range: Option<(usize, usize)>,
     pub diagnostic: Diagnostic,
--- a/lapce-ui/src/app.rs
+++ b/lapce-ui/src/app.rs
@@ -9,8 +9,8 @@ use lapce_data::{
     db::{TabsInfo, WindowInfo},
 };
 
-use crate::window::LapceWindowNew;
 use crate::logging::override_log_levels;
+use crate::window::LapceWindowNew;
 
 pub fn build_window(data: &LapceWindowData) -> impl Widget<LapceData> {
     LapceWindowNew::new(data).lens(LapceWindowLens(data.window_id))
--- a/lapce-ui/src/lib.rs
+++ b/lapce-ui/src/lib.rs
@@ -27,3 +27,5 @@ mod tab;
 pub mod terminal;
 pub mod title;
 pub mod window;
+
+pub mod widgets;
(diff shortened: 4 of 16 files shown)


dzhou121:
If I’m looking at this correctly. Is each item becoming a single Widget? I’m not sure about this approach because if say we have 1000 items, the layout call feels quite expensive.

bugadani:
It's fine.

bugadani:
There's still one issue where the hover state gets stuck when scrolling with the mouse wheel. The issue is not obvious but I'll debug it tomorrow.

dzhou121:
I've had a test with a list of 2K+ warnings. And the lag is very much noticeable. That's the expensive layout call which is triggered on every scroll. 

It's inevitable that you'll have to draw on a canvas if there are lots of elements, to achieve good performance.

bugadani:
Maybe. I'm curious how perf would look like when the height of the individual items are correctly measured, in the original implementation.

Also, there's very little reason to recalculate things if the list doesn't change, so I think this can still be improved. Let's not be so negative, please.

Also, 2k warnings in a project should be rather rare. Or if it's not, the Problems panel is probably not open very often. But I understand your concern.

dzhou121:
I agree that not recalculating would greatly improve that, but the initial layout calculating still can potentially block my typing which I'd like to avoid. 

I can see how UI composition can simplify things and do agree that we should use them more(e.g. buttons).

bugadani:
I get what you mean and I have ideas. An optimized, scrollable list view could help, without giving up much of the advantages. That will take more time, though.

dzhou121:
It would be fantastic if it could work out. Then it can be used for file explorer etc.

bugadani:
> It would be fantastic if it could work out

Definitely!

What did you use to generate 2k warnings? I would like to have something I can use to test.

dzhou121:
It’s another project I’m working on that’s not open source. 

I guess you could put lots of unused imports, or hard code some warnings.

bugadani:
I've given some thought to what's happening with performance.

Layouting will always be an O(n) operation if we want to handle multi-line diagnostics correctly. There's no way around it, because every piece of text must be measured. I believe the effects can be mitigated by only adding certain number of rows at once to the list and making sure only the new ones are laid out. A re-layout request would need to rebuild sequentially again, which means scroll position will probably be lost in those cases.

On the other hand, event handling with this PR is O(n) instead of O(1). I believe this is where the bad performance comes from right now: events are routed to all the children, I think regardless whether they can be handled or not. A few things can be done here:
 - `Flex` layout contains a bunch of trait objects that make it impossible to optimize out no-op event handlers. A custom linear layout or an event gate widget might help here, but they only reduce the scope of the problem.
 - A combined scroll-list widget could route mouse events to only the visible widgets. The result won't be completely general, because some events might need to be routed all the way down. This might still be worth investigating.

We can also limit how many problems we show, but that's probably the worst thing we could do. I still haven't given up on this PR, but this isn't an easy one.

PR #481, opened 2022-05-09
make deps on the tree-sitter lang crates optional

zhaojiangbin:
This can cut down the binary size noticeably. For example, with only rust, toml, markdown and json selected, the release build of the lapce binary is 26M stripped compared to 35M with all languages.

For backward compatibility, a simple `cargo build` still compiles all languages in. To select only some languages, one would need to do this, for example:

  `cargo build -p lapce-ui --no-default-feautures --features lang-rust,lang-toml,lang-md,lang-json`

If no language is selected at all, build still succeed with a bunch of "unused" and "unreachable" warnings:

  `cargo build -p lapce-ui --no-default-features`

Perhaps we should have some languages as non-optional?

As an interesting note, helix can build the tree-sitter lang crates into dynamic libs according to a user setting and install them into its runtime directory. Its build system can query the user setting and only compiles the selected languages after git pull.


lapce-core/src/language.rs
```diff
@@ -4,179 +4,388 @@ use tree_sitter::{Parser, TreeCursor};
...
-    "decorated_definition",
-    "block",
-];
-const PYTHON_CODE_LENS_IGNORE_LIST: &[&str] =
-    &["source_file", "import_statement", "import_from_statement"];
-const JAVASCRIPT_CODE_LENS_LIST: &[&str] = &["source_file", "program"];
-const JAVASCRIPT_CODE_LENS_IGNORE_LIST: &[&str] = &["source_file"];
 
-#[derive(Eq, PartialEq, Hash, Copy, Clone, Debug)]
+struct Setting {
```
bugadani, line 75: I think this struct should have the file extensions as well. I'm also not sure about the name (these are not settings, but items in a pre-defined set).

lapce-core/src/language.rs
```diff
@@ -4,179 +4,388 @@ use tree_sitter::{Parser, TreeCursor};
...
+    language: fn() -> tree_sitter::Language,
+    /// For most languages, it is `tree_sitter_$crate::HIGHLIGHT_QUERY`.
+    highlight: &'static str,
+    /// The comment token.  "#" for python, "//" for rust for example.
+    comment: &'static str,
+    /// The indent unit.  "\t" for python, "    " for rust, for example.
+    indent: &'static str,
+    /// If unsure, use `DEFAULT_CODE_LENS_LIST` and
+    /// `DEFAULT_CODE_LENS_IGNORE_LIST`.
+    code_lens: (&'static [&'static str], &'static [&'static str]),
```
bugadani, line 92: Can we have two fields here instead of a tuple?

lapce-core/src/language.rs
```diff
@@ -4,179 +4,388 @@ use tree_sitter::{Parser, TreeCursor};
...
-            LapceLanguage::Elixir => "#",
-            LapceLanguage::C => "//",
-            LapceLanguage::Cpp => "//",
-            LapceLanguage::Json => "",
-            LapceLanguage::Markdown => "",
-            LapceLanguage::Ruby => "#",
-            LapceLanguage::Html => "",
-            LapceLanguage::Java => "//",
-        }
+    fn find(&self) -> &Setting {
```
bugadani, line 364: `find` is an iterator method. I would like this to be called something else to avoid confusion. Maybe `details` would work?

lapce-ui/Cargo.toml
```diff
@@ -50,3 +50,43 @@ structdesc = { git = "https://github.com/lapce/structdesc" }
...
+lang-javascript= ["lapce-core/lang-javascript"]
+lang-typescript = ["lapce-core/lang-typescript"]
+lang-python = ["lapce-core/lang-python"]
+lang-toml = ["lapce-core/lang-toml"]
+lang-elixir = ["lapce-core/lang-elixir"]
+lang-php = ["lapce-core/lang-php"]
+lang-ruby= ["lapce-core/lang-ruby"]
+lang-c = ["lapce-core/lang-c"]
+lang-cpp = ["lapce-core/lang-cpp"]
+lang-json= ["lapce-core/lang-json"]
```
bugadani, line 89: I'm not going to point out formatting issues, but here's one (and one for JS) that I don't think rustfmt would fix.

bugadani (commented):
I know I said this is a good idea but I'm no longer sure. A 35MB executable shouldn't be considered large today, in the age of multi-GB software.

In the long run, I would prefer if we either moved the bulk of the syntax highlight code to plugins, or not but provide an interface for plugins to bring in support for other languages. The current hard-coded approach works for now, but I don't think it's sustainable.

Having said that, during development the most time consuming part is linking the debug binary, which takes around 30s for me. If omitting tree-sitter binaries help with that, I'm all in for this PR.

zhaojiangbin:
> I know I said this is a good idea but I'm no longer sure. A 35MB executable shouldn't be considered large today, in the age of multi-GB software.
> 
> In the long run, I would prefer if we either moved the bulk of the syntax highlight code to plugins, or not but provide an interface for plugins to bring in support for other languages. The current hard-coded approach works for now, but I don't think it's sustainable.
> 
> Having said that, during development the most time consuming part is linking the debug binary, which takes around 30s for me. If omitting tree-sitter binaries help with that, I'm all in for this PR.
> 
> Yes, I'm aware that I'm contradicting myself in this comment 😅

It was the long build time, too, that motivated me to make this change. But the saving in build time isn't noticeable to me, probably because I have replaced the default linker with zld on macOS at the same time.

Yeah in long term lapce will need to put the tree-sitter parsers in plugins as the number of supported languages keeps growing. This is why I mentioned what helix does in the PR description.

lapce-core/src/language.rs
zhaojiangbin: > I think this struct should have the file extensions as well. I'm also not sure about the name (these are not settings, but items in a pre-defined set).

For the struct name, how about `SyntaxProperties`?

I actually thought about having the file extensions in this struct. I give up because the `from_path()` method will then probably not be able to use pattern match any more.

Do you think a linear search in the `LANGUAGES` array will be acceptable? It will not be in any tight loops.

lapce-core/src/language.rs
zhaojiangbin: > Can we have two fields here instead of a tuple?

Sure.

lapce-core/src/language.rs
zhaojiangbin: > `find` is an iterator method. I would like this to be called something else to avoid confusion. Maybe `details` would work?

It was actually named as `find_setting` before. I will rename it as `find_properties` or `find_details` depending on the struct name.

lapce-ui/Cargo.toml
zhaojiangbin: > I'm not going to point out formatting issues, but here's one (and one for JS) that I don't think rustfmt would fix.

I am not following. Can you elaborate?

lapce-core/src/language.rs
bugadani: I don't think a linear search is a problem at this scale. I'm not entirely convinced that pattern matching is not linear in this case, anyway.

`SyntaxProperties` sounds fine to me.

lapce-core/src/language.rs
bugadani: I don't think you need to "find" something if you know where it is ;) But sure, that's a good solution.

lapce-ui/Cargo.toml
bugadani:
```suggestion
lang-json = ["lapce-core/lang-json"]
```

lapce-core/src/language.rs
zhaojiangbin: > I don't think a linear search is a problem at this scale. I'm not entirely convinced that pattern matching is not linear in this case, anyway.

I will try to implement the linear search then. That will cut down the need for cfg feature gates, too.

> `SyntaxProperties` sounds fine to me.

Ok.

lapce-core/src/language.rs
zhaojiangbin: I was trying to add some comments to explain the tuple. But I don't know what to say about the two lists without knowing tree-sitter. I will leave a "TODO" in the comment for someone else to address it.

lapce-core/src/language.rs
```diff
@@ -425,3 +410,77 @@ fn walk_tree(
...
+    #[test]
+    #[cfg(feature = "lang-rust")]
+    fn test_rust_lang() {
+        use super::LapceLanguage;
+        use std::path::PathBuf;
+        let path = PathBuf::from("test.rs");
+        let lang = LapceLanguage::from_path(&path);
+
+        assert!(lang.is_some());
+        let lang = lang.unwrap();
```
bugadani, line 441: `unwrap()` panics if `lang` is `None` so you don't need the assert.

lapce-core/src/language.rs
```diff
@@ -425,3 +410,77 @@ fn walk_tree(
...
+// If clicking the "Run Test" buttons attached to the functions, RA will add
+// `--feature lang-rust`, for example, for you to the cargo test command line,
+// in addition to any features in its workspace settings.
+#[cfg(test)]
+mod test {
+    #[test]
+    #[cfg(feature = "lang-rust")]
+    fn test_rust_lang() {
+        use super::LapceLanguage;
+        use std::path::PathBuf;
```
bugadani, line 436: These look weird. In this function, you import these types, in the others you don't. It would be better to import at the module level and allow unused code on them, I think.

lapce-core/src/language.rs
```diff
@@ -425,3 +410,77 @@ fn walk_tree(
...
+        let props = lang.properties();
+        assert_eq!(lang.comment_token(), props.comment);
+        assert_eq!(lang.indent_unit(), props.indent);
+
+        // If a programming language in the future uses this file extension, it
+        // will not be Rust.
+        let path = PathBuf::from("test.not_rust");
+
+        let lang = LapceLanguage::from_path(&path);
+        if lang.is_none() {
```
bugadani, line 452: `assert!(true)` still does not do anything.

lapce-core/src/language.rs
```diff
@@ -425,3 +410,77 @@ fn walk_tree(
...
+        use std::path::PathBuf;
+        let path = PathBuf::from("test.rs");
+        let lang = LapceLanguage::from_path(&path);
+
+        assert!(lang.is_some());
+        let lang = lang.unwrap();
+        assert_eq!(lang, LapceLanguage::Rust);
+        let props = lang.properties();
+        assert_eq!(lang.comment_token(), props.comment);
+        assert_eq!(lang.indent_unit(), props.indent);
```
bugadani, line 445: You can extract all these checks into a helper function and call that function in all three of these tests.

lapce-core/src/language.rs
```diff
@@ -425,3 +410,77 @@ fn walk_tree(
...
+        // If a programming language in the future uses this file extension, it
+        // will not be Rust.
+        let path = PathBuf::from("test.not_rust");
+
+        let lang = LapceLanguage::from_path(&path);
+        if lang.is_none() {
+            assert!(true)
+        } else {
+            assert_ne!(lang.unwrap(), LapceLanguage::Rust);
+        }
```
bugadani, line 456:
```suggestion
       assert!(!matches!(lang, Some(LapceLanguage::Rust));
```

bugadani: If this test is done only once, why are the others repeated?

bugadani (changes requested):
Thanks for the changes!

I think the tests could use a bit more love. It's not entirely clear what the tests actually do (i.e. verifying that the struct returned by `LapceLanguage::properties()` contains the same data as the individual getters give). The checks are (in my opinion) needlessly repeated across the three tests.

The really useful part is testing the `from_path` methods.

I think the tests should look something along the lines of: (assert functions containing all the checks you have)

```rust
fn test_rust() {
assert_language("test.rs", LapceLanguage::Rust);
assert_not_language("test.not_rs", LapceLanguage::Rust);
}
```

zhaojiangbin:
> I think the tests could use a bit more love. 

You are right. They were not well thought of. Will follow up with your comments.

jirutka:
It would be much better to support dynamically loadable tree-sitter grammars. Please see my comment on this in https://github.com/Wilfred/difftastic/issues/123 for more information.

bugadani:
> It would be much better to support dynamically loadable tree-sitter grammars. Please see my comment on this in [Wilfred/difftastic#123](https://github.com/Wilfred/difftastic/issues/123) for more information.

Helix's implementation is a good one we could follow, for sure. Thanks for the link!

I'm not sure how we should proceed with this PR. I think this would be a good baseline at first, and we could see how much of an improvement dynamic loading would give.

bugadani (approved):
I'm going to approve this PR as it looks a lot better now, thanks for the changes! @zhaojiangbin would you be interested in working dynamic loading after this?

zhaojiangbin:
> I'm going to approve this PR as it looks a lot better now, thanks for the changes! @zhaojiangbin would you be interested in working dynamic loading after this?

Sure I can give it a try. But it will be low on my list. I will create an feature request issue for reference.

PR #490, opened 2022-05-11
Fix ci

tormeh:
Fixes https://github.com/lapce/lapce/issues/491


bugadani:
Thanks for the fix :)

PR #508, opened 2022-05-15
Don't clone syntax highlight results

bugadani:


bugadani:
I was considering adding a mutex to limit how many syntax highlight processes can happen at once, but decided against it. I might revisit later if I can find a perf improvement when working with large files, I didn't try that.

PR #537, opened 2022-05-22
Show diagnostics on hover

MinusGix:
This lets the user hover over a diagnostic in the editor and have the text displayed in the hover window.  
I decided to put the diagnostics before the documentation (VSCode does after) because typically they are small and more pertinent anyway, though making a config setting to support both would not be hard.  
There are some improvements that could be made, such as showing related things (ex, I believe: if you use a non-mutable variable in a way that needs it to be mutable, and rustc will provide an extra underline under the definition of that variable along with the diagnostic where you use it incorrectly), but this works as an initial impl.
I drew a rect for the separator since I don't know of any other place that uses a separator like that, so I was unsure if there was existing methods for doing it and/or colors.

image
image
  
Though, perhaps there should be a bit of margin on the top?

### DIFF
--- a/lapce-ui/src/hover.rs
+++ b/lapce-ui/src/hover.rs
@@ -1,7 +1,7 @@
 use std::sync::Arc;
 
 use druid::{
-    theme, ArcStr, BoxConstraints, Command, Data, Env, Event, EventCtx,
+    kurbo::Line, theme, ArcStr, BoxConstraints, Command, Data, Env, Event, EventCtx,
     FontDescriptor, LayoutCtx, LifeCycle, LifeCycleCtx, PaintCtx, Point,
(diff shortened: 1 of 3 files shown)


lapce-ui/src/hover.rs
```diff
@@ -281,28 +309,80 @@ impl Widget<LapceTabData> for Hover {
...
+                let x0 = rect.x0 + side_margin;
+                let y0 = diagnostic_text_metrics.size.height + line_height;
+                let x1 = rect.x1 - side_margin;
+                let y1 = y0 + SEP_HEIGHT;
+                Rect::new(x0, y0, x1, y1)
+            };
+
+            // Draw the separator
+            ctx.fill(
+                line_rect,
```
dzhou121, line 374: Maybe use a druid::Line instead of a Rect?

dzhou121:
Awesome work! I like the diagnostics is above documentation.

Issue #374, opened 2022-04-04
Typescript Support


divee789:
Are there any plans to support typescript ?

bugadani:
Hi!

Lapce should already have typescript syntax highlighting. Lapce can also support [TypeScript Language Server](https://github.com/typescript-language-server/typescript-language-server) in principle, but it's not done currently. We are still working on the basics and wider language support should come at a later stage, I believe.

Having said that, contributions to enable LSP support for various languages are welcome!

ash-hashtag:
How to setup typescript?

trsh:
Typescript is a must be :P

panekj:
There is already a typescript plugin https://github.com/panekj/lapce-typescript/tree/wip but it currently doesn't work

ghishadow:
typescript lsp is working for me for sometime. Thanks to @panekj , does this issue require anything else ?
<img width="779" alt="tscompeletion" src="https://user-images.githubusercontent.com/9583775/194459675-0bce1c36-32d0-443d-9e50-d9dbcf68daa4.png">

panekj:
It was fixed via #1263 (probably)

PR #549, opened 2022-05-25
Golang Support

### LINKED ISSUE #365: Golang support?
Does lapce support golang? If so, please describe a needed configuration, thanks.

nheuillet:
This is a new version of the Golang support PR, up to date and cleaner than the old #385, as discussed with @dzhou121.
Here is a paste of the previous PR message:

## Introduction

This PR pushes the LSP client to a state where the Golang LSP server (`gopls`) is usable.
While lapce might be missing plenty of capabilities (some of which might already be implemented and wirable), the goal of this PR is not to fill this gap but to make sure the base client is robust enough for any LSP server to run without any primary issues.

This PR should allow starting proper work on #365 and #374, and any other LSP server support.

*Note: A very basic and hacky gopls plugin can be found [here](https://github.com/nheuillet/lapce-go) for the time being.* 

closes #365 

## Work
-  [X] Implement response on `window/workDoneProgress/create` request (Blocking "handshake" for gopls)
-  [X] Fix EditDocument error on gopls edit suggestion following a `textDocument/completion` request
-  [X] Fix lowercased suggestion on ui
-  [X] Implement env passing for wasi runtime
-  [x] Fix misplaced error marker  

## How to test

- Clone this PR and build the code
- Clone the [lapce-go plugin](https://github.com/nheuillet/lapce-go) wherever you want
- Build the plugin
  - I personally use these commands (see [this for cargo wasi](https://github.com/bytecodealliance/cargo-wasi)):
```bash
  # link the plugin folder to lapce plugin directory
  ln -s $PWD ~/.lapce/plugins/lapce-go
  #  build the actual wasm binary
  cargo wasi build && cp target/wasm32-wasi/debug/lapce-go.rustc.wasm lapce-go.wasm 

```
- Lapce should be able to launch the plugin without any issue. To confirm, look for the output of lapce. if "Starting lapce-go plugin!" appears, the plugin is working.

### DIFF
--- a/lapce-data/src/editor.rs
+++ b/lapce-data/src/editor.rs
@@ -289,10 +289,12 @@ impl LapceEditorBufferData {
                     })
                     .collect::<Vec<(lapce_core::selection::Selection, &str)>>()
             });
-        let additional_edit: Option<Vec<_>> =
-            additional_edit.as_ref().map(|edits| {
+        let additional_edit: Vec<_> = additional_edit
+            .as_ref()
+            .map(|edits| {
                 edits.iter().map(|(selection, c)| (selection, *c)).collect()
-            });
+            })
+            .unwrap_or_default();
 
         let text_format = item
             .insert_text_format
@@ -317,7 +319,7 @@ impl LapceEditorBufferData {
                                 .do_raw_edit(
                                     &[
                                         &[(&selection, edit.new_text.as_str())][..],
-                                        &additional_edit.unwrap_or_default()[..],
+                                        &additional_edit[..],
                                     ]
                                     .concat(),
                                     lapce_core::editor::EditType::InsertChars,
@@ -340,7 +342,7 @@ impl LapceEditorBufferData {
                                 .do_raw_edit(
                                     &[
                                         &[(&selection, text.as_str())][..],
-                                        &additional_edit.unwrap_or_default()[..],
+                                        &additional_edit[..],
                                     ]
                                     .concat(),
                                     lapce_core::editor::EditType::InsertChars,
@@ -397,7 +399,7 @@ impl LapceEditorBufferData {
                     &selection,
                     item.insert_text.as_deref().unwrap_or(item.label.as_str()),
                 )][..],
-                &additional_edit.unwrap_or_default()[..],
+                &additional_edit[..],
             ]
             .concat(),
             lapce_core::editor::EditType::InsertChars,
--- a/lapce-proxy/src/lsp.rs
+++ b/lapce-proxy/src/lsp.rs
@@ -57,6 +57,7 @@ pub struct LspState {
 #[derive(Clone)]
 pub struct LspClient {
     exec_path: String,
+    binary_args: Vec<String>,
     options: Option<Value>,
     state: Arc<Mutex<LspState>>,
     dispatcher: Dispatcher,
@@ -84,13 +85,49 @@ impl LspCatalog {
         language_id: &str,
         options: Option<Value>,
     ) {
+        let binary_args = self.get_plugin_binary_args(options.clone());
         let client = LspClient::new(
             language_id.to_string(),
             exec_path,
             options,
+            binary_args.to_owned(),
             self.dispatcher.clone().unwrap(),
         );
         self.clients.insert(language_id.to_string(), client);
+    }
+
+    fn get_plugin_binary_args(&mut self, option: Option<Value>) -> Vec<String> {
+        let mut vals = Vec::new();
+
+        let option = match option {
+            Some(val) => val,
+            None => {
+                return vals;
+            }
+        };
+
+        let binary = match option["binary"].as_object() {
+            Some(binary) => binary,
+            None => return vals,
+        };
+
+        let option = match binary.get("binary_args") {
+            Some(binary_args) => binary_args,
+            None => return vals,
+        };
+
+        let option = if let Some(option) = option.as_array() {
+            option
+        } else {
+            println!("binary_args value should be of type [String].");
+            return vals;
+        };
+
+        for val in option {
+            vals.push(String::from(val.as_str().unwrap()));
+        }
+
+        return vals;
     }
 
     pub fn new_buffer(
@@ -368,15 +405,18 @@ impl LspClient {
         _language_id: String,
         exec_path: &str,
         options: Option<Value>,
+        binary_args: Vec<String>,
         dispatcher: Dispatcher,
     ) -> Arc<LspClient> {
-        let mut process = Self::process(exec_path);
+        //TODO: better handling of binary args in plugin
+        let mut process = Self::process(exec_path, binary_args.clone());
         let writer = Box::new(BufWriter::new(process.stdin.take().unwrap()));
         let stdout = process.stdout.take().unwrap();
 
         let lsp_client = Arc::new(LspClient {
             dispatcher,
             exec_path: exec_path.to_string(),
+            binary_args: binary_args,
             options,
             state: Arc::new(Mutex::new(LspState {
                 next_id: 0,
@@ -414,8 +454,12 @@ impl LspClient {
         });
     }
 
-    fn process(exec_path: &str) -> Child {
+    fn process(exec_path: &str, binary_args: Vec<String>) -> Child {
         let mut process = Command::new(exec_path);
+
+        for arg in binary_args {
+            process.arg(arg);
+        }
         #[cfg(target_os = "windows")]
         let process = process.creation_flags(0x08000000);
         process
@@ -426,7 +470,8 @@ impl LspClient {
     }
 
     fn reload(&self) {
-        let mut process = Self::process(&self.exec_path);
+        //TODO: avoid clone using a &[String] ?
+        let mut process = Self::process(&self.exec_path, self.binary_args.clone());
         let writer = Box::new(BufWriter::new(process.stdin.take().unwrap()));
         let stdout = process.stdout.take().unwrap();
 
@@ -473,9 +518,15 @@ impl LspClient {
     }
 
     pub fn handle_message(&self, message: &str) {
+        println!("Received message! {}", message);
         match JsonRpc::parse(message) {
-            Ok(JsonRpc::Request(_obj)) => {
-                // trace!("client received unexpected request: {:?}", obj)
+            Ok(value @ JsonRpc::Request(_)) => {
+                let id = value.get_id().unwrap();
+                self.handle_request(
+                    value.get_method().unwrap(),
+                    id,
+                    value.get_params().unwrap(),
+                )
             }
             Ok(value @ JsonRpc::Notification(_)) => {
                 self.handle_notification(
@@ -497,6 +548,20 @@ impl LspClient {
         }
     }
 
+    pub fn handle_request(&self, method: &str, id: Id, _params: Params) {
+        match method {
+            "window/workDoneProgress/create" => {
+                // Token is ignored as the workProgress Widget is always working
+                // In the future, for multiple workProgress Handling we should
+                // probably store the token
+                self.send_success_response(id, &json!({}));
+            }
+            method => {
+                println!("Received unhandled request {method}");
+            }
+        }
+    }
+
     pub fn handle_notification(&self, method: &str, params: Params) {
         match method {
             "textDocument/publishDiagnostics" => {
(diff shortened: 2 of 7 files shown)


nheuillet:
@bugadani Here is a cleaner PR, I have cleaned a few of the last issues you raised, but some might still be there. I think before going further a proper discussion with @dzhou121 regarding the plugin runtime capabilities (fully sandboxed? What should be allowed, not allowed? What should be a Lapce command, what should be deferred at plugin runtime or before starting the plugin?).

The biggest example in this PR is the handling of the environment variables generation, which could be delegated to the plugin using a Lapce command "Exec".

lapce-proxy/src/lsp.rs
```diff
@@ -84,15 +85,51 @@ impl LspCatalog {
         language_id: &str,
         options: Option<Value>,
     ) {
+        let binary_args = self.get_plugin_binary_args(options.clone());
         let client = LspClient::new(
             language_id.to_string(),
             exec_path,
             options,
+            binary_args.to_owned(),
```
bugadani, line 93: Why do you call `to_owned` on a `Vec`? If you need a copy of it, call `clone()` instead, it's cleaner. Otherwise, just pass ownership as I don't see any other use of the variable.

lapce-proxy/src/lsp.rs
```diff
@@ -84,15 +85,51 @@ impl LspCatalog {
...
     }
 
+    fn get_plugin_binary_args(&mut self, option: Option<Value>) -> Vec<String> {
+        let mut vals = Vec::new();
+
+        let option = match option {
+            Some(val) => val,
+            None => {
+                return vals;
+            }
```
bugadani, line 106: Here's a bit of a stylistic inconsistency.

lapce-proxy/src/lsp.rs
```diff
@@ -84,15 +85,51 @@ impl LspCatalog {
...
+
+        let option = match binary.get("binary_args") {
+            Some(binary_args) => binary_args,
+            None => return vals,
+        };
+
+        let option = if let Some(option) = option.as_array() {
+            option
+        } else {
+            println!("binary_args value should be of type [String].");
```
bugadani, line 122: This should probably be a `log::warn!` instead?

lapce-proxy/src/lsp.rs
```diff
@@ -84,15 +85,51 @@ impl LspCatalog {
...
+            return vals;
+        };
+
+        for val in option {
+            vals.push(String::from(val.as_str().unwrap()));
```
bugadani, line 127: This might be much nicer as `option.iter().map(Value::as_str).map(Option::unwrap).(ToString::to_string).collect::<Vec<_>>()` or something like that, but my main issue is the `String::from` here, which just smells weird on a `str`.

You might not even need to specify the type in `collect`, I don't know.

lapce-proxy/src/lsp.rs
```diff
@@ -84,15 +85,51 @@ impl LspCatalog {
...
+        }
+
+        return vals;
```
bugadani, line 130: - You don't need to spell out the `return`
 - I think with the above comment you can avoid the `vals` variable completely
 - If this function returned `Option<Vec...` instead, you might simplify the early returns into `?`s. Also this would help differentiating between "no args" and the error cases.

lapce-proxy/src/lsp.rs
```diff
@@ -368,15 +405,18 @@ impl LspClient {
...
-        let mut process = Self::process(exec_path);
+        //TODO: better handling of binary args in plugin
+        let mut process = Self::process(exec_path, binary_args.clone());
         let writer = Box::new(BufWriter::new(process.stdin.take().unwrap()));
         let stdout = process.stdout.take().unwrap();
 
         let lsp_client = Arc::new(LspClient {
             dispatcher,
             exec_path: exec_path.to_string(),
+            binary_args: binary_args,
```
bugadani, line 419:
```suggestion
            binary_args,
```

lapce-proxy/src/lsp.rs
```diff
@@ -414,8 +454,12 @@ impl LspClient {
         });
     }
 
-    fn process(exec_path: &str) -> Child {
+    fn process(exec_path: &str, binary_args: Vec<String>) -> Child {
         let mut process = Command::new(exec_path);
+
+        for arg in binary_args {
+            process.arg(arg);
```
bugadani, line 461: I think https://doc.rust-lang.org/std/process/struct.Command.html#method.args would work better here.

lapce-proxy/src/lsp.rs
```diff
@@ -414,8 +454,12 @@ impl LspClient {
         });
     }
 
-    fn process(exec_path: &str) -> Child {
+    fn process(exec_path: &str, binary_args: Vec<String>) -> Child {
```
bugadani, line 457:
```suggestion
    fn process(exec_path: &str, binary_args: &[String]) -> Child {
```

Oh, I see this is one of your TODOs.

lapce-proxy/src/plugin.rs
```diff
@@ -143,10 +143,15 @@ impl PluginCatalog {
 
         let output = Pipe::new();
         let input = Pipe::new();
+        let env = match plugin_desc.get_plugin_env() {
+            Ok(env) => env,
+            Err(err) => return Err(err),
+        };
```
bugadani, line 149:
```suggestion
        let env = plugin_desc.get_plugin_env()?;
```

Assuming this compiles, of course :)

lapce-rpc/src/plugin.rs
```diff
@@ -26,3 +27,81 @@ pub struct PluginInfo {
...
+    pub fn get_plugin_env(&self) -> Result<Vec<(String, String)>, Error> {
+        let mut vars = Vec::new();
+
+        let conf = match &self.configuration {
+            Some(val) => val,
+            None => {
+                return Err(format_err!(
+                    "Empty configuration for plugin {}",
+                    self.display_name
+                ))
```
bugadani, line 41: I would prefer `;`s after these returns.

lapce-data/src/editor.rs
```diff
@@ -284,9 +284,10 @@ impl LapceEditorBufferData {
                     })
                     .collect::<Vec<(lapce_core::selection::Selection, &str)>>()
             });
-        let additioal_edit: Option<Vec<_>> = additional_edit.as_ref().map(|edits| {
-            edits.iter().map(|(selection, c)| (selection, *c)).collect()
-        });
+        let additional_edit: Option<Vec<_>> =
```
bugadani, line 287: I know you only fixed the typo here, but can we have the `unwrap_or_default` call here, please?

bugadani (commented):
As usual, I mostly have stylistic comments, which is actually not a bad thing.

>  I think before going further

Good idea.

yuriizinets:
Any progress on this? Really interested in Go support.

lapce-proxy/src/lsp.rs
Eraden: Or even

```
let vals = option.iter().filter_map(Value::as_str).map(ToString::to_string).collect::<Vec<_>>()
```

lapce-proxy/src/lsp.rs
```diff
@@ -84,15 +85,51 @@ impl LspCatalog {
...
+            println!("binary_args value should be of type [String].");
+            return vals;
+        };
+
+        for val in option {
+            vals.push(String::from(val.as_str().unwrap()));
+        }
+
+        return vals;
+    }
```
Eraden, line 131: I feel entire function would benefit in changing return type from `Vec<String>` to `Option<Vec<String>>` and handling None outside with `.unwrap_or_default()`

At least 3 sections can be change to

```
let option = option["binary"]
    .as_object()?
    .get("binary_args")?
    .as_array()?;
```

lapce-rpc/src/plugin.rs
```diff
@@ -26,3 +27,81 @@ pub struct PluginInfo {
...
+            }
+        };
+
+        for l in data.lines() {
+            if let Some((key, value)) = l.split_once('=') {
+                vars.push((String::from(key), String::from(value)));
+            };
+        }
+
+        return Ok(vars);
```
Eraden, line 105:
```suggestion
        data.lines()
            .filter_map(|l| l.split_once('='))
            .map(|(k, v)| (k.into(), v.into()))
            .collect::<Vec<(String, String)>>()
```

And `let mut vars = Vec::new();` can be removed

yuriizinets:
@nheuillet Hi!
Is there any way I can try this plugin without waiting for merge into master?
I really would like to try this editor as my primary one, but Go support is critical in my case :)

nheuillet:
Hello @yuriizinets ,
I have been on vacation for the past weeks and just came back, so I should start working on it this weekend. In the meantime as long as you compile this branch by yourself and use the go plugin I wrote and linked in the description of the PR, you should be ready to go! I will try to rebase asap.

nheuillet:
Hi @yuriizinets,

this PR is now rebased on master. You only need to follow the instructions in the description to try it!

yuriizinets:
Many thanks for your work!

nheuillet:
No problem!

Please let me know if you encounter some weird behaviors. I know one annoying bug (though I think this falls on the capability not yet implemented) is the fact that the "delete" format doesn't work (as in, an import that became obsolete is not removed).

dzhou121:
Thank you so much!

yuriizinets:
That's strange. I've installed latest night build, compiled a plugin, moved to plugins directory, but nothing appears in the Lapce (completion is not working too).

image

image

ghishadow:
@yuriizinets looks like it is not added to https://github.com/lapce/lapce.github.io/blob/master/plugins.json from which plugin list is shown, it will still work though

yuriizinets:
On the screenshot I've provided you can see as I'm trying to get completions on line 45. It doesn't work, unfortunately.
Is there any way to check logs in nightly build, without building the whole editor from sources?

panekj:
First of all, I would open new issue for that. PRs are not meant to debug specific LSP plugin issues.
Second thing, verify in task manager if you have process for go lsp under lapce.

nheuillet:
@yuriizinets The problem is most likely that you do not appear to have everything in the folder . If plugin.toml is missing then yeah it won't work. I would suggest simply cloning in the plugins folder, then compiling. You should see a "starting lapce-go plugin!" Message appearing after a few seconds

phucngodev:
Hello @nheuillet, I appreciate your effort to add Go support Lapce. I followed your instructions and compiled the lapace-go plugin, put it into ~/.lapce/plugins/lapce-go but I am still cannot get it working. 

When I run lapce from terminal I got the following error message, do you know what I'm missing?
image

I use Lapce nightly build version

nheuillet:
Hello @phucngodev , 
From the look of it, the path to gopls is incorrectly generated. I would need to check more in detail, would you be able to come on the discord so that we can have an easier time debugging? You can find the link on the [website](https://lapce.dev)

phucngodev:
@nheuillet If I launch the Lapce.app (Mac App) it is unable to access environment variables (GOBIN, GOPATH), but when I start the editor from terminal it can find the gopls and the plugin start working. I also noted one more thing is that if I didn't set GOBIN the go_bin_path = ""
```rust
 let go_bin_path = match env::var("GOBIN") {
            Ok(var) => var,  // =====> GOBIN = ""
            Err(error) => match error {
                env::VarError::NotPresent => match env::var("GOPATH") {
                    Ok(var) => format!("{var}/bin"),
                    Err(error) => {
                        panic!("Couldn't get GOPATH: {error}")
                    }
                },
                env::VarError::NotUnicode(val) => {
                    let val = val.to_string_lossy();
                    panic!("GOBIN is not in unicode format: '{val}'")
                }
            },
        };

```

nheuillet:
In that case make sure the `env_command` is in your plugin.toml. it looks like there is some shenanigans around your env variables

psydvl:
I have latest nightly version and built `lapce-go` in both `~/.local/share/lapce-nightly/plugins` and `~/.lapce/plugins`
But nothing on plugins pane or in logs
What should I do to fix it?

ghishadow:
@psydvl `lapce-go` is not ported to new volt api I think, panekj is porting it here https://github.com/panekj/lapce-go/tree/volt

PR #557, opened 2022-05-27
added basic disambiguation of same name files

### LINKED ISSUE #551: Disambiguate files with the same name
Currently, if opening two files with the same name (but obviously a different location) there is nothing to tell which one is which if you are not keeping track of their position.  
VSCode does this by adding part of their path to their titles, 
image

nheuillet:
This PR introduces some visual syntactic sugar around same name files from different folders.

* Every path created is truncated to 20 characters, which is completely arbitrary and can be discussed/changed*
- When two same-name files are open, a path hint will appear for as long as multiple same name files are opened
- When two files from the same workspace are open, the names of the folders (full path minus workspace root) are used
- When a file from outside of the workspace is opened, the full path is used
- When a file at the workspace root is opened, `./` is used

The implementation is not as efficient as it could be (some clones and a lot of conversions), but I am no Rust expert, so feel free to show me how to improve :)

closes #551

### DIFF
--- a/lapce-ui/src/editor/tab_header_content.rs
+++ b/lapce-ui/src/editor/tab_header_content.rs
@@ -258,14 +258,14 @@ impl LapceEditorTabHeaderContent {
         let workspace_path = match workspace_path {
             Some(p) => p,
             None => {
-                let mut file_path_str = match file_path.to_str() {
-                    Some(p) => p.to_string(),
-                    None => return "".to_string(),
-                };
-
-                file_path_str.truncate(20);
-
-                return file_path_str;
+                // file_path.truncate(20);
+                let file_path = file_path
+                    .to_string_lossy()
+                    .char_indices()
+                    .filter(|(i, _)| *i < 20)
+                    .map(|(_, ch)| ch)
+                    .collect::<String>();
+                return file_path;
             }
         };
 
@@ -278,7 +278,7 @@ impl LapceEditorTabHeaderContent {
             // We dont need to keep the file name, only the parents.
             file_path_mut.pop();
 
-            while file_path_mut.cmp(&workspace_path) != Ordering::Equal {
+            while file_path_mut != workspace_path {
                 let file_name = file_path_mut.file_name().unwrap();
                 let file_name_str = file_name.to_str().unwrap().to_string();
 
@@ -575,6 +575,15 @@ mod test {
         let f4 = PathBuf::from("/home/user/proj/file.rs");
         let f5 = PathBuf::from("/home/user/toolongprojectshouldtruncate/file.rs");
         let f6 = PathBuf::from("/home/user/myproject/file.rs");
+        let f7 = PathBuf::from(
+            "/home/user/myproject/🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆",
+        );
+        let f8 = PathBuf::from(
+            "/home/user/proj/🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆",
+        );
+        let f9 = PathBuf::from(
+            "/home/user/myproject/🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆/🎉🎉🎉🎉🎉🎉🎉/🎊",
+        );
 
         let mut tab = LapceEditorTabHeaderContent::new(WidgetId::next());
 
@@ -584,6 +593,9 @@ mod test {
         let r4 = tab.get_effective_path(workspace_path.clone(), &f4);
         let r5 = tab.get_effective_path(workspace_path.clone(), &f5);
         let r6 = tab.get_effective_path(workspace_path.clone(), &f6);
+        let r7 = tab.get_effective_path(workspace_path.clone(), &f7);
+        let r8 = tab.get_effective_path(workspace_path.clone(), &f8);
+        let r9 = tab.get_effective_path(workspace_path.clone(), &f9);
 
         assert_eq!(r1, "folder1");
         assert_eq!(r2, "folder2");
@@ -591,5 +603,8 @@ mod test {
         assert_eq!(r4, "/home/user/proj");
         assert_eq!(r5, "/home/user");
         assert_eq!(r6, "./");
+        assert_eq!(r7, "./");
+        assert_eq!(r8, "/home/user/proj");
+        assert_eq!(r9, "🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆🎆/🎉🎉🎉🎉🎉🎉🎉");
     }
 }
--- a/lapce-ui/src/editor/tab_header_content.rs
+++ b/lapce-ui/src/editor/tab_header_content.rs
@@ -1,12 +1,13 @@
-use std::{iter::Iterator, sync::Arc};
+use std::{cmp::Ordering, iter::Iterator, path::PathBuf, sync::Arc};
 
 use druid::{
     kurbo::Line,
     piet::{Text, TextLayout as TextLayoutTrait, TextLayoutBuilder},
-    BoxConstraints, Command, Env, Event, EventCtx, LayoutCtx, LifeCycle,
+    BoxConstraints, Command, Env, Event, EventCtx, FontStyle, LayoutCtx, LifeCycle,
     LifeCycleCtx, MouseButton, MouseEvent, PaintCtx, Point, RenderContext, Size,
     Target, UpdateCtx, Widget, WidgetId,
 };
+use hashbrown::HashMap;
 use lapce_core::command::FocusCommand;
 use lapce_data::{
     command::{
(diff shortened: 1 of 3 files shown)


lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -248,6 +249,79 @@ impl LapceEditorTabHeaderContent {
...
+
+    fn get_effective_path(
+        &mut self,
+        workspace_path: Option<PathBuf>,
+        file_path: &PathBuf,
+    ) -> String {
+        let workspace_path = match workspace_path {
+            Some(p) => p,
+            None => {
+                let mut file_path_str = match file_path.to_str() {
```
MinusGix, line 261: Minor: `file_path.to_str().unwrap_or("")` (then call `.to_string` on that)

MinusGix: You could also just use https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.to_string_lossy

MinusGix: Minor: In Rust, shadowing names (so using file_path for the String version too) is normal when they are not needed anymore (like your converting it into a String).

lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -248,6 +249,79 @@ impl LapceEditorTabHeaderContent {
...
+                    Some(p) => p.to_string(),
+                    None => return "".to_string(),
+                };
+
+                file_path_str.truncate(20);
```
MinusGix, line 266: This has potential issue if the name contains special characters, since 20 might lie in the middle of a multibyte character which would panic.  
I add some issue in finding if there was a better solution than truncate for this task, unfortunately.  
```rust
let file_path_str = file_path_str.char_indices()
        .filter(|(i, _)| *i < 20)
        .map(|(_, ch)| ch)
        .collect::<String>();
```
Ugly, and I feel like there should be a better way, but I believe it should work.

lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -248,6 +249,79 @@ impl LapceEditorTabHeaderContent {
...
+        // file is a workspace file: we can truncate to workspace folders
+        if file_path.starts_with(&workspace_path) {
+            let mut reversed_path: Vec<String> = Vec::<String>::new();
+            // let mut parent = file_path.parent();
+            let mut file_path_mut = file_path.clone();
+
+            // We dont need to keep the file name, only the parents.
+            file_path_mut.pop();
+
+            while file_path_mut.cmp(&workspace_path) != Ordering::Equal {
```
MinusGix, line 281: I think this can just be `while file_path_mut != workspace_path {`?

lapce-ui/src/editor/tab_header_content.rs
nheuillet: It looks like it works like a charm yeah! I'll add to test for unicode-based names to avoid regressions.

lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -351,9 +425,31 @@ impl Widget<LapceTabData> for LapceEditorTabHeaderContent {
...
+            match child {
+                EditorTabChild::Editor(view_id, _, _) => {
+                    let editor = data.main_split.editors.get(view_id).unwrap();
+                    if let BufferContent::File(path) = &editor.content {
+                        if let Some(file_name) = path.file_name() {
+                            let name = file_name.to_str().unwrap();
+                            let nb = match name_count.get(name) {
+                                Some(name) => *name,
+                                None => 0,
+                            };
```
MinusGix, line 439: Minor: `name_count.get(name).copied().unwrap_or(0)`, which copies the held reference to an integer value and then unwraps to it or 0.

lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -362,6 +458,18 @@ impl Widget<LapceTabData> for LapceEditorTabHeaderContent {
...
+                                let nb = name_count
+                                    .get(file_name.to_str().unwrap())
+                                    .unwrap();
+                                if *nb > 1 {
+                                    file_path = format!(
+                                        " {}",
+                                        self.get_effective_path(
+                                            data.workspace.path.clone(),
+                                            path
+                                        )
```
MinusGix, line 470: `get_effective_path` could take an `Option<&PathBuf>` to avoid unneeded cloning. (`data.workspace.path.as_ref()` to get a reference into the option)

MinusGix:
The file-counting logic should maybe go in `update` instead of `layout`, and it should be a field on data somewhere so that it can be reused. However, I think we can avoid having the hashmap and string allocations at all.  
Most uses of the editor would have significantly less than even 100 editors open and so just iterating over each of them to check if they have a duplicate filename seems like it would be better (compared to allocating a bunch of strings).  
Ex:  
```rust
for (i, child) in editor_tab.children.iter().enumerate() {
    // the variables defined at the start
    match child {
        EditorTabChild::Editor(view_id, _, _) => {
            let editor = data.main_split.editors.get(view_id).unwrap();
            if let Some(file_name) = path.file_name() {
                if has_other_editors_with_same_name(data, editor_tab, i, file_name) {
                    // though, file_path could perhaps be an Option<String>?
                    file_path = format!(" {}", self.get_effective_path(data.workspace.path.clone()));
                }
            }
        }
    }
}

// define some separate helper function if you want
/// Checks if there is another editor with the same file name
fn has_other_editors_with_same_name(data: &LapceTabData, editor_tab: &LapceEditorTabData, index: usize, file_name: &OsStr) -> bool {
    for (i, child) in editor_tab.children.iter().enumerate() {
        // Skip past itself
        if i == index {
            continue;
        }
        
        match child {
            EditorTabChild::Editor(view_id, _, _) => {
                let editor = data.main_split.editors.get(view_id).unwrap();
                if let BufferContent::File(path) = &editor.content {
                    if let Some(child_filename) = path.file_name() {
                        if child_filename == file_name {
                            // There is a copy
                            return true;
                        }
                    }
                }
            }
            EditorTabChild::Settings(_, _) => {}
        }
    }
    
    // There was no copy
    false
}
```
(Note: The above is untested, I wrote it out from looking at the existing code)

lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -248,6 +249,79 @@ impl LapceEditorTabHeaderContent {
...
+        file_path: &PathBuf,
+    ) -> String {
+        let workspace_path = match workspace_path {
+            Some(p) => p,
+            None => {
+                // file_path.truncate(20);
+                let file_path = file_path
+                    .to_string_lossy()
+                    .char_indices()
+                    .filter(|(i, _)| *i < 20)
```
bugadani, line 265: Are you looking for `.chars().take(20)`? :)

lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -248,6 +249,79 @@ impl LapceEditorTabHeaderContent {
...
+                let file_name_str = file_name.to_str().unwrap().to_string();
+
+                reversed_path.push(file_name_str);
+                if !file_path_mut.pop() {
+                    break;
+                }
+            }
+
+            // File is at workspace root
+            if reversed_path.len() == 0 {
```
bugadani, line 292: Clippy will complain here, you can use `.is_empty()` instead

lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -248,6 +249,79 @@ impl LapceEditorTabHeaderContent {
             ));
         }
     }
+
+    fn get_effective_path(
+        &mut self,
+        workspace_path: Option<PathBuf>,
+        file_path: &PathBuf,
```
bugadani, line 256: Taking a `&Path` would be more idiomatic.

lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -248,6 +249,79 @@ impl LapceEditorTabHeaderContent {
...
+                    .char_indices()
+                    .filter(|(i, _)| *i < 20)
+                    .map(|(_, ch)| ch)
+                    .collect::<String>();
+                return file_path;
+            }
+        };
+
+        // file is a workspace file: we can truncate to workspace folders
+        if file_path.starts_with(&workspace_path) {
```
bugadani, line 273: I haven't given this much thought, really, but if you want to trim the workspace prefix, you can do that with `str`'s [`strip_prefix`](https://doc.rust-lang.org/std/primitive.str.html#method.strip_prefix). I think the problems panel implementation does this already if you want an example.

lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -351,9 +425,31 @@ impl Widget<LapceTabData> for LapceEditorTabHeaderContent {
 
         self.rects.clear();
         let mut x = 0.0;
+        let mut name_count: HashMap<String, i32> = HashMap::new();
+        for (_i, child) in editor_tab.children.iter().enumerate() {
```
bugadani, line 429: Why do you `enumerate` if you ignore the index afterwards?

lapce-ui/src/editor/tab_header_content.rs
bugadani: I think we want to use the entry API here. That would avoid hashing the name twice.

Also, "the number equals the name" is hella confusing :) Using the same variable name to refer to a different thing basically one line away isn't very clear.

lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -351,9 +425,31 @@ impl Widget<LapceTabData> for LapceEditorTabHeaderContent {
...
+            match child {
```
bugadani, line 430: This would be better as an `if let` since you only handle a single branch.

lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -384,7 +492,19 @@ impl Widget<LapceTabData> for LapceEditorTabHeaderContent {
                 )
                 .build()
                 .unwrap();
-            let text_size = text_layout.size();
+            let path_layout = ctx
```
bugadani, line 495: I don't see clearly on Github's UI, but you might want to check if `file_path` is empty here and skip this whole thing in that case. `TabRect` could have an optional `path_layout` to make it work.

lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -362,6 +458,18 @@ impl Widget<LapceTabData> for LapceEditorTabHeaderContent {
                         if let Some(file_name) = path.file_name() {
                             if let Some(s) = file_name.to_str() {
                                 text = s.to_string();
+                                let nb = name_count
+                                    .get(file_name.to_str().unwrap())
```
bugadani, line 462: `s` is already this value, isn't is?

lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -351,9 +425,31 @@ impl Widget<LapceTabData> for LapceEditorTabHeaderContent {
 
         self.rects.clear();
         let mut x = 0.0;
+        let mut name_count: HashMap<String, i32> = HashMap::new();
```
bugadani, line 428: I think building this hashmap could have a better place in `update`. In general, you want to do it when the data changes, not when the UI changes.

bugadani:
> The file-counting logic should maybe go in update instead of layout

I should have read the comments before adding my own. Sorry about that :')

lapce-ui/src/editor/tab_header_content.rs
MinusGix: Yeah that's better, forgot about `take` when I suggested it.

lapce-ui/src/editor/tab_header_content.rs
nheuillet: I do want to trim the workspace prefix (make it so that the path shown will be relative inside the workspace). Nevertheless, using `strip_prefix` involves some normalization issues (trailing slashes when popping the file, etc.), so I would say using pathbuff would be the easiest way to handle this case. There might be a more efficient way though, and I will try to look into it

bugadani:
Wait, I have reviewed this PR already? Can we move forward on this?

nheuillet:
@bugadani there was some comments about the way to store and handle / compute the data that I never got around to resolve in the end. If someone wants to take care of it then by all means. Otherwise, I'll revisit and complete this PR in a few weeks, as I want to finish PSP first.

hbina:
Let me give it a shot.

nheuillet:
> Let me give it a shot.

Gladly! thanks a lot @hbina

hbina:
Do i need to create a new PR or can I reuse this somehow?

panekj:
> Do i need to create a new PR or can I reuse this somehow?

create new PR

nheuillet:
@hbina Create a new PR but feel free to copy whatever you need from this PR :)

nheuillet:
closing as #1521 got merged

Issue #548, opened 2022-05-25
Saving File Does Not Keep Permissions


MinusGix:
When saving a .sh script file (after modifying it) that was marked as executable, it loses the executable bit/permission.  
It would be better not to mess with these at all. Are we overwriting the file rather than opening it and then writing out the data?

dzhou121:
We save it to a temp file first and copy it over. We'll need to duplicate the file permissions for the temp file.

Issue #382, opened 2022-04-06
Execution permission on file after save


thomasmarcel:
On a Mac with MacOS Monterey,

A script file with permissions `-rwxr-xr-x` (755) have its permissions changed to `-rw-r--r--` (644) after modifying it and saving it.

The permissions on the file shouldn't be modified on save.

PR #564, opened 2022-05-30
fix: don't overwrite permissions

panekj:
Currently lapce overwrites permissions whenever file is saved
I've noticed that when my scripts lost exec bits
Don't know if it's proper fix in correct place


MinusGix:
Fixes #548

panekj:
fixes #382

dzhou121:
Thanks!

Issue #572, opened 2022-06-03
nightly cargo needed


holtzermann17:
```
joe@Proteus:~/lapce$ cargo build --release
error: failed to parse manifest at `/home/joe/lapce/Cargo.toml`

Caused by:
  feature `named-profiles` is required

  this Cargo does not support nightly features, but if you
  switch to nightly channel you can add
  `cargo-features = ["named-profiles"]` to enable this feature
```

Fixed by `rustup toolchain install nightly && rustup default nightly` and rerun above.

informaticalex:
Custom `named-profiles` are a feature of [Cargo version 1.57](https://doc.rust-lang.org/cargo/reference/unstable.html?highlight=named-profile#custom-named-profiles). Regardless of whether a custom profile is specified, building with Cargo < 1.57 will show the above error.

As mentioned, upgrading Cargo to nightly, or 1.57+, fixes this.

Tested on [Fedora Workstation 36](https://getfedora.org/) using [this method](https://users.rust-lang.org/t/how-can-i-revert-to-older-version-of-rust/20497) to downgrade Cargo/Rust.

informaticalex:
Related to the above, Cargo version 1.57-1.59 will have a different error:

```
error: failed to load manifest for workspace member `C~\lapce\lapce-ui`

Caused by:
  failed to load manifest for dependency `lapce-core`

Caused by:
  failed to parse manifest at `C:\Users\mieza\lapce\lapce-core\Cargo.toml`

Caused by:
  namespaced features with the `dep:` prefix are only allowed on the nightly channel and requires the `-Z namespaced-features` flag on the command-line
```

Lapce is using _optional_ dependencies, a feature implemented in [Cargo 1.60](https://github.com/rust-lang/cargo/blob/master/CHANGELOG.md#cargo-160-2022-04-07).

Cargo must be version 1.60+ for lapce to build.

panekj:
Fixed in https://github.com/lapce/lapce/pull/573

PR #573, opened 2022-06-04
Indicate minimum toolchain version preventing #572

informaticalex:
Suggest closing #572.

(also added single whitespace docs)


panekj:
At least Rust 1.60 is required

informaticalex:
@panekj Yes, indeed. 

Build will fail with a different error to #572, but I've documented this within the issue and updated the PR.

dzhou121:
Thanks!

PR #579, opened 2022-06-05
Implement event-subscription API for plugins

MinusGix:
Related: https://github.com/lapce/lapce-plugin-rust/pull/2  
This implements a basic event system, where plugins can subscribe to specific `PluginEventKind`s and then receive them whenever they occur.
A plugin can subscribe to events even after initialization, at least at the moment. This seems useful to me, since you may only want to subscribe to events that you are handling, and perhaps a plugin only cares about certain events for specific projects or settings.  
This does not add a method for unsubscribing from events, but that could be added without issue.  
  
This pull-request does not actually implement all that much, leaving other specific features of the plugin system for later pull-requests, but would hopefully serve as the foundation to begin adding plugin functionality. The basic events added can serve as examples, too.  
  
Current questions/open problems:
 - This adds a debug_log notification, though the companion PR for lapce-plugin-rust does not add a function for calling it directly. I think there should be some way of logging from the extension, though perhaps it should go to the log file, but I don't think lapce-proxy has access to that?
 - I dislike serializing the `PluginEvent` to the rpc, then deserializing it, then reserializing it for the plugin
   - Possible methods to avoid this:
     - Synchronize the subscribed events between the proxy and lapce, so that lapce can check that some plugin actually wants the event (which avoids making this feature somewhat costly for users without notable plugins)
       - This avoids unneeded serialization in the first place, which is nice
     - Rather than deserializing it, get the field in the json for the name and get the `PluginEventKind` via that
   - I'll put this off for now
 - I'd like to share types between lapce and lapce-plugin-rust, since the `PluginEvent` and `PluginEventKind` should be the same
   - Lapce could depend directly on lapce-plugin-rust (putting plugin/wasm specific things behind a feature flag that is by default on, so we don't have to depend on them?)
 - `PluginEvent::FileEditorClosed` has the path stored as a `PathBuf`, but is this safe to transmit over JSON (since it is backed by an `OsString`) and deserialize on a different platform (proxy host os -> wasm)?
 - We may want events that are broadcast to all plugins regardless of whether they subscribe
   - Ex: Shutdown
   - This is pretty easy to add, if needed
 - `PluginDesc` is cloned several times (since `PluginEnv`/`PluginNew` are cloned to move instances into other threads), but we could avoid this with an `Arc` around it. Should I implement that?
 - `broadcast_event` could be improved by serializing the event to a string once, and then giving out a `&str` to a `PluginNew::send_event_str` function.
   - However, this has the issue of being able to pass it into the thread due to lifetimes. Is it reasonable to do `wasi_write_string` before starting the thread?

### DIFF
--- a/lapce-data/src/proxy.rs
+++ b/lapce-data/src/proxy.rs
@@ -13,6 +13,7 @@ use druid::Target;
 use druid::{ExtEventSink, WidgetId};
 use flate2::read::GzDecoder;
 use lapce_proxy::dispatch::Dispatcher;
+use lapce_proxy::plugin::PluginEvent;
 use lapce_rpc::buffer::BufferId;
 use lapce_rpc::core::{CoreNotification, CoreRequest};
 use lapce_rpc::plugin::PluginDescription;
@@ -398,6 +399,12 @@ impl LapceProxy {
     pub fn install_plugin(&self, plugin: &PluginDescription) {
         self.rpc
             .send_rpc_notification("install_plugin", &json!({ "plugin": plugin }));
+    }
+
+    /// Broadcast an event to plugins which are subscribed to it
+    pub fn plugin_broadcast(&self, event: PluginEvent) {
+        self.rpc
+            .send_rpc_notification("plugin_broadcast", &json!({ "event": event }));
     }
 
     pub fn get_buffer_head(
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2223,6 +2223,7 @@ dependencies = [
  "jsonrpc-lite",
  "lapce-core",
  "lapce-data",
+ "lapce-proxy",
  "lapce-rpc",
  "lazy_static",
  "libloading",
--- a/lapce-proxy/src/dispatch.rs
+++ b/lapce-proxy/src/dispatch.rs
@@ -1,6 +1,6 @@
 use crate::buffer::{get_mod_time, load_file, Buffer};
 use crate::lsp::LspCatalog;
-use crate::plugin::PluginCatalog;
+use crate::plugin::{PluginCatalog, PluginEvent};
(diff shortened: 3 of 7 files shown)


MinusGix:
> Thanks for making a great effort on plugins! It might take some time for me to think about the architecture if that's all right.

Yeah, that's fine!

dzhou121:
>> I'd like to share types between lapce and lapce-plugin-rust, since the PluginEvent and PluginEventKind should be the same

Do you think the types could belong to lapce-rpc

MinusGix:
> > > I'd like to share types between lapce and lapce-plugin-rust, since the PluginEvent and PluginEventKind should be the same
> 
> Do you think the types could belong to lapce-rpc

That would work fine I think. (and then the lapce-plugin-rust just depends on lapce-rpc)

lapce-data/src/proxy.rs
```diff
@@ -400,6 +401,12 @@ impl LapceProxy {
             .send_rpc_notification("install_plugin", &json!({ "plugin": plugin }));
     }
 
+    /// Broadcast an event to plugins which are subscribed to it
+    pub fn plugin_broadcast(&self, event: PluginEvent) {
```
dzhou121, line 405: Another thought on this. Between the communication between UI and Proxy, we probably don't make them plugin specific. E.g., ```new_buffer```, ```save_buffer``` would be plugin events at Proxy's layer.

MinusGix: While I agree some events shouldn't be plugin specific, it still seems useful to be able to only send an event to a certain plugin. Ex: A plugin registers a workbench command. Then the proxy/whatever can specifically send the "workbench command used" event just to the plugin that registered it. I think I might be misunderstanding what you mean, though.

PR #588, opened 2022-06-08
switch ureq to reqwest for system proxy support

dzhou121:
Lapce's plugin list is hosted on github pages. Some users need proxy to be able to access github sites. ureq doesn't read system proxy settings, which will cause an empty plugin list for some users. Switch it back to reqwest to fix that.


mrxiaozhuox:
It didn't work on my machine, maybe we can add a manual proxy config?
image

dzhou121:
It might need ```HTTP_RPOXY``` or ```HTTPS_PROXY``` environment variables.

mrxiaozhuox:
oh, yes!
image

dzhou121:
Thanks for testing!

PR #616, opened 2022-06-16
Handle unicode characters in palette line-search

### LINKED ISSUE #614: Searching for german umlauts will crash Lapce in modal mode
# Description
It is not possible to search for german umlauts if modal mode is enalbed.

# Environment
**OS:** Windows 10 (20H2 Build 19042.1526)
**Lapce:** v0.1.2, Windows-Portable

# Steps to reproduce:
- create a blank new file
- add the following content
   ```
    abc  
    üäö  
    jkm
  ```
- press '/'
- type 'ä' or 'ö' or 'ü'
  - Lapce will freeze for a short time and then crash silently

MinusGix:
Fixes #614 
The issue was that it was just having the end of the range `i+1` rather than the length of the character which panics (though the docs for `range_attribute` only indirectly allude to indexing like a slice) on multibyte characters.  
There's still a bit of a bug (unicode characters will hit that `log::warn`) because of fuzzy-matcher's indices not working for unicode, which I mentioned in the discord, but that just stops it from highlighting them properly.

### DIFF
--- a/lapce-ui/src/palette.rs
+++ b/lapce-ui/src/palette.rs
@@ -675,14 +675,31 @@ impl PaletteContent {
                     .get_color_unchecked(LapceTheme::EDITOR_FOREGROUND)
                     .clone(),
             );
-        for i in &text_indices {
-            let i = *i;
+        for &i_start in &text_indices {
+            let i_end = full_text
+                .char_indices()
+                .find(|(i, _)| *i == i_start)
+                .map(|(_, c)| c.len_utf8() + i_start);
+            let i_end = if let Some(i_end) = i_end {
+                i_end
+            } else {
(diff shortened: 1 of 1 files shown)


lapce-ui/src/palette.rs
```diff
@@ -675,14 +675,32 @@ impl PaletteContent {
                     .get_color_unchecked(LapceTheme::EDITOR_FOREGROUND)
                     .clone(),
             );
-        for i in &text_indices {
-            let i = *i;
+        for i_start in &text_indices {
+            let i_start = *i_start;
```
Eraden, line 679:
```suggestion
        for i_start in text_indices.iter().copied() {
```

PR #617, opened 2022-06-16
Add a setting to toggle relative line numbering in modal mode

### LINKED ISSUE #35: Make line numbering scheme user-selectable
image

MinusGix:
Fixed #35  
This just allows the user to change the line numbering in modal mode to the more typical sequential method.  
This leaves the default as relative, like it is now.  
Though, I do think it would be a nice future feature to have this more customizable (ex: relative line mode in insert mode?), but this covers the case where people want it always sequential.


dzhou121:
Thanks!

PR #622, opened 2022-06-17
Added result numbers to the FindBox widget.

Kirottu:
Added the amount of results found for the query into the `FindBox` widget in this fashion:
image

Does not relate to any github issue I can find, but was a thing someone was looking for.

### DIFF
--- a/lapce-ui/src/find.rs
+++ b/lapce-ui/src/find.rs
@@ -1,4 +1,5 @@
 use druid::{
+    piet::{Text, TextLayoutBuilder},
     BoxConstraints, Command, Env, Event, EventCtx, LayoutCtx, LifeCycle,
     LifeCycleCtx, MouseEvent, PaintCtx, Point, Rect, RenderContext, Size, Target,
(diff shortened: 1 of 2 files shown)


lapce-ui/src/find.rs
```diff
@@ -205,6 +234,45 @@ impl Widget<LapceTabData> for FindBox {
...
+        }
+
+        let text_layout = ctx
+            .text()
+            .new_text_layout(if buffer.doc.find.borrow().occurrences().len() > 0 {
+                format!(
+                    "{}/{}",
+                    match index {
+                        Some(index) => format!("{}", index + 1),
+                        None => "?".to_string(),
```
dzhou121, line 254: It looks a bit odd like ```?/215```, maybe we just show ``` 215 found ``` instead?

Kirottu: Hmmm yeah agreed, that would make more sense.

Kirottu: Now it shows `215 results` instead of `?/215`, so I presume this is resolved.

dzhou121:
Thanks!

PR #636, opened 2022-06-18
Add tree-sitter langs Haskell, CodeQL, and Swift

VixieTSQ:
Tested all three already!

### DIFF
--- a/lapce-core/src/language.rs
+++ b/lapce-core/src/language.rs
@@ -132,6 +132,12 @@ pub enum LapceLanguage {
     Java,
     #[cfg(feature = "lang-elm")]
     Elm,
+    #[cfg(feature = "lang-swift")]
+    Swift,
+    #[cfg(feature = "lang-ql")]
+    QL,
+    #[cfg(feature = "lang-haskell")]
+    Haskell,
 }
 
 // NOTE: Elements in the array must be in the same order as the enum variants of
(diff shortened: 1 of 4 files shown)


dzhou121:
Thanks!

lapce-core/src/language.rs
```diff
@@ -340,6 +346,36 @@ const LANGUAGES: &[SyntaxProperties] = &[
...
+    },
+    #[cfg(feature = "lang-haskell")]
+    SyntaxProperties {
+        id: LapceLanguage::Haskell,
+        language: tree_sitter_haskell::language,
+        highlight: tree_sitter_haskell::HIGHLIGHTS_QUERY,
+        comment: "--",
+        indent: "  ",
+        code_lens: (DEFAULT_CODE_LENS_LIST, DEFAULT_CODE_LENS_IGNORE_LIST),
+        extensions: &["hs"],
```
VixieTSQ, line 377: there's another extension here, `lhs` which I believe would require special treatment that our current system does not allow. This is because in `lhs` files, comments are implicit and you must mark code with a special symbol

lapce-core/src/language.rs
```diff
@@ -340,6 +346,36 @@ const LANGUAGES: &[SyntaxProperties] = &[
...
+    },
+    #[cfg(feature = "lang-ql")]
+    SyntaxProperties {
+        id: LapceLanguage::QL,
+        language: tree_sitter_ql::language,
+        highlight: tree_sitter_ql::HIGHLIGHTS_QUERY,
+        comment: "//",
+        indent: "  ",
+        code_lens: (DEFAULT_CODE_LENS_LIST, DEFAULT_CODE_LENS_IGNORE_LIST),
+        extensions: &["ql"],
```
VixieTSQ, line 367: I think `qll` might also be correct? but I'm not entirely sure

PR #639, opened 2022-06-18
Improve the behaviour of Home in non-vim mode

DJMcNab:
Previously, pressing <kbd>Home</kbd> would always return to the first non-indent character

However, a nicer behaviour once you are at that character is to move to the start of the line instead.
This is consistent with the behaviour of vscode.

### DIFF
--- a/lapce-data/src/document.rs
+++ b/lapce-data/src/document.rs
@@ -1319,7 +1319,11 @@ impl Document {
             Movement::FirstNonBlank => {
                 let line = self.buffer.line_of_offset(offset);
                 let new_offset = self.buffer.first_non_blank_character_on_line(line);
-                (new_offset, Some(ColPosition::FirstNonBlank))
+                if offset == new_offset {
+                    (self.buffer.offset_of_line(line), Some(ColPosition::Start))
+                } else {
+                    (new_offset, Some(ColPosition::FirstNonBlank))
+                }
             }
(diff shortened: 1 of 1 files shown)


lapce-data/src/document.rs
```diff
@@ -1319,7 +1319,11 @@ impl Document {
             Movement::FirstNonBlank => {
                 let line = self.buffer.line_of_offset(offset);
                 let new_offset = self.buffer.first_non_blank_character_on_line(line);
-                (new_offset, Some(ColPosition::FirstNonBlank))
+                if offset == new_offset {
```
dzhou121, line 1322: Maybe it should be ```if offset <= new_offset``` ?

DJMcNab: So, the reason for this version is that if your line is
```
<|>    blah
```
and you press home, the correct behaviour is to make your cursor move to:
```
    <|>blah
```
At least, that's what vscode does.
Basically, pressing home should cycle between the two states, defaulting to the first non-whitespace character.

dzhou121: I just tested VSCode,

If the cursor is at the left of the first non-whitespace character, Home gets to the line beginning?

DJMcNab: Ah, great. I'll implement that version then

dzhou121:
Thanks!

PR #653, opened 2022-06-20
Add context menu to file explorer sidebar

### LINKED ISSUE #621: Tree view context menu
A huge blocker for daily driving Lapce for me at the moment is the lack of a tree view context menu. 

Without the ability to create new files at a specific location in my project or being able to rename files quickly, it is really hard to use lapce.

MinusGix:
Closes #621   
This PR adds a context menu to the file explorer sidebar that allows the following actions:   
 - Create a new file
 - Create a new directory
 - Rename a file/directory
 - Move a file/directory to the trash
    
The inputs for what the name should be used are taken in through the palette input. This is simpler to implement than something like vscode's method of editing the filename in the sidebar.  
  
This does however raise further the probable existing issues around files being renamed out from under Lapce, and new files being created. As well, there's more potential now for cross platform issues where P

Notes:  
- Expands the layout height of the `FileExplorerFileList` to take up the full height of the sidebar. This is in order to capture right clicks in the empty space below the files. This would have to be changed around (such as moving right click handling up a 'layer') if there were other areas added to the file explorer sidebar later on, but would not be hard to do.  
- Minor: `get_node_by_index` is renamed `get_node_by_index_mut` and an immutable version is added


dzhou121:
Thanks so much for working on this. 

I think it would be better if the file/folder name is in the sidebar, which gives you a direct visual confirmation of where the file/folder is.

MinusGix:
(It closed and then reopened because I accidentally pushed nothing at first)  

Okay, I reworked the PR to allow naming/renaming the file directly in the sidebar. It isn't perfect and has some issues that should be fixed/improved-upon in the future, but I think it is in a workable state? 
image Could look better with some more work
(Potential features/improvements/issues):
 - Feature: Doesn't show an icon while typing the file's name. VSCode automatically updates the icon when you type the extension even before the rename.
 - Improvement: We should maybe have a limit on what the filename input can take, so people don't accidentally copy&paste a thousand lines of text into the rename box
 - Improvement: Should maybe stop certain characters from being inputted (such as tabs)
 - Improvement: It would look nicer to put the new-file naming box directly after the folders rather than at the top?
 - Improvement: It should scroll the naming box into view if it isn't already
 - Improvement: I think it would be good to make the font it uses match the ui font for consistency with the sidebar
 - Improvement: It would also be nice to decrease the vertical height a bit, which would make it feel less out of place
 - Issue: Refuses file list updates until after the file has been created or renamed
   - This is needed due to using indexes, which could get messed with by updating the children.
     - We can get around this by either switching to not using indices or making sure that the indices are kept valid (such as moving them as needed)
     - but that's more complexity, when I kinda just want to get this PR finished
   - This does mean there is a bug if a file is created whilst naming a file, and so it won't appear.
   - Might be able to work around this by requesting it to reread files?

PR #656, opened 2022-06-20
Add various tree-sitter languages

VixieTSQ:
Partially fixes https://github.com/lapce/lapce/issues/272. 

I haven't tested any of these, and they really should be tested. We could @ some people in the discord server who have the OS testing roles but I doubt a lot of them read that the roles were for testing...

### DIFF
--- a/lapce-core/src/language.rs
+++ b/lapce-core/src/language.rs
@@ -138,6 +138,18 @@ pub enum LapceLanguage {
     QL,
     #[cfg(feature = "lang-haskell")]
     Haskell,
+    #[cfg(feature = "lang-glimmer")]
+    Glimmer,
+    #[cfg(feature = "lang-haxe")]
+    Haxe,
+    #[cfg(feature = "lang-hcl")]
+    HCL,
+    #[cfg(feature = "lang-ocaml")]
+    OCaml,
+    #[cfg(feature = "lang-ocaml")]
+    OCamlInterface,
+    #[cfg(feature = "lang-scss")]
+    SCSS,
 }
 
 // NOTE: Elements in the array must be in the same order as the enum variants of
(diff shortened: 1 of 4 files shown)


lapce-core/src/language.rs
```diff
@@ -376,6 +388,66 @@ const LANGUAGES: &[SyntaxProperties] = &[
         code_lens: (DEFAULT_CODE_LENS_LIST, DEFAULT_CODE_LENS_IGNORE_LIST),
         extensions: &["hs"],
     },
+    #[cfg(feature = "lang-glimmer")]
+    SyntaxProperties {
+        id: LapceLanguage::Glimmer,
+        language: tree_sitter_glimmer::language,
+        highlight: tree_sitter_glimmer::HIGHLIGHTS_QUERY,
+        comment: "{{!",
```
VixieTSQ, line 396: Glimmer comments in reality look like `{{! text here }}` but our current system here doesn't allow for comments like that.

lapce-core/src/language.rs
```diff
@@ -376,6 +388,66 @@ const LANGUAGES: &[SyntaxProperties] = &[
...
+        indent: "  ",
+        code_lens: (DEFAULT_CODE_LENS_LIST, DEFAULT_CODE_LENS_IGNORE_LIST),
+        extensions: &["hcl"],
+    },
+    #[cfg(feature = "lang-ocaml")]
+    SyntaxProperties {
+        id: LapceLanguage::OCaml,
+        language: tree_sitter_ocaml::language_ocaml,
+        highlight: tree_sitter_ocaml::HIGHLIGHTS_QUERY,
+        comment: "(*",
```
VixieTSQ, line 426: Ocaml comments in reality look like `(* text here *)` but our current system here doesn't allow for comments like that.

lapce-core/src/language.rs
```diff
@@ -376,6 +388,66 @@ const LANGUAGES: &[SyntaxProperties] = &[
...
+        indent: "  ",
+        code_lens: (DEFAULT_CODE_LENS_LIST, DEFAULT_CODE_LENS_IGNORE_LIST),
+        extensions: &["mli"],
+    },
+    #[cfg(feature = "lang-scss")]
+    SyntaxProperties {
+        id: LapceLanguage::SCSS,
+        language: tree_sitter_scss::language,
+        highlight: tree_sitter_scss::HIGHLIGHTS_QUERY,
+        comment: "//",
```
VixieTSQ, line 446: SCSS also has support for comments like `/* text here */` but our current system doesn't allow for multiple comment types.

VixieTSQ (commented):
Comments really need to be reworked for Lapce. We could just let that field take a closure that returns string slices per a given string.

lapce-core/src/language.rs
VixieTSQ: > Comments really need to be reworked for Lapce. We could just let that field take a closure that returns string slices per a given string.

Now that I think about, this comment field is probably used for automatically commenting things with a shortcut, isn't it? In that case, one comment only makes sense but still isn't preferable honestly, we might wanna have a thing where it uses one comment type for multi line comments and vice versa. But all of this still is a problem with languages that have comments with two symbols on the sides of your text.

VixieTSQ:
> Comments really need to be reworked for Lapce. We could just let that field take a closure that returns string slices per a given string.

Now that I think about, this comment field is probably used for automatically commenting things with a shortcut, isn't it? In that case, one comment only makes sense but still isn't preferable honestly, we might wanna have a thing where it uses one comment type for multi line comments and vice versa. But all of this still is a problem with languages that have comments with two symbols on the sides of your text.

dzhou121:
> > Comments really need to be reworked for Lapce. We could just let that field take a closure that returns string slices per a given string.
> 
> Now that I think about, this comment field is probably used for automatically commenting things with a shortcut, isn't it? In that case, one comment only makes sense but still isn't preferable honestly, we might wanna have a thing where it uses one comment type for multi line comments and vice versa. But all of this still is a problem with languages that have comments with two symbols on the sides of your text.

We could create an issue for supporting multi-line comment.

bugadani:
Please don't mention "fixes #issue" in the OP unless the PR resolves the complete issue.

VixieTSQ:
> Please don't mention "fixes #issue" in the OP unless the PR resolves the complete issue.

I didn't know github does that, sorry!

PR #665, opened 2022-06-22
add python language_id

dholth:


dholth (commented):
testing with lapce-python plugin; does not appear to send any commands to python language server, except "show me your capabilities"

dholth:
Is there a reason not to go ahead and put all the Microsoft-recommended language id's in here

PR #691, opened 2022-06-26
Add button for committing to Git

### LINKED ISSUE #686: Add source control commit button
It would be useful to have a button to apply a commit in the git pane. Currently one has to bind a key to the commit command to actually apply it.

MinusGix:
Fixed #686   
This adds a button to make it easier to commit to Git, without having to bind a key to the source_control.commit command.
image
  
Issue: Due to the way panels work, if you change theme then it won't update the button's colors properly. Is there a good way to tell it to recreate the panel? I could create a new button widget that uses the theme's values, but that feels excessive.  
We could use Druid's existing [Key](https://docs.rs/druid/0.7.0/druid/struct.Key.html) to store the theme colors in, and then have change them just update the keys?


dzhou121:
You can create the text layout in paint.

MinusGix:
Yes, but that would require recreating the `Button` widget to allow it to have a TextLayout instance directly? I might be missing something.

MinusGix:
Added the druid code for the `Button` widget, which will also let us add customizations to it in the future. This fixes the issue of it not properly updating with the theme.

PR #710, opened 2022-06-30
feat: implement custom titlebar matching editor style

panekj:


dzhou121:
If possible, maybe we should draw the "minimize", "maximise", and "close" icons like a normal windows APP with title.

dzhou121:
image

something like this

panekj:
Normal window:
image
Maximised window:
image
Full titlebar:
image


@dzhou121 few things:

- [ ] do we want lapce logo in titlebar? current position is ok? same for colour? which platforms will have logo? for example VSCode has logo on Windows, but not on macOS due to window controls on left side

- [ ] do we want title? what should title have? should title be same as "os" title (currently it's workspace/folder name)

- [ ] do we want window controls on windows to match perfectly? standard Windows window controls have big padding between each other. they also look a bit blurred(?)
image


Example of draggable lapce:

https://user-images.githubusercontent.com/50457605/176743017-73656470-f7fa-4fda-9246-50dc7e57a162.mp4

dzhou121:
>do we want lapce logo in titlebar? current position is ok? same for colour? which platforms will have logo? for example VSCode has logo on Windows, but not on macOS due to window controls on left side

Lapce logo only for Windows.

>do we want title? what should title have? should title be same as "os" title (currently it's workspace/folder name)

Probably yes for Windows only. 

>do we want window controls on windows to match perfectly? standard Windows window controls have big padding between each other. they also look a bit blurred(?)

Yes please if that's not too much of trouble.

panekj:
image

panekj:
The only thing left is to add dynamic title to show workspace name but I haven't figured that out yet. PR can be merged and I'll push fix for title later.
Also, restoring window currently doesn't work from window control, it works on double click on titlebar though.

dzhou121:
Thanks!

I'm perfectly happy with it for not showing project name, because it's already at the very left. 

The restore thing might be a Druid thing. I can look into that.

lapce-ui/src/app.rs
```diff
@@ -83,6 +83,15 @@ where
         .title(LocalizedString::new("Lapce").with_placeholder("Lapce"))
         .window_size(size)
         .set_position(pos);
+
+    #[cfg(target_os = "windows")]
+    if Config::load(&lapce_data::data::LapceWorkspace::default())
```
dzhou121, line 88: You can pass the Config in from the parameter.

lapce-ui/src/title.rs
```diff
@@ -97,7 +105,19 @@ impl Widget<LapceWindowData> for Title {
                     )
                 }
             }
-            _ => {}
+            _ => {
+                #[cfg(target_os = "windows")]
+                {
+                    ctx.submit_command(
```
dzhou121, line 111: Is this for when the ```custom_titlebar``` is toggled in settings? It's better to put in Update and run this when the settings is actually changed.

panekj: moved to `update`

lapce-ui/src/app.rs
panekj: done

lapce-ui/src/title.rs
```diff
@@ -126,7 +148,15 @@ impl Widget<LapceWindowData> for Title {
...
-        Size::new(bc.max().width, 28.0)
+        #[cfg(not(target_os = "windows"))]
+        {
+            Size::new(bc.max().width, 28.0)
+        }
+
+        #[cfg(target_os = "windows")]
+        {
+            Size::new(bc.max().width, 32.0)
+        }
```
panekj, line 159: Changed titlebar height on Windows to match most other apps and improve readability slightly

PR #715, opened 2022-07-01
Implement inlay hints

MinusGix:
image
This PR implements inlay hints.  
Question: Where is the best place to initialize inlay hints on open documents? Currently it waits until you press a key, but that's not the best time to do it. Adding the code for when an editor-view is opened could work, but would also want it to work on editor views that are open before the LSP begins initializing?  
Notes:  
 - This does not implement a custom background color for inlay hints. The code provides an iterator so it would be relatively easy to implement later.
 - Does not implement commands that are activated by clicking on the inlay hints; does not implement tooltips for inlay hints
 - Adds `smallvec` as a direct dependency. This makes the code nicer and lets us avoid extra allocs in an area where speed is nice.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2279,6 +2279,7 @@ dependencies = [
  "serde 1.0.137",
  "serde_json",
  "sled",
+ "smallvec",
  "structdesc",
  "strum 0.24.1",
  "strum_macros 0.24.0",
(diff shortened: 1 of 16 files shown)


follower:
I've found this a _super_ great addition when doing Rust development, it really helped me (still pretty early days with my Rust journey) with being more aware of the current context of the code I'm working on.

When I saw this hit nightly I grabbed a nightly & was glad I didn't wait 'til the release. :)

So, thanks for your efforts to add it! :)

PR #733, opened 2022-07-03
Add more icons

wkaelj:
More icon support
 - CMake
 - Make
 - C/C++ header files
 - Cargo files
 - Git files (.gitignore, .gitmodules)


lapce-ui/src/svg.rs
```diff
@@ -49,15 +49,33 @@ pub fn get_svg(name: &'static str) -> Option<Svg> {
...
-    let file_type = if path.file_name().and_then(OsStr::to_str) == Some("LICENSE") {
+    
+    // catch if path.to_str() is Some without calling it twice
+    let path_str = path.to_str();
+    if path_str == None {
+        panic!("Missing path");
+    }
+    let file_name = path_str.unwrap(); // unwrap path_str safely now
+
+    let file_type = if file_name == "LICENSE" {
```
wkaelj, line 60: Should be file_name.contains("LICENSE")

Shasta84 (approved):

lapce-ui/src/svg.rs
```diff
@@ -49,15 +49,33 @@ pub fn get_svg(name: &'static str) -> Option<Svg> {
...
+    // catch if path.to_str() is Some without calling it twice
+    let path_str = path.to_str();
+    if path_str == None {
+        panic!("Missing path");
+    }
+    let file_name = path_str.unwrap(); // unwrap path_str safely now
+
+    let file_type = if file_name.contains("LICENSE") {
         "file_type_license.svg"
+    } else if file_name.to_lowercase().contains("makefile") {
```
bugadani, line 62: If you're still willing to work on this PR, please extract the `.to_lowercase()` conversion. In fact, if you could use some sort of case insensitive matching here, that would be best. Thanks :)

lapce-ui/src/svg.rs
```diff
@@ -49,15 +49,33 @@ pub fn get_svg(name: &'static str) -> Option<Svg> {
 }
 
 pub fn file_svg(path: &Path) -> Svg {
-    let file_type = if path.file_name().and_then(OsStr::to_str) == Some("LICENSE") {
+    
+    // catch if path.to_str() is Some without calling it twice
+    let path_str = path.to_str();
+    if path_str == None {
```
bugadani, line 55: This isn't really necessary. `path_str.expect("Missing path");` should be enough in place of the unwrap below.

lapce-ui/src/svg.rs
```diff
@@ -71,12 +89,15 @@ pub fn file_svg(path: &Path) -> Svg {
...
+                    (&["svg"], "file_type_svg.svg"),
+                    (&["cmake"], "file_type_cmake.svg"),
+                    (&["make"], "file_type_make.svg"),
                 ];
 
                 for (exts, file_type) in TYPES {
                     for ext in exts.iter() {
                         if extension.eq_ignore_ascii_case(ext) {
-                            return Some(*file_type);
+                            Some(*file_type);
```
bugadani, line 100: Why remove the return?

lapce-ui/src/svg.rs
```diff
@@ -85,8 +106,13 @@ pub fn file_svg(path: &Path) -> Svg {
             })
             .unwrap_or("default_file.svg")
     };
-
-    get_svg(file_type).unwrap()
+    
+    let svg_file = get_svg(file_type);
+    if svg_file.is_none() {
+        panic!("Missing svg icon '{}'", file_type);
```
bugadani, line 112: Same as above: prefer `.expect()` (or in this case, `unwrap_or_else(|| panic!())`) over an `if`

lapce-ui/src/svg.rs
wkaelj: I got a issue about it when I submit the pr the first time. Something about it not following the style guide.

lapce-ui/src/svg.rs
wkaelj: changed it to .eq_ignore_ascii_case

panekj (changes requested):
I'm still against adding technology specific icons due to licence constraint

bugadani:
Good point. Also I would prefer landing #768 first, if ever.

wkaelj:
Icons not wanted

PR #738, opened 2022-07-04
feat: parse commandline args

### LINKED ISSUE #157: Basic command line options
Currently there are no command line options.
It would be nice to have basic options like:
`lapce --version`
`lapce --help`

And a positional argument to open files from command line.
`lapce myfile.c`
Or open a directory
```
lapce .
lapce /home/user/myprojects
```

### LINKED ISSUE #734: Lapce doesn't open via WSL
I was able to build the project from source on wsl. But when I tried to run ```./lapce``` from ```/target/release``` , it gave me the following error

```Unable to init server: Could not connect: Connection refused
(lapce:1627): Gtk-WARNING **: 11:33:25.709: cannot open display:```

### PR DESCRIPTION
**Pull request by panekj**
fixes #734 
fixes #157 

I decided to not add any arg handling library as we are already
heavy on deps and I don't expect much cli options for now

TODO:
- [ ] support for single files as cli arg
- [x] turns out it wasn't working because `\\wsl.localhost` doesn't exist
  - [x] terminal workdir should be same as path from args
  - [x] indicate that it's opened in WSL

### DIFF
--- a/lapce-ui/src/app.rs
+++ b/lapce-ui/src/app.rs
@@ -11,6 +11,7 @@ use lapce_data::{
     config::Config,
     data::{LapceData, LapceWindowData, LapceWindowLens},
     db::{TabsInfo, WindowInfo},
+    proxy::VERSION,
 };
 
 use crate::logging::override_log_levels;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: lapce-ui/src/app.rs [hunk 1]
```diff
@@ -26,6 +27,32 @@ pub fn build_window(data: &mut LapceWindowData) -> impl Widget<LapceData> {
...
+    if args.len() > 1 {
+        args.next();
+        for arg in args {
+            match arg.as_str() {
+                "-v" | "--version" => {
+                    println!("lapce v{VERSION}");
+                    return;
+                }
+                "-h" | "--help" => {
+                    println!("lapce [-h|-help] [-v|--version] [PATH]");
```
panekj, line 41: oops, typo
```suggestion
                    println!("lapce [-h|--help] [-v|--version] [PATH]");
```

panekj:
> Why not use `clap`?

Why use `clap`?

TitanNano:
Because it's robust, very straightforward and comes with numerous features. For instance, with clap, there is no need to implement `--help` or `--version`. Especially, `--help` has to be kept in sync with the actual implementation, when doing hand rolled argument parsing...

panekj:
> Because it's robust, very straightforward and comes with numerous features. For instance, with clap, there is no need to implement `--help` or `--version`. Especially, `--help` has to be kept in sync with the actual implementation, when doing hand rolled argument parsing...

Did you read what I wrote in first post though?

TitanNano:
Yes, but it seems arbitrary. Laying groundwork by introducing an implementation that is hard to maintain, just to avoid a fundamental dependency, is counterproductive.

panekj:
I don't see a reason to shove a dependency for whatever basic functionality we introduce.
If we will have an advanced usecase that will benefit from argument parsing library, then
it will be appropriate time to introduce such library. Now is not the time, we have 2 flags
and one arg.

dzhou121:
@panekj Is this PR all ready now?

panekj:
> @panekj Is this PR all ready now?

Yes, unless you have any objections towards current implementation?

PR #763, opened 2022-07-12
feat: add terminal as a build-time feature

panekj:
sprinkled `#[cfg(feature...)]` across whole repo so lapce can be used on Windows versions that don't have pseudo console API (but the change is universal, so terminal can be removed on linux/macos as well)


panekj:
It's worth noting that last Windows version that was actively supported by Microsoft which did not support Pseudo Console API hasn't been supported for more than a year already.
I will be perfectly fine if the change won't be accepted as it introduces maintenance burden onwards.

dzhou121:
I'd prefer not to merge this.

PR #767, opened 2022-07-13
New Color Scheme

felipetesc:
Theme inspired on Sweet Theme for KDE Neon


panekj:
That should be a plugin, example: https://github.com/MinusGix/lapce-aleph

ghishadow:
You have to add custom theme repo link here https://github.com/lapce/lapce.github.io/blob/master/plugins.json, and Lapce will show in plugins tab.

felipetesc:
Thank You

Issue #121, opened 2022-03-17
release Linux packages in both RPM and DEB format as well as a tarball


SingingBush:
It would be great to get a repo setup so that Fedora users can install using `dnf install lapce`

Polpetta:
I suggest to have a look at https://github.com/PackageFoundation/yap, it allows to package software for different distributions directly from a single simil-PKGBUILD definition. It could be useful for targeting many distros with ease!

Disclamer: I'm a yap maintainer

Titaniumtown:
I have created a copr repository here with fedora 36 and rawhide builds for both x86_64 and aarch64: https://copr.fedorainfracloud.org/coprs/titaniumtown/lapce My spec file to build it can be found here: <https://github.com/Titaniumtown/lapce-copr/blob/main/lapce.spec>

All you need to do to install lapce on fedora is the following
```bash
sudo dnf copr enable titaniumtown/lapce # add copr repository hosting lapce builds
sudo dnf update
sudo dnf install lapce
```

I've also created #780 to address this as well in relation to Fedora.

panekj:
I'd be in favour of using [nFPM](https://nfpm.goreleaser.com/) for making deb/rpm.

blinxen:
I would like to package `lapce` for fedora and from my point of view the only thing that is keeping `lapce` from getting into the official fedora repos is that it uses [dependencies from git repositories](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories). Most of those dependencies are `tree-sitter` dependencies and they are located in `lapce-core`. Is there any plan / intent to change this in the near future?

panekj:
> Is there any plan / intent to change this in the near future?

Yes, #486

PR #780, opened 2022-07-15
Add fedora spec for packaging

Titaniumtown:
Addresses #121 (in relation to Fedora)

Adds a `.spec` file that can be used to compile the latest git version of Lapce for Fedora in copr. 

I need to do some testing and add possibly it to my [copr repository](https://copr.fedorainfracloud.org/coprs/titaniumtown/lapce) which I mentioned in #121

Corresponding package for master builds can be found here: <https://copr.fedorainfracloud.org/coprs/titaniumtown/lapce/package/lapce-git/>

### DIFF
--- a/lapce.spec
+++ b/lapce.spec
@@ -21,8 +21,8 @@ It is designed with Rope Science from the Xi-Editor which makes for lightning-fa
 cargo build --profile release-lto
 
 %install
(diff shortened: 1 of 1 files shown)


Titaniumtown:
Think this should've been squashed and merged, but that works too haha

PR #781, opened 2022-07-16
fix plugin panel scroll bar

### LINKED ISSUE #779: Plugin panel doesn't scroll
The plugin list is now longer than the plugin panel, but the panel doesn't have a scroll bar nor respond to the scroll wheel or scroll gesture on mac. I have tested on the latest master and scrolling doesn't seem to be supported on plugin panel. 

Test environment: Mac OS 12 on M1

bbrabbitt:
Try to fix #779 

Use `LapceScroll` to wrap the original `Plugin` component.

### DIFF
--- a/lapce-ui/src/plugin.rs
+++ b/lapce-ui/src/plugin.rs
@@ -1,3 +1,4 @@
+use crate::scroll::LapceScroll;
 use druid::{
     piet::{Text, TextAttribute, TextLayout as PietTextLayout, TextLayoutBuilder},
     BoxConstraints, Color, Cursor, Env, Event, EventCtx, FontWeight, LayoutCtx,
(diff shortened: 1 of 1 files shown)


bbrabbitt:
I've created a new commit to fix fmt and clippy warnings. Two commits are rebased for better git history.

dzhou121:
You can simply wrap LapceScroll on Plugin itself I think.

bbrabbitt:
> You can simply wrap LapceScroll on Plugin itself I think.

Theoretically yes but I don't know how to do it elegantly. So I imitate what file explorer did, using a wrapper to add LapceScroll to the plugin list.

dzhou121:
Something like this

```
impl Plugin {
    pub fn new() -> Self {
        Self { line_height: 25.0 }
    }

    pub fn new_panel(data: &LapceTabData) -> LapcePanel {
        let split_id = WidgetId::next();
        LapcePanel::new(
            PanelKind::Plugin,
            data.plugin.widget_id,
            split_id,
            vec![(
                split_id,
                PanelHeaderKind::None,
                LapceScroll::new(Self::new()).boxed(),
                None,
            )],
        )
    }
}

```

bbrabbitt:
Thank you for your advice, I'll update the PR accordingly.

dzhou121:
Thanks!

PR #788, opened 2022-07-17
Utilize the `fs` module

ImajinDevon:


lapce-proxy/src/buffer.rs
```diff
@@ -3,7 +3,6 @@ use lapce_rpc::buffer::BufferId;
 use std::ffi::OsString;
 use std::fs;
 use std::fs::File;
-use std::io::Read;
```
ImajinDevon, line 6: Remove unused import

lapce-proxy/src/buffer.rs
```diff
@@ -81,15 +80,13 @@ impl Buffer {
         self.rev += 1;
         let content_change = get_document_content_changes(delta, self);
         self.rope = delta.apply(&self.rope);
-        let content_change = match content_change {
```
ImajinDevon, line 84: Remove unnecessary code, produces the same result

dzhou121:
Can you run rustfmt on the code?

PR #821, opened 2022-07-19
Display documentation of completion items

### LINKED ISSUE #759: Show documentation for currently hovered code completion suggestion
During code prototyping, it's often useful to have documentation being shown for currently hovered/selected code completion suggestion. 

Example preview of the desired functionality present in VSCode:
image

MinusGix:
Closes #759   
image
  
This shows the documentation, if available, of the selected completion item. There is a setting to toggle this on/off.  
This also moves the markdown code to its own module, because it is now shared for usage by hover and this documentation (they are separate, so you can still hover while this documentation is being shown).  
There is also a few changes to how the markdown works, making so it no longer requires a syntax instance and supports multiple languages. (I'm not sure why I implemented the syntax the way it was before..)

### DIFF
--- a/defaults/settings.toml
+++ b/defaults/settings.toml
@@ -11,6 +11,7 @@ line-height = 25
 tab-width = 4
 show-tab = true
 scroll-beyond-last-line = true
+completion-show-documentation = true
 hover-delay = 300             # ms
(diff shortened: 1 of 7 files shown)


workingJ:
this looks great!

PR #822, opened 2022-07-19
Add keymap for going to previous tab

MinusGix:
This adds a keymap (Ctrl + Shift + Tab) for going to the previous tab. This should have been a part of #806, but I forgot it.


MinusGix:
(Okay this is actually broken, oops)

dzhou121:
It's actually ctrl+(shift)+tab for macOS as well.

MinusGix:
Fixed the bug, and did the suggested move of the keybindings to Ctrl+(shift)+Tab for both non-mac/mac.

PR #828, opened 2022-07-20
Change color of subtext on plugin menu

jath03:
This switches the subtext of the plugin menu to use the `editor.dim` color instead of `editor.foreground`.  It currently is works for the default themes, because the foreground colors aren't very bright, but for themes with brighter foreground colors, it's hard to differentiate between the plugin names, descriptions and authors visually.

### DIFF
--- a/lapce-ui/src/plugin.rs
+++ b/lapce-ui/src/plugin.rs
@@ -224,7 +224,7 @@ impl Widget<LapceTabData> for Plugin {
                     )
                     .text_color(
                         data.config
-                            .get_color_unchecked(LapceTheme::EDITOR_FOREGROUND)
+                            .get_color_unchecked(LapceTheme::EDITOR_DIM)
                             .clone(),
                     )
                     .build()
@@ -253,7 +253,7 @@ impl Widget<LapceTabData> for Plugin {
                         )
                         .text_color(
                             data.config
-                                .get_color_unchecked(LapceTheme::EDITOR_FOREGROUND)
+                                .get_color_unchecked(LapceTheme::EDITOR_DIM)
                                 .clone(),
                         )
                         .build()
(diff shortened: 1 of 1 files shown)


jath03:
image vs image

dzhou121:
Maybe we should have new color ```plugin.description``` for it. ```editor.dim``` looks a bit dark.

panekj:
Plugin author colour should also be slightly different from the description or maybe it could have bold font weight

jath03:
This example above is a bit dark, but it's from my monokai theme which is a bit of a wip, so it can be changed. What else is `editor.dim` used for?  Because this seems like the perfect situation for a slightly dimmer generic theme color.  Too much specialization for this stuff will just get annoying imo.

panekj:
> Too much specialization for this stuff will just get annoying imo.

As we add components and more features we will have to deal with it anyway so, the sooner the better.

bugadani:
It looks to me that this PR isn't needed any more. Can we close it?
image

jath03:
@bugadani That looks like the default theme, which has always looked fine.  The issue is that with other themes where the main foreground text color is brighter than the default theme there is a problem.

bugadani:
I see, thanks. I don't believe dim is the right choice here. In fact, I find the current state (bold for name, regular for others) good enough. Unfortunately I don't find the current color names (focused, foreground, dim) descriptive enough to know immediately what their intended use is, this looks like a use case for a "list primary" "list secondary" pair in some way. We already try to do something like that in the problems panel, which is currently completely inconsistent with the plugins panel (it uses foreground/dim). There is definitely room for streamlining in the UI and list consistency is as good a place to start as any.

Out of curiousity, how would the theme in the OP look now, that the plugin names are bold?

PR #833, opened 2022-07-20
Fix crashing on semantic tokens going past last line

MinusGix:
There is a crash where on a large project (like lapce, I typically did testing in the middle of lsp.rs) it would crash if you typed quickly (easy method: hold down enter). I believe this is a timing related crash.  
This sort of fixes the issue by just stopping the processing of the semantic tokens request if there is a newer revision of the buffer. This works because we just do a full semantic tokens request each time, and so if there's newer buffer (different rev) then we can just stop processing. This is probably a minor perf benefit in of itself.  
  
My theory as to what was happening:  
I noticed that it crashed when it processed (`request_semantic_tokens` finishing) a newer revision *before* an older revisions request. You'd be going through a bunch of revisions as you're typing quickly.  
I believe RA is getting told of the new file contents before the old request has finished, and so it provides offsets for the new file contents rather than the old ones. Then those new offsets cause a crash on the old text because they go out of bounds in it.

### DIFF
--- a/lapce-proxy/src/lsp.rs
+++ b/lapce-proxy/src/lsp.rs
@@ -262,7 +262,19 @@ impl LspCatalog {
             }
 
             let uri = client.get_uri(&buffer);
+            let old_buffer = buffer;
+
             client.request_semantic_tokens(uri, move |lsp_client, result| {
+                let buffers = lsp_client.dispatcher.buffers.lock();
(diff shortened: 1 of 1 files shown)


MinusGix:
Actually I think this is fine? We don't use the results of the semantic highlighting if the revision doesn't match so there isn't actually a change. I was just noticing an existing issue of semantic highlighting not being immediate.. I think.

PR #840, opened 2022-07-20
Save unsaved files on close, and restore unsaved files when opening lapce

nheuillet:
This PR Fixes the issue of losing all of the unsaved files when closing Lapce by doing the following:
- When a user closes Lapce, all the files that contain unsaved work are saved
- When a user opens the Workspace again, the files are fetched, and the buffer is applied again: this makes Undo work even after startup

Working example:


lapce-data/src/db.rs
```diff
@@ -428,6 +430,46 @@ impl LapceDb {
...
+        let sled_db = self.get_db()?;
+        let mut buffers = im::HashMap::new();
+        let mut unsaved_paths = Vec::new();
+
+        for val in sled_db.iter().keys() {
+            let val = val.unwrap();
+            let s = String::from_utf8((&val).to_vec())
+                .expect("invalid utf-8 sequence retrieving unsaved buffer");
+
+            if s.contains("unsaved_buffer") {
```
MinusGix, line 446: This could perhaps be `s.starts_with("unsaved_buffer:")` to avoid grabbing anything that just happens to be named `unsaved_buffer`.

lapce-data/src/db.rs
```diff
@@ -518,7 +560,30 @@ impl LapceDb {
...
+    }
+
+    fn insert_unsaved_buffer(&self, main_split: &LapceMainSplitData) -> Result<()> {
+        let sled_db = self.get_db()?;
+
+        for (path, doc) in &main_split.open_docs {
+            if !doc.buffer().is_pristine() && doc.content().is_file() {
+                let path_str = path.to_str().unwrap();
+                let buf_text = doc.buffer().text().to_string();
+                sled_db.insert(
```
dzhou121, line 579: I think we need to insert all unsaved buffers as one key, so that it can override it when the file is saved.

nheuillet: What would be the advantage compared to what is done right now?
it inserts a new key for each unsaved file, and on startup, the data is popped and then removed (see `lapce-data/src/db.rs:464`)

PR #843, opened 2022-07-21
wip: fix dirs and add multiarch/os

panekj:


dholth:
Is this running several commands on the remote host only if the proxy fails to start (and presumably doesn't exist)? It looks like it has a combined commands and a separate commands section.

PR #844, opened 2022-07-22
Load installed plugins first before making request

bbrabbitt:
Currently the plugin panel will wait for all requests to get completed before showing plugin information. With only 13 plugins this has already caused a ~5s delay on loading and listing plugins. This PR will load all local plugins before attempting loading everything from GitHub.


bbrabbitt:
Become part of #848

PR #863, opened 2022-07-24
Fix prev_boundary function

### LINKED ISSUE #778: Ctrl+Backspace skips whitespace
If you have code like:  
```rust
fn main() {}
|
```
Where `|` is the cursor location.
Ctrl+Backspace will transform it to:  
```rust
fn main() |
```
The correct method is:
```rust
fn main() {}|
```
  
As well,  
```rust
fn main() {}
 |
```
(where there's a space right before the cursor)
Ends up as
```rust
fn main() |
```
When it should be:
```rust
fn main() {}
|
```
(removing the whitespace, and if there was multiple spaces it would remove all of them)

Vovanoxin:
Resolve #778. It appeared that ``prev_boundary()`` was not working as expected. It makes sense to check also for the ``WordProperty::End``, which can be done using ``is_boundary()``.  I also had to change ``classify_boundary()`` since (Lf, Space) is more likely to be the Start.
This is the first OpenSource contribution in my life! Please review and let me know what you think, thank you!

### DIFF
--- a/lapce-core/src/word.rs
+++ b/lapce-core/src/word.rs
@@ -50,7 +50,7 @@ impl<'a> WordCursor<'a> {
             let mut candidate = self.inner.pos();
             while let Some(prev) = self.inner.prev_codepoint() {
                 let prop_prev = get_word_property(prev);
-                if classify_boundary(prop_prev, prop).is_start() {
+                if classify_boundary(prop_prev, prop).is_boundary() {
                     break;
                 }
                 prop = prop_prev;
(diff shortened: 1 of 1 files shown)


dzhou121:
Thanks for your contribution, especially your very first one!

The CI is complaining in tests. You might need to change the tests as well.

Vovanoxin:
Well, I dig a bit deeper and it does not look that simple. The tests look perfectly fine, though I think about adding some more. I would like to continue working on this issue

panekj:
This seems to be already fixed by #1294

PR #864, opened 2022-07-25
Fix: don't show the loading message when the plugin list is empty

bbrabbitt:
This is a temporary fix for loading message showing up when the plugin list is loaded but is empty. This patch adds a dummy plugin when the plugin list is loading. This is not a perfect fix but I think for now it's ok.

### DIFF
--- a/lapce-data/src/data.rs
+++ b/lapce-data/src/data.rs
@@ -193,6 +193,13 @@ impl LapceData {
                 )),
                 Target::Auto,
             );
+            let _ = event_sink.submit_command(
+                LAPCE_UI_COMMAND,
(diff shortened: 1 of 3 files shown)


MinusGix:
This could perhaps be done with something like  
```rust
enum PluginList {
    /// Still loading plugins
    Loading,
    /// Failed to load plugins
    Failure,
    Loaded(Vec<PluginDescription>),
}
```
Defaults to `PluginList::Loading`. Then if there's an error in the request it just switches to `Failure`. Then when getting a response (even an empty list) it just sets it to `PluginList::Loaded(plugins)`.

bbrabbitt:
I happen to have more free time to work on this so I'll close it for now.

PR #880, opened 2022-07-27
Add functionalities to enable, disable and remove plugins

bbrabbitt:
This PR adds a basic functionality to make wasm plugins able to be enabled, disabled or removed. When disabling a plugin, if the wasm plugin provides a `stop` function, it will be called. Otherwise, the added code will try to find the LSP started by the plugin and stop it. The PR adds a context menu to the plugin panel to provide enable/disable and remove options. It also fixes the loading problem mentioned in #848 by adding a enum to keep track of the plugin loading status.
Screen Shot 2022-07-27 at 2 53 59 AM
.

### DIFF
--- a/lapce-data/src/command.rs
+++ b/lapce-data/src/command.rs
@@ -398,6 +398,13 @@ pub enum LapceWorkbenchCommand {
     PreviousEditorTab,
 }
 
+#[derive(Debug, Clone)]
+pub enum PluginLoadingStatus {
+    Loading,
+    Failed,
+    Ok(Vec<PluginDescription>),
+}
+
 #[derive(Debug)]
 pub enum EnsureVisiblePosition {
     // Move the view so the cursor line will be at the center of the window.  If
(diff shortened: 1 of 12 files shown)


dzhou121:
A slight improvement is to change the Installed button to a cog icon or something to indicate there's some menu here now.

bbrabbitt:
> A slight improvement is to change the Installed button to a or something to indicate there's some menu here now. 

Sure I'll add it tomorrow

MinusGix:
I think it would be better to instead have a single `plugins.toml` file (in the config folder?) that has a list of what plugins are enabled (or it could be a setting in `settings.toml`?). This would also extend more naturally to workspace-specific configuration when we add that.

bbrabbitt:
> I think it would be better to instead have a single `plugins.toml` file (in the config folder?) that has a list of what plugins are enabled (or it could be a setting in `settings.toml`?), rather than a field on each plugins.toml file. This would also extend more naturally to workspace-specific configuration when we add that.

Sure, but I'll store disabled plugins since I just have a HashMap storing disabled plugins.

bbrabbitt:
I've made the changes accordingly, but there are conflicts to resolve.

PR #884, opened 2022-07-27
Use UTF16 encoding for LSPs

MinusGix:
This PR adds an `encoding.rs` file for switching between UTF16 and UTF8 offsets given just an iterator over the char indices (which lets us work with both `&str` and a `Rope` without allocating).   
- Query: I put the `encoding.rs` in `lapce-core` so that it could be accessed by `lapce-core/src/buffer.rs` and `lapce-proxy/src/buffer.rs`, but that required making `lapce-proxy` depend on `lapce-core`. I could add a separate crate `lapce-encoding` (or `lapce-util` if we want it to be more general), so that we don't make it as tightly interlinked?  
  
This removes the `offset_encoding` field on the lsp capabilities, because supporting the UTF8 extension for LSP clients would be challenging. It is easier and less complex for the code to just use UTF16 offsets.  
This PR became much easier after doing the split between UTF16 `lsp_types::Position` and UTF8 `usize`s.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2312,6 +2312,7 @@ dependencies = [
  "hotwatch",
  "ignore",
  "jsonrpc-lite",
+ "lapce-core",
  "lapce-rpc",
  "locale_config",
  "log",
--- a/lapce-core/src/buffer.rs
+++ b/lapce-core/src/buffer.rs
@@ -699,37 +699,36 @@ impl Buffer {
     }
 
     /// Converts a UTF8 offset to a UTF16 LSP position  
-    /// Panics if it is not a valid UTF16 offset
-    pub fn offset_to_position(&self, offset: usize) -> Position {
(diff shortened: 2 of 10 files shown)


dzhou121:
Can we not panic if the offset is in the middle of chat?

MinusGix:
Yeah, I can. Will require more changes, and I think I noticed some bugs in the semantic highlighting so I'll fix those up.

MinusGix:
Okay, I think this should be fixed. It is a second commit, just to make it somewhat separate (though not really that well).  
Places which translate between utf8 and utf16 will now fail in nicer ways, typically logging an error.
Though for some of them, like goto definition, it would be better to show something (notification? entry in the status bar?) about the failure to jump.  
I also fixed a bug in semantic lighting that I had accidentally introduced in the first commit which took most of the time. It needed a custom struct for iteration, because it needed the previous iterator's last offset.

PR #909, opened 2022-07-31
ci: build completely static lapce-proxy

panekj:


dzhou121:
would be musl build be slower?

panekj:
You mean in CI or like running?

dzhou121:
I meant running

panekj:
I don't see any reason why would it be slower

dzhou121:
something about the memory allocation?

panekj:
> [musl-1.2.1.tar.gz](https://musl.libc.org/releases/musl-1.2.1.tar.gz) ([sig](https://musl.libc.org/releases/musl-1.2.1.tar.gz.asc)) - August 4, 2020
This release features the new "mallocng" malloc implementation, replacing musl's original dlmalloc-like allocator that suffered from fundamental design problems. Its major user-facing new properties are the ability to return freed memory on a much finer granularity and avoidance of many catastrophic fragmentation patterns. In addition it provides strong hardening against memory usage errors by the caller, including detection of overflows, double-free, and use-after-free, and does not admit corruption of allocator state via these errors.

panekj:
There might be performance difference between jemalloc and mallocng if lapce-proxy heavily relies on allocating, and I could try to link to jemalloc statically but IMO, slightly degrading performance in comparison to having universally working proxy on all Linux x64 is worth it.

panekj:
https://github.com/actions/virtual-environments/issues/6002

PR #914, opened 2022-08-01
Fix hotkey for redo

rtsuk:
On Mac when a character is pressed along with shift and
command, the character returned is not the upper case
variant.  This code allows recognition of "Redo", command-
shift-z, to work.


rtsuk:
This works on Windows and Mac, but I'm open to discussing if some other implementation would be better.

rtsuk:
I cleaned up the implementation a bit, now it is really ready for testing/review.

PR #915, opened 2022-08-01
fix: replace toml-rs with toml_edit

panekj:
toml-rs is abandoned and not fixed anymore, it also has a bug
that in certain cases might prevent plugins getting installed
because of struct serialization, where `Value` is seralized
in alphabetical order and certain keys end after table which
is incorrect in Toml

this is also the reason why `cargo` switched to `toml_edit`
and I think it would be good to follow same dependencies as
cargo

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -423,6 +423,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3d7b894f5411737b7867f4827955924d7c254fc9f4d91a6aad6b097804b1018b"
 
 [[package]]
+name = "combine"
+version = "4.6.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


dzhou121:
Does toml_edit preserve the order it parses the file? It's useful when updating the config file.

panekj:
https://docs.rs/toml_edit/0.14.4/toml_edit/#limitations
image

panekj:
Mhm, `toml_edit::easy` doesn't preserve at all apparently

dzhou121:
It's actually fine 

```
let toml = r#"
[[bin]] # bin 1
[a.b.c.e]
[a]
[other.table]
[[bin]] # bin 2
[a.b.c.d]
[a.b.c]
[[bin]] # bin 3
"#;
    let expected = r#"
[[bin]] # bin 1
[[bin]] # bin 2
[[bin]] # bin 3
[a]
[a.b.c]
[a.b.c.e]
[a.b.c.d]
[other.table]
"#;
```

PR #916, opened 2022-08-01
Don't show autocomplete after whitespace keys

### LINKED ISSUE #742: Don't show autocomplete after whitespace keys
If you have text like:  
```rust
fn do_thing() {
|other_func();
}
```  
where `|` is the cursor  
or  
```rust
|fn do_thing() {
```
Then pressing Space/Tab/Enter should not open the completion menu.  
I think this can just be done as a check to ignore those specific characters, since I don't believe there's a normal place where you should be getting autocomplete after you press them?

zarathir:
Fixes #742
Check for whitespace command in `run_edit_command` and `receive_char` before triggering autocomplete.

### DIFF
--- a/lapce-data/src/editor.rs
+++ b/lapce-data/src/editor.rs
@@ -1269,7 +1269,13 @@ impl LapceEditorBufferData {
             }
         }
 
-        self.update_completion(ctx, false);
+        match cmd {
+            EditCommand::InsertNewLine
+            | EditCommand::InsertTab
+            | EditCommand::NewLineAbove
+            | EditCommand::NewLineBelow => {}
+            _ => self.update_completion(ctx, false),
+        }
         self.apply_deltas(&deltas);
 
         CommandExecuted::Yes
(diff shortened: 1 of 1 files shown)


lapce-data/src/editor.rs
```diff
@@ -1269,7 +1269,13 @@ impl LapceEditorBufferData {
             }
         }
 
-        self.update_completion(ctx, false);
+        match cmd {
+            EditCommand::InsertNewLine
```
Cody-Duncan, line 1273: Recommend: encapsulating expression of whether an EditCommand is one that inserts a whitespace character.
Something like
```
impl EditCommand {
    fn is_insert_whitespace(&self) _> bool {
        use EditCommand::*;
        match self {
            InsertNewLine
            | InsertTab
            | NewLineAbove
            | NewLineBelow => true,
            _ => false
        }
    }
}
```
Note: I did not check if this compiles.

Rationale: If anything else needs to similarly adapt to whitespace, it'll be handy to expect they have the same behavior. Makes it look more similar to its counterpart in expressing "if not whitespace, then update_completion", which seems like a requirement to check beforehand for 2/3 calls to update_completion.

Significance: Pedantic. I'd recommend accepting this Pull Request with or without this suggestion.

Cody-Duncan (approved):
Short and Straightforward. I see no reason to reject this change unless there's an argument that the user would want autocompletion on whitespace.

Checking behavior on vscode, it doesn't seem to trigger the autocomplete dropdown on whitespace. Seems like a good idea from the perspective of not pinging the Language Server needlessly and without context of something to complete.

Recommend: Approve from a feature parity and (minor) performance standpoint.

PR #927, opened 2022-08-04
Make completion and palette pay attention to page-up/down

### LINKED ISSUE #924: PageUp/PageDown for drop-down widgets
It would be useful to be able to navigate drop-down widgets (e.g. autocomplete widgets) using pageup/pagedown to scroll quicker through them.

MinusGix:
Fixes #924.  
This makes so using page-up/page-down work in completion and in the palette.


abmyii:
Thank you for the rapid fix!

PR #930, opened 2022-08-04
Prevent crash on save

rtsuk:
This prevents the crash described in #929, but I'm not certain it makes sense for the save command to be sent to anything but the front window. I'm also unsure how the path from the front window ends up getting sent to the empty, rear window.

I'm opening this PR for some guidance on what the right fix is in this case.

### DIFF
--- a/lapce-data/src/editor.rs
+++ b/lapce-data/src/editor.rs
@@ -1217,6 +1217,7 @@ impl LapceEditorBufferData {
             let rev = self.doc.rev();
             let event_sink = ctx.get_external_handle();
             let view_id = self.editor.view_id;
+            let tab_id = self.main_split.tab_id.clone();
             let (sender, receiver) = bounded(1);
             thread::spawn(move || {
                 proxy.get_document_formatting(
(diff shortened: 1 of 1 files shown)


dzhou121:
I think the issue is because the event was sent to ```Target::Auto```. They should be sent to ```Target::Widget(WindowTabWidgetId)```

https://github.com/lapce/lapce/blob/master/lapce-data/src/editor.rs#L1243
https://github.com/lapce/lapce/blob/master/lapce-data/src/editor.rs#L1957

rtsuk:
I'll give that a try.

rtsuk:
Although it's not clear to me how to chase down the window ID in that context.

dzhou121:
It would be ```self.main_split.tab_id```

rtsuk:
Cool I will fix it tonight.

rtsuk:
Unfortunately, [`document_save`](https://github.com/lapce/lapce/blob/master/lapce-data/src/data.rs#L1955) wants to submit at `BufferSave` command and uses auto, but since a widget ID isn't send I can't change it to target the widget. It then crashes for the same reason in the [`BufferSave` match arm in tab.rs](https://github.com/lapce/lapce/blob/master/lapce-ui/src/tab.rs#L1101).

PR #936, opened 2022-08-06
ci: add lapce-proxy for darwin (macos)

panekj:


dholth:
Does that include linux arm64?

panekj:
No

PR #937, opened 2022-08-06
Don't show autocomplete after backspacing over whitespace

### LINKED ISSUE #743: Don't show autocomplete after backspacing over whitespace
If backspacing deletes a whitespace character (such as deleting indentation) then it shouldn't show autocomplete.

zarathir:
Fixes #743 
Added `EditCommand::DeleteBackward` and `EditCommand::DeleteForward` to the match statement that checks for whitespace commands. On receiving one of these commands the deleted string gets checked if it's all whitespace.

Added `EditCommand::Undo` and `EditCommand::Redo` to the check, because i think it's not necessary to show autocompletion after an undo or redo.

Maybe it would make sense to extract this into a function?

### DIFF
--- a/lapce-data/src/editor.rs
+++ b/lapce-data/src/editor.rs
@@ -1327,6 +1327,7 @@ impl LapceEditorBufferData {
     ) -> CommandExecuted {
         let modal = self.config.lapce.modal && !self.editor.content.is_input();
         let doc = Arc::make_mut(&mut self.doc);
+        let doc_old = doc.buffer().text().to_string();
         let register = Arc::make_mut(&mut self.main_split.register);
         let cursor = &mut Arc::make_mut(&mut self.editor).cursor;
         let yank_data =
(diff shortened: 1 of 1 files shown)


lapce-data/src/editor.rs
```diff
@@ -2408,3 +2405,43 @@ fn apply_code_action(
...
+                _ => &0,
+            };
+
+            let end = match &deltas[0].0.els[1] {
+                xi_rope::DeltaElement::Copy(end, _) => end,
+                _ => &0,
+            };
+
+            if start > &0 && end > &0 && end > start {
+                let slice = &doc[*start..*end];
```
dzhou121, line 2429: You could use the ```slice_to_cow``` on the rope to make it more efficient without converting the rope to a string.

https://github.com/lapce/lapce/blob/master/lapce-core/src/buffer.rs#L931

zarathir: Would you pass the `Rope` to `show_completion(...)`? Because `slice_to_cow` takes a `Range` as argument it would end up something like this:
```rust
fn show_completion(cmd: &EditCommand, doc: &Rope, deltas: &[(RopeDelta, InvalLines)]) -> bool {
    [...]
    if start > &0 && end > start {
        !doc.slice_to_cow(*start..*end)
            .chars()
            .all(|c| c.is_whitespace() || c.is_ascii_whitespace())
    } else {
        true
}
```
I'll have to try if this works and update with results 😄

PR #944, opened 2022-08-08
fix: create unix-like paths when connected to wsl

hexnaught:
When running Lapce on a windows host, connected to and with a project
open via WSL (Windows Subsystem for Linux) - File creation and directory
creation were both failing due to `std::path` module constructing the new
folder and file paths in a Windows OS style rather than one appropriate
for Linux.

----

### Testing

I tested this change in the following two scenarios and both appeared to work fine:
- Windows Host -> Windows FileSystem
- Windows Host -> WSL FileSystem

I could not get Connect via SSH to work (connecting to a local network Pi) via the UI.
I can boot in to Arch to test Linux host in the morning (unless anybody beats me to testing Linux host while I sleep).

----

I am pretty new to rust (I had poked around in a spotify client written in Rust using Druid before) and it was fun to dive in and try to resolve the issue, after looking at various ways to handle the string/path construction and not being able to rely on OS detection (due to Windows being detected, as Host is running Lapse) I settled with the below.

More than happy to learn if there is a more appropriote way to do this!

### DIFF
--- a/lapce-data/src/explorer.rs
+++ b/lapce-data/src/explorer.rs
@@ -15,6 +15,7 @@ use xi_rope::Rope;
 
 use crate::data::LapceMainSplitData;
 use crate::data::LapceWorkspace;
+use crate::data::LapceWorkspaceType;
 use crate::document::LocalBufferKind;
 use crate::proxy::LapceProxy;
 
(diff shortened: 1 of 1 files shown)


dzhou121:
This is a tricky one. Maybe a better way is to do the join at the proxy side. @MinusGix Any thoughts?

MinusGix:
> This is a tricky one. Maybe a better way is to do the join at the proxy side. @MinusGix Any thoughts?

Doing as much path stuff specifically on the proxy would be good, yes. Our current method of using pathbuf is iffy for cross-platform proxies.  
We could deliberately use the LSP URI type as a way to represent paths, and just write some utility functions for common uses of it (like appending a path)? Since we'd be using URI for paths in plugins too.  
Or a custom path type that with some enum of the representation style (windows/linux/mac/?) on it, and the utility functions make it properly behave, and then we just use

Issue #815, opened 2022-07-19
Add File Explorer Refresh Button


MinusGix:
This would help soothe the issues where our watching code is breaking (as mentioned in a few other issues). It would probably require adding an extra beginning section to the file explorer panel to render the button.

jmj0502:
@MinusGix Hey! Do you think this issue will be too tough? I would like to give it a shot c:

MinusGix:
I think it shouldn't be too tough.  
- This could be done either as a button or a context menu option (latter is easier due to not having to do any mouse hit-testing on an icon; and not having to make a containing widget)
- Probably need a `LapceUICommand::FileExplorerRefresh` for the button / context menu option to send out when activated, which is then received by the file explorer  
- I think you just need to call `FileExplorerData::reload` when it receives the command, (which is what we used and fixed a good chunk of the reloading issues, I believe)

jmj0502:
Cool! I'll reach out (through discord or even here) in case I have questions c:

sanxiyn:
I think this issue should be closed.

PR #956, opened 2022-08-12
Adding A Refresh Button For The Context Menu Of The File Explorer

jmj0502:
# Solves #815 
* A new variant (`FileExplorerRefresh`) was added to the `LapceUICommand` enum.
* The `Refresh` button was added to the context menu and its respective event handler was added to `lapce-ui/src/explorer.rs`.


MinusGix:
Hm, it looks like clippy is failing on unrelated files. This is probably due to the recent 1.63 rust update, so that can be ignored.

jmj0502:
@MinusGix Ohh ok. So everything's alright feature? C:

PR #957, opened 2022-08-12
Replace tree-sitter highlighting code with Helix's impl

MinusGix:
This uses Helix's tree-sitter code, which is a modification of the official tree-sitter highlighter code to better support incremental highlighting. Rather than figuring out the details of what is needed for that and adapting our current method (which has issues because of it assuming you're only parsing one language at a time), we can use their code. (Which is under the MPL, but we can attribute the syntax.rs file to the helix project because it is a compatible license)  
It might be nice to turn it into a crate some time, since *for the most part* it is relatively independent.  
This also comes with a fix for markdown, since that's why I started on this whole change. The markdown crate is separated into `markdown` and `markdown.inline`. The `markdown` (block) does basic parsing, then uses injections to do the `markdown.inline` parsing, which provides most of the highlighting.  
  
I store the language `HighlightConfiguration` in a global lazy initialized instance. This was actually first change I made, because for injections we need to be able to get other language highlighters while we're highlighting some other language. The `RefCell<HashMap<LapceLanguage, HighlightConfiguration>>` method wouldn't work for that, since we might need to modify it (to initialize a language) while there was an existing reference.  

Bugs:
- There's imperfect injection highlighting for code blocks in markdown files. The pull requests to make the new markdown split work in helix mention this, but also mention that they fix it, and I use an adapted version of their code so I'm still not sure why it isn't working. Other injections in html seem to work fine at a glance, so presumably something to do with the special handling that markdown.inline needs.


archseer:
Quick reminder: helix is [MPLv2](https://github.com/helix-editor/helix/blob/master/LICENSE) licensed so `syntax.rs` will require proper attribution.

Issue #926, opened 2022-08-04
Shift + Split Editor Button should split down rather than to the right


MinusGix:
Currently the split editor button splits the editor to the right when clicked. However, when pressing shift and clicking it, it should open downwards instead. As well, it would be good to change the UI button to reflect that it would open downwards, like other editors do.  
![image](https://user-images.githubusercontent.com/13157904/182772652-881da9ae-3a62-49bd-a7f4-ec43ff7c55a6.png)

jmj0502:
@MinusGix Hey! I would like to give this one a shot too! (I'll need guidance, but that's probably the best way to keep on learning, so c:)

MinusGix:
You probably want to edit `lapce-ui/src/tab_header.rs` in the layout function.  
- You'd need to store whether shift is being currently held (in `event`, can probably just store it as part of the widget structure. Listen for KeyDown/KeyUp events then request paint on changes of shift key)  
- If shift is being held, use "split-vertical.svg" instead of "split-horizontal.svg". I just noticed that we call splitting left/right 'vertical' because it is on the vertical axis, but the file is named horizontal. Yay.  
- If shift is being held, use the alternate focus command  
- At a glance, I think you need to modify `tab.rs` to redispatch the horizontal focus command to make this work.  
For the icon, I think you can just use: https://github.com/lapce/lapce/blob/b05fd81b03bcdd0c326dce2062ae0c03755e86be/icons/ui/split-vertical.svg from (pj's icon pr)

jmj0502:
Cool! I'll reach out through discord in case I have questions c:

PR #963, opened 2022-08-14
Shift + Split Editor Button split down rather than to the right

jmj0502:
# Solves #926


jmj0502:
@MinusGix Hey! Hope you're doing great! 
I was able to implement part of the expected functionality. Currently the split icon changes when shift is pressed and the split command is handled appropriately. However, I haven't been able to effectively capture the updated value of `is_shift_down` from `EditorTabHeader`.

I added the `key_up` function to  `KeyPressData` and event handlers for `Event::KeyUp` (I call `key_up` in such handlers), but the value of `is_shift_down` never updates to `false`, despite the fact the `key_up` is called.

Something interesting tho', is the fact that `EditorTabHeader` effectively gets the value of `is_shift_down` when `key_down` is called. I think I'm missing something, but I don't really know what that may be. Any clue will help me a lot (I'm almost done implementing the feature :'v)

PR #970, opened 2022-08-17
fix: use -p to create full tree of directories

panekj:


panekj:
it also prevents failure if path exists already

dzhou121:
just realised it needs to check the remote platform for ```-p```

panekj:
oh yeah, I'm dumb

PR #978, opened 2022-08-21
Use `Segoe Fluent Icons` for window controls when available

pbnoyz:
Use `Segoe Fluent Icons` for window controls on **Windows 11** and fallback to ` Segoe MDL2 Assets` on the others. In addition, change code points to correct values.


panekj:
Hi, this change is not necessary as the font icons will be replaced with SVGs: https://github.com/lapce/lapce/pull/768

pbnoyz:
> Hi, this change is not necessary as the font icons will be replaced with SVGs: https://github.com/lapce/lapce/pull/768

That's pretty nice. Close this.

PR #981, opened 2022-08-22
fix: differentiate between stable/nightly/debug

panekj:


lapce-proxy/src/lib.rs
```diff
@@ -20,11 +20,17 @@ use lapce_rpc::{
...
-#[cfg(debug_assertions)]
-pub const APPLICATION_NAME: &str = "Lapce-debug";
+pub static APPLICATION_NAME: Lazy<&str> = Lazy::new(application_name);
 
-#[cfg(not(debug_assertions))]
-pub const APPLICATION_NAME: &str = "Lapce";
+fn application_name() -> &'static str {
+    if cfg!(debug_assertions) {
+        "Lapce-Debug"
+    } else if option_env!("RELEASE_TAG_NAME") == Some("nightly") {
```
dzhou121, line 28: They are nightly-shortsha

panekj: Right, I forgot to check updated code in `version()`
Would below be ok for both `application_name()` and `version()`? 
```suggestion
    } else if option_env!("RELEASE_TAG_NAME").unwrap_or("").starts_with("nightly") {
```

IMO it's simpler that current code:
```rs
    } else if option_env!("RELEASE_TAG_NAME").is_some()
        && option_env!("RELEASE_TAG_NAME")
            .unwrap()
            .starts_with("nightly")
    {
```

dzhou121: Yes it is better.

PR #982, opened 2022-08-22
feat: add remaining language IDs

panekj:
make most extensions matching case-insensitive and add case-sensitive
matching
sort extensions alphabetically (like in LSP spec page)

### DIFF
--- a/lapce-proxy/src/buffer.rs
+++ b/lapce-proxy/src/buffer.rs
@@ -173,21 +173,78 @@ pub fn read_path_to_string_lossy<P: AsRef<Path>>(
 pub fn language_id_from_path(path: &Path) -> Option<&'static str> {
     // recommended language_id values
     // https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentItem
-    Some(match path.extension()?.to_str()? {
-        "rs" => "rust",
-        "go" => "go",
-        "py" => "python",
-        "jl" => "julia",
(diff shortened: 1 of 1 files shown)


lapce-proxy/src/buffer.rs
```diff
@@ -173,21 +173,78 @@ pub fn read_path_to_string_lossy<P: AsRef<Path>>(
...
-        "c" | "h" => "c",
-        "js" => "javascript",
-        "jsx" => "javascriptreact",
-        "ts" => "typescript",
-        "tsx" => "typescriptreact",
-        "ex" | "exs" => "elixir",
-        _ => return None,
+        // Handle paths without extension
+        None => match path.file_name()?.to_str()? {
+            "dockerfile" => "dockerfile",
```
L1LLIAN, line 244: For Dockerfile's sometimes I like to have them with extensions for something like mutli-module jvm applications.
i.e: `Dockerfile.<module>` editor plugins like vscode's docker extensions support doing that so I think it would be nice if Lapce could as well. Though I feel like it should probably not clash with other existing language extensions. Probably a little niche but IMO a nice to have.

panekj: This will be handled by detecting file content which lapce currently can't.

L1LLIAN: > This will be handled by detecting file content which lapce currently can't.

Ah, alright :)

PR #983, opened 2022-08-22
feat: add LapceUICommand::OpenURI

panekj:
This is base for multiple other commands we can have in future like `Open Setting location`, `Reveal this file/folder in File Explorer` etc. and URLs in terminal/editor/UI

This was originally part of icon theme PR but I'm trying to decouple it into more smaller PRs so they can get merged already

### DIFF
--- a/lapce-ui/src/tab.rs
+++ b/lapce-ui/src/tab.rs
@@ -1223,6 +1223,20 @@ impl LapceTab {
                     LapceUICommand::UpdateKeymap(keymap, keys) => {
                         KeyPressData::update_file(keymap, keys);
                     }
+                    LapceUICommand::OpenURI(uri) => {
+                        ctx.set_handled();
+                        if !uri.is_empty() {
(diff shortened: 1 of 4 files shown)


lapce-ui/src/tab.rs
```diff
@@ -1223,6 +1223,20 @@ impl LapceTab {
                     LapceUICommand::UpdateKeymap(keymap, keys) => {
                         KeyPressData::update_file(keymap, keys);
                     }
+                    LapceUICommand::OpenURI(uri) => {
+                        ctx.set_handled();
+                        if !uri.is_empty() {
+                            log::debug!(target: "lapce_ui::tab::handle_event::open_uri", "uri: {uri}");
+                            match open::that(uri) {
```
dzhou121, line 1230: Maybe we want to run this in a different thread

panekj: `open::that` is non-blocking, what would be the benefit of separate thread?

dzhou121: Ah OK didn't know that.

Issue #134, opened 2022-03-17
Fails to run on Ubuntu 18.04 LTS


adamplumb:
I downloaded the binary from the homepage today and attempted to run.  I'm running Ubuntu 18.04.  Yep it's pretty old, though still under active support from Canonical.  Feel free to close this if Ubuntu 18 isn't a target for you but I figured I would file a ticket just in case.

```
$ ./lapce
./lapce: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by ./lapce)
./lapce: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.29' not found (required by ./lapce)
./lapce: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.28' not found (required by ./lapce)
```

Here's my `uname` output:
```
$ uname -a
Linux aplumb-Blade 5.4.0-104-generic #118~18.04.1-Ubuntu SMP Thu Mar 3 13:53:15 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

Looks like I'm running glibc 2.27

leonqli:
Same here.

dypa:
you can try build binary in docker container

create container `docker pull ubuntu:18.04 && docker run -it ubuntu:18.04 `

in docker container run commands

> apt update
> apt install -y curl git
> curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
> source $HOME/.cargo/env
> git clone https://github.com/lapce/lapce.git /opt
> apt install -y cmake pkg-config libfreetype6-dev libfontconfig1-dev libxcb-xfixes0-dev libxkbcommon-dev python3
> cd /opt/lapce
> cargo build --release

then copy binary to host system
`docker cp <containerId>:/opt/lapce/target/release/lapce .`

dypa:
fixed in 912768066e5bc942c7bc9b08146882b355a18923

panekj:
Lapce is built on ubuntu 18.04 for now so this can be closed

PR #984, opened 2022-08-22
ci: switch to ubuntu-latest

panekj:
ubuntu-18.04 is being deprecated and will soon be unavailable

https://github.com/actions/runner-images/issues/6002

Test run: https://github.com/panekj/lapce/runs/7954568797?check_suite_focus=true

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,19 +86,27 @@ jobs:
           retention-days: 1
 
   linux:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
+    container: ubuntu:18.04
     needs: tagname
     env:
       RELEASE_TAG_NAME: ${{ needs.tagname.outputs.tag_name }}
+      DEBIAN_FRONTEND: noninteractive
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -86,19 +86,27 @@ jobs:
           retention-days: 1
 
   linux:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
+    container: ubuntu:18.04
```
dzhou121, line 90: What is this for?

panekj: So it still builds against ubuntu 18.04 but doesn't depend on ubuntu-18.04 github runners

dzhou121: The reason we needed 18.04 was because of the GLIBC version. We can probably just use latest?

panekj: Sure, but what about https://github.com/lapce/lapce/issues/134 then?
Maybe I should add a build for each Ubuntu LTS from 18.04 to 22.04 and we can upload artefacts for all three (18.04/20.04/22.04)?

PR #985, opened 2022-08-22
fix: add *bsd targets

panekj:
`target_family = "unix"` includes all Unix-like systems, like *BSD, etc.

### DIFF
--- a/lapce-data/src/update.rs
+++ b/lapce-data/src/update.rs
@@ -100,7 +100,7 @@ pub fn extract(src: &Path, process_path: &Path) -> Result<PathBuf> {
     Ok(dest.join("Lapce.app"))
 }
 
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "linux", target_os = "freebsd", target_os = "openbsd"))]
 pub fn extract(src: &Path, process_path: &Path) -> Result<PathBuf> {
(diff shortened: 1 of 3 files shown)


dzhou121:
Can Lapce build on *BSD?

panekj:
> Can Lapce build on *BSD?

According to this https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=265892#c16, yes

dzhou121:
Maybe using #[cfg(any(target_os = "linux", target_os = "openbsd"))]

panekj:
That doesn't include `target_os = "freebsd"`

dzhou121:
That's how it is on Druid. So I'm really surprised.

I'm thinking the other way around. You add the thing that you know that's gonna work.

rtsuk:
Last time I checked, Rust considers MacOS as Unix-like, which might make that config less useful for something like Lapce.

panekj:
> Last time I checked, Rust considers MacOS as Unix-like, which might make that config less useful for something like Lapce.

That's why there was `not(target_os = "macos")` but you're commenting on already out of date information.

rtsuk:
Yep, just adding what I thought was a useful fact. Sorry for the noise.

PR #987, opened 2022-08-23
fix: replace File::open with directory opening

panekj:
Directories cannot be opened as files on Windows, that will be presented
with rather confusing error message: AccessDenied. Replacing it with dir
open function


panekj:
I tried to make it through `OpenOptions` but it just doesn't seem to work even though Win API states that using `GENERIC_READ` will open directory as file

PR #1001, opened 2022-08-26
docs: add pronunciation hint to README.md

### LINKED ISSUE #846: How to pronounce Labce?
How to pronounce Labce?

### LINKED ISSUE #973: Petition to change the project's name
Hi!

I'm glad to have discovered this project recently and I've started to take a keen interest in its development.

To me it looks highly promising considering – among other great things – it's written in Rust from scratch and it's using the tree-sitter language grammars.

However, I'm not particularly fond of the name. I hope you'd agree that the name of a software application – even if it's mainly for developers – does play a role in its potential and ability for gaining popularity and a role in its adoption.

I think with "lapce" the main issue is that it's hard to determine how to pronounce it. It just doesn't roll off the tongue. This will affect how presenters, reviewers etc. will talk about it and in turn it will affect the adoption rate.

I realize there's a website domain, screenshots and code and everything, but in my opinion the name is so important that it would be worth the trouble to at least rename the GUI labels and titles.

So here are two alternatives I would like to suggest for now:

1. **lapiz:** from the Spanish *lápiz* which means pencil. I think this one is very apt because the pencil is a writing tool and you're writing code with it so to speak. I've done some research and it seems it's not really taken by other projects. Best thing is that the pronunciation is unambiguous and it still fits the logo.
2. **lapice:** adding the vowel *i* makes it much easier to read / pronounce. But is it pronounced "lap-ice"? Or would people more likely say "lapiece"? It's ambiguous. You could come up with a meaning for the *i* so that the acronym still works.

Personally, I would definitely go with the 1st suggestion.

Hoping you'll consider my petition and also see what others have to say!

panekj:
fixes #846 
fixes #973 

Pronunciation poll: https://discord.com/channels/946858761413328946/1010249749565276160
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@
 <br/>
 
 
-Lapce ([pronounced `lapse`, IPA: /læps/](https://en.wiktionary.org/wiki/lapse)) is written in pure Rust with a UI in [Druid](https://github.com/linebender/druid) (which is also written in Rust). It is designed with [Rope Science](https://xi-editor.io/docs/rope_science_00.html) from the [Xi-Editor](https://github.com/xi-editor/xi-editor) which makes for lightning-fast computation, and leverages [OpenGL](https://www.opengl.org/) for rendering. More information about the features of Lapce can be found on the [main website](https://lapce.dev) and user documentation can be found on [GitBook](https://docs.lapce.dev/).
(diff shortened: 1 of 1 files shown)


dzhou121:
Can we just put the IPA in.

panekj:
Sure

PR #1005, opened 2022-08-27
CI: disable MacOS release in forked repos.

azizk:
I'd like to point out that everyone who forks the repo will get an email every night for a failing MacOS release build. That's because a certificate needs to be configured for it to work. Here is the failing step:

```
Import Certificate 0s
1 > Run apple-actions/import-codesign-certs@v1
11 Error: At least one of p12-filepath or p12-file-base64 must be provided
```

I'm posting the quick fix to disable it in forked repos. More elaborate solution is to add a job that checks whether a certificate is provided and to only run the `macos` job depending on that. Like here for example: https://github.com/vector-im/element-ios/pull/5496/files

Besides, perhaps it's worth reconsidering running the cron job every night? It seems a bit wasteful to have that run in every single forked repository as well, especially since it takes over 30 to 40 minutes to build one release for each system. Should I open a new issue?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -206,6 +206,7 @@ jobs:
           retention-days: 1
 
   macos:
+    if: github.repository == 'lapce/lapce'
     runs-on: macos-11
     needs: tagname
     env:


panekj:
You can disable workflow
image

panekj:
> I'd like to point out that everyone who forks the repo

Everyone who forks repo and **enables the workflows**
Workflows are not enabled by default in forked repos
image

azizk:
Oh, thanks! Didn't know that. I disabled it for now. But when I'm more experienced with the project it would be nice to run the release workflow. I guess I'd have to enable it again when that time arrives. But still, it would be annoying to get emails again. Not really satisfied with the solution yet...

panekj:
> But when I'm more experienced with the project it would be nice to run the release workflow.

Then it's up to you to change that in your fork.
The release workflow is specfically made to work best for lapce and macOS certificate is required to notarise the nightly/tag release builds since we want to provide those macOS builds.

azizk:
Of course, you have to provide the macOS builds. It would just be nice for devs, who fork the project, not to have to research what's happening when they receive an email about the failing release workflow...

bugadani:
In principle, I agree with the idea of this change - release workflows (not just mac) should not run by default on forks. I don't think it's often someone wants to fork Lapce, then require a daily build of their fork. If yes (because they want to create a derivative work, not (just) contribute back), they have the option to update the workflows to suit their needs.

So, IMO the release workflows should by default be disabled on forks.

> Everyone who forks repo and enables the workflows
> Workflows are not enabled by default in forked repos

Unfortunately, this isn't true for already existing forks. I was also quite surprised by daily failures when I rebased my fork a week or two ago. But indeed, new forks need to enable workflows manually:

image

> It would just be nice for devs, who fork the project, not to have to research what's happening

I don't think this change was made in the last 3 weeks, and I believe you have also rebranded your fork. This would suggest to me that you manually enabled workflows on your fork. This usually isn't something a contributor would want to do, as CI runs on contributions anyway.

As it looks like this PR isn't complete (only affects macos), but workflows are opt-in on forks anyway, I'm closing this PR. Thank you for opening it, I've learnt a few thing because of this! :)

Issue #681, opened 2022-06-25
Quit / Exit command


cart:
It looks like there isn't a way to quit the editor via a command. I'd like to be able to close the editor using something like `ctrl + q`.

MinusGix:
I'm mildly against a default keybinding of Ctrl+Q since it makes it easy to accidentally use when you're trying to use Ctrl+Tab (next tab) or Ctrl+W (close current tab), but it would be good to at least have a command that can be bound and a workbench command for it.

cart:
Makes sense! I agree with the "`Ctrl+Q` is a bad default" assessment. I'd probably roll with "no default binding, but providing a workbench command".

panekj:
Duplicate of #234

PR #1007, opened 2022-08-28
app: quit on custom key shortcut.

azizk:
My first pull request. :grin:

### DIFF
--- a/defaults/keymaps-common.toml
+++ b/defaults/keymaps-common.toml
@@ -4,6 +4,10 @@
 key = "F1"
 command = "palette.command"
 
+# [[keymaps]]
+# key = "ctrl+q"
+# command = "quit"
+
 # --------------------------------- Basic editing ---------------------------------------
 
 [[keymaps]]
--- a/defaults/keymaps-macos.toml
+++ b/defaults/keymaps-macos.toml
@@ -27,6 +27,10 @@ command = "open_settings"
 key = "meta+k meta+s"
 command = "open_keyboard_shortcuts"
 
+# [[keymaps]]
+# key = "meta+q"
+# command = "quit"
+
 # --------------------------------- Basic editing ---------------------------------------
 
 [[keymaps]]
--- a/lapce-data/src/command.rs
+++ b/lapce-data/src/command.rs
@@ -441,6 +441,10 @@ pub enum LapceWorkbenchCommand {
     #[strum(message = "Save All Files")]
     #[strum(serialize = "save_all")]
     SaveAll,
+
+    #[strum(serialize = "quit")]
+    #[strum(message = "Quit Editor")]
+    Quit,
 }
 
 #[derive(Debug)]
--- a/lapce-data/src/data.rs
+++ b/lapce-data/src/data.rs
@@ -1678,6 +1678,9 @@ impl LapceTabData {
                     }
                 }
             }
+            LapceWorkbenchCommand::Quit => {
+                ctx.submit_command(druid::commands::QUIT_APP);
+            }
         }
     }
 


MinusGix:
This fixes #681.  
Though, like mentioned there: I'm mildly against having the default keybinding for quitting be Ctrl+Q due to its location typically (on qwerty) being near commonly used keys like Ctrl+W (close tab) and Ctrl+Tab (change tab). It is good to have a command for that, though.

panekj:
> Though, like mentioned there: I'm mildly against having the default keybinding for quitting be Ctrl+Q due to its location typically (on qwerty) being near commonly used keys like Ctrl+W (close tab) and Ctrl+Tab (change tab). It is good to have a command for that, though.

This, but also I'm strongly against that keybind because it's specific to one platform only (macOS) while other platforms have different defaults for such action (e.g. Windows - `LAlt` + `F4`)

jath03:
> it's specific to one platform only (macOS

Personally I've tried to quit the editor with Ctrl+Q on linux and was disappointed that it didn't work, but I understand if you all don't want it enabled by default.

Do we want to come up with an alternative binding or just leave it unbound by default?

panekj:
> Do we want to come up with an alternative binding or just leave it unbound by default?

It should be unbound by default but available for user to bind. There should be provided another keybind for platform specific implementation (macOS injects CMD+Q so no binding(?), on Windows `LAlt` + `F4` needs to be manually binded by lapce, Linux (GNOME/KDE/etc.)???)

azizk:
Well, I see where you're coming from, but I don't think it's such an issue for several reasons:

1. I've probably never hit Ctrl+Q accidentally in all the years I've been programming with Sublime Text. And if I did it was no biggie. The editor restores unsaved changes/files.
2. Many other applications have the Ctrl+Q and Ctrl+W bindings, for example: Firefox (Linux). Notably Chromium doesn't have a default binding for quitting.
3. Binding it to something else would go against the grain. I think I remember from my Windows days that Ctrl+Q was quite common.

So yeah, I can see the usability aspect but on the other hand it's pretty much standard and it's never been a problem. It should especially not be a problem once Lapce saves and restores files that haven't been saved yet.

azizk:
> > Though, like mentioned there: I'm mildly against having the default keybinding for quitting be Ctrl+Q due to its location typically (on qwerty) being near commonly used keys like Ctrl+W (close tab) and Ctrl+Tab (change tab). It is good to have a command for that, though.
> 
> This, but also I'm strongly against that keybind because it's specific to one platform only (macOS, although CMD + Q is done by system already) while other platforms have different defaults for such action (e.g. Windows - `LAlt` + `F4`)

It's not just for macOS. I'm on Xfce Manjaro and there are a few apps that have that binding.

`LAlt+F4` is also a common shortcut on Linux. It closes or should close the current window, but not the whole application. I think it behaves the same on Windows.

That's what I'm missing. Closing the app with a shortcut and having all windows + unsaved files restored on restart.

panekj:
> So yeah, I can see the usability aspect but on the other hand it's pretty much standard and it's never been a problem. It should especially not be a problem once Lapce saves and restores files that haven't been saved yet.

Just because it's mitigated doesn't mean we should introduce a behaviour that would reduce user UX.

> 1. I've probably never hit Ctrl+Q accidentally in all the years I've been programming with Sublime Text. And if I did it was no biggie. The editor restores unsaved changes/files.

It's only your experience but we should always think about all users.

> 2. Many other applications have the Ctrl+Q and Ctrl+W bindings, for example: Firefox (Linux). Notably Chromium doesn't have a default binding for quitting.

Firefox (Windows) = `Ctrl` + `Shift` + `Q`
Chromium-based browsers = no binding
VSCode = No binding, `LeftAlt` + `F4` closes VSC window
Intellij = No binding

jath03:
All on Linux
Firefox = Ctrl+Q
Discord = Ctrl+Q
vscodium = Ctrl+Q
atom = Ctrl+Q

These are all browser/electron based though, so maybe Ctrl-Q just originated from there?  

I think we should do like a poll on discord or something to see what people want for the default, because that's ultimately what matters most: what the users want, right?

panekj:
> I think we should do like a poll on discord

I think not, ~25 users should not decide fate of rest of the userbase

> or something to see what people want for the default, because that's ultimately what matters most: what the users want, right?

No, because not every user uses all platforms and we should not introduce a new behaviour onto a platform.

> That's what I'm missing. Closing the app with a shortcut and having all windows + unsaved files restored on restart.

I'll give you an offer, remove the default keybind and the PR can get merged (almost) instantly.
We can discuss default keybind later under separate issue/discord.

azizk:
Ok! I just left it in commented out if that's alright. I also added `"Quit Editor"` as the command's message, renamed `"quit_app"` and `"QuitApp"` to `"quit"`/`"Quit"` and moved the `LapceWorkbenchCommand::Quit` case to the bottom of the match statement. :+1:

bugadani:
I think without the defaults this PR should not be controversial and should be good to go.

(Side note: Interesting that I can't link #681 to this PR in the "Development" section to the right.)

PR #1024, opened 2022-08-31
fix typescript highlighting

ConnorSMorrison:
Added strings, comments, many keywords, true, false, undefined, null, and function calls to highlighting


ghishadow:
Instead of forking `tree-sitter-typescript`, you might want to add queries directly to Lapce, like Julia 
https://github.com/lapce/lapce/tree/master/lapce-core/queries/julia 
https://github.com/lapce/lapce/blob/master/lapce-core/src/language.rs#L521 
that way, it will be easier to maintain and sync with `tree-sitter-typescript` upstream.

Issue #918, opened 2022-08-02
font/settings: Sluggish performance upon editing the font family in Editor Settings.


schwrzstrbn:
The previous problems were solved but it did not include the Editor Settings -> Font Family.
It produces a sluggish performance when editing the Font family in Editor settings, barely usable when the bug is reproduced.

![image](https://user-images.githubusercontent.com/45195351/182289972-ee71a5eb-b852-4a65-8cff-defae2a9cbac.png)

schwrzstrbn:
Sudden spikes in cpu usage when editing Editor Settings.

![image](https://user-images.githubusercontent.com/45195351/182290393-aeac5b40-b35b-44ed-a29e-21b813511f2b.png)

schwrzstrbn:
And after closing lapce, 

![image](https://user-images.githubusercontent.com/45195351/182290595-d8d91ed9-f043-4b86-a97d-3d2cc851c1bd.png)

StoyK:
I encountered the same problem when I tried to leave the font family field empty so that it uses the system default, as said in the description.

My guess on why this happens is that when you leave the font family field empty, it doesn't remove the field in the settings.toml file, it just sets it to an empty string. So it would be reasonable to assume that the editor keeps trying to open a font with an "" name which doesn't exist. 
![image](https://user-images.githubusercontent.com/67196527/188288783-51373cc5-8878-43e2-8b4f-696015a8b1fc.png)

The fix to the issue is manually removing the font-family field in the settings file, which makes the editor revert to system defaults

DuckDuckWhale:
Looks like this is fixed for me. Did #1030 forget to close the issue?

schwrzstrbn:
> Looks like this is fixed for me. Did #1030 forget to close the issue?

Well yeah, i'm closing this now.

filipencopav:
I tried to escape by pressing tab, it inserted a literal tab character, and it lagged just like the author of the issue describes it.

Isn't there a reason for this resource hogging and intensive CPU usage? It's trying to do something, for an indefinite amount of time. Maybe there should be a limit check, or maybe it should validate the input before doing more operations on it?

PR #1030, opened 2022-09-03
Add emptiness check for editor font family field

StoyK:
My first pull request!

Currently, there is an issue #918 where if you leave the font family field in the Editor Settings empty, Lapce keeps trying to open an empty font which doesn't exist, causing the editor to become unusable. The problem was that this font_family function didn't have a guard for the field being empty like the font_family function in the UI settings.


dzhou121:
Thank you!

Issue #662, opened 2022-06-21
Shift+Backspace should act like Backspace


imsuck:
Instead of deleting characters in front of the cursor like Backspace does, Shift+Backspace does nothing.

Platform: Windows
Lapce version: 0.1.2

lnicola:
Which Windows programs do this? In my experience, Shift-Delete cuts the selection or current line to clipboard.

PR #1040, opened 2022-09-04
Add keybindings for `Shift-Del`, `Shift-Ins`, `Ctrl-Ins`

lnicola:
These are probably less known, but AFAIK they work in Gtk apps, VS, VS Code and even older IDEs like Delphi and Turbo/Borland Pascal. Probably inherited from WordStar.

I _think_ I've seen `Shift-Del` working like Backspace once, but never as Del.

CC @MinusGix from https://github.com/lapce/lapce/issues/662

### DIFF
--- a/defaults/keymaps-common.toml
+++ b/defaults/keymaps-common.toml
@@ -22,11 +22,6 @@ command = "delete_forward"
 mode = "i"
 
 [[keymaps]]
-key = "shift+Delete"
-command = "delete_forward"
-mode = "i"
-
-[[keymaps]]
 key = "backspace"
 command = "delete_backward"
 mode = "i"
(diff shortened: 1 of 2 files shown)


lnicola:
I don't know, maybe someone with a MacBook can try it.

lnicola:
I think most Macs don't even have an Insert key, and Delete is Fn+Backspace. These shortcuts are probably not a priority for them.

bugadani (approved):
Got it, thanks.

lnicola:
Is there anything I can do to push this forward?

panekj:
It waits for @dzhou121

lnicola:
@dzhou121 I think this PR got reverted in the past couple of months, but the git history doesn't show anything. Do you have any idea about it?

panekj:
> I think this PR got reverted in the past couple of months, but the git history doesn't show anything.

It wasn't

lnicola:
Oops, you're right, I was looking at my own fork.

https://github.com/lapce/lapce/pull/2071 added a new command for `Shift-Delete`, which broke copying to clipboard.

PR #611, opened 2022-06-14
609: Collapse problem file

Eraden:
Resolves https://github.com/lapce/lapce/issues/609

obraz

PR #1050, opened 2022-09-05
Collapsible problem groups

### LINKED ISSUE #609: Request: Collapse file sections in problems view
It would be nice if I was able to collapse warnings / errors from specific files in the problems view.

bugadani:
Supersedes #611 
Resolves #609


bugadani:
So far I haven't done much. I've rebased and fixed a crash, but I have not really dug into the actual feature. What's obvious is that we should add some sort of indicator next to the collapsed file's name.

bugadani:
Sorry, there's still something I want to do

bugadani:
As I can't make it crash anymore, @dzhou121 this is ready for you.

panekj (approved):
Looks ok to me

PR #1052, opened 2022-09-05
Don't clone strings when looking up font family

bugadani:
The strings are immediately converted into an `Arc<str>` so this should save an allocation on each call.


bugadani:
Maybe a better approach would be to store the FontFamily object directly to avoid any allocations, but this is an improvement as is.

PR #1053, opened 2022-09-05
Clean up settings update

bugadani:
There's no reason to copy a bunch of static data into vectors. This PR cleans up these clones and deduplicates converting structured data into hashmap form.

### DIFF
--- a/lapce-ui/src/settings.rs
+++ b/lapce-ui/src/settings.rs
@@ -27,6 +27,7 @@ use lapce_data::{
     keypress::KeyPressFocus,
     settings::{LapceSettingsFocusData, SettingsValueKind},
 };
+use serde::Serialize;
 use xi_rope::Rope;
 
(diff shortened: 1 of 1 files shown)


lapce-ui/src/settings.rs
```diff
@@ -375,74 +376,56 @@ impl LapceSettings {
...
             self.children.push(WidgetPod::new(
                 LapcePadding::new(
                     (10.0, 10.0),
                     LapceSettingsItem::new(
                         data,
-                        kind.clone(),
-                        field.clone(),
-                        descs[i].to_string(),
-                        settings.get(&field).unwrap().clone(),
+                        kind.to_string(),
```
bugadani, line 425: Ideally, all of these strings should be turned into `&'static str` but I did not want to limit flexibility. They might turn into `ReadOnlyString`s, though.

PR #1084, opened 2022-09-07
Use host shell in terminal when running inside Flatpak

Newbytee:
This improves the usability of Lapce significantly for people unaware of the `$ flatpak-spawn --host bash` trick. With this, the Lapce Flatpak's default terminal runs on the host outside of the container and has access to all your regular programs. Without this patch, what you get is a sandboxed shell that only has access to whatever is available inside of the Flatpak container, which can be unexpected and confusing.

### DIFF
--- a/lapce-proxy/src/terminal.rs
+++ b/lapce-proxy/src/terminal.rs
@@ -55,14 +55,38 @@ impl Terminal {
                 BaseDirs::new().map(|d| PathBuf::from(d.home_dir()))
             };
         let shell = shell.trim();
-        if !shell.is_empty() {
+        let inside_flatpak = is_inside_flatpak();
+
+        if !shell.is_empty() || inside_flatpak {
             let mut parts = shell.split(' ');
-            let program = parts.next().unwrap();
-            if let Ok(p) = which::which(program) {
+
+            if inside_flatpak {
+                let flatpak_spawn_path = "/usr/bin/flatpak-spawn".to_string();
+                let host_shell = flatpak_get_default_host_shell();
+
+                let args = if shell.is_empty() {
+                    vec!["--host".to_string(), host_shell]
+                } else {
+                    vec![
+                        "--host".to_string(),
+                        host_shell,
+                        "-c".to_string(),
+                        shell.to_string(),
+                    ]
+                };
+
                 config.pty_config.shell = Some(Program::WithArgs {
-                    program: p.to_str().unwrap().to_string(),
-                    args: parts.map(|p| p.to_string()).collect::<Vec<String>>(),
+                    program: flatpak_spawn_path,
+                    args,
                 })
+            } else {
+                let program = parts.next().unwrap();
+                if let Ok(p) = which::which(program) {
+                    config.pty_config.shell = Some(Program::WithArgs {
+                        program: p.to_str().unwrap().to_string(),
+                        args: parts.map(|p| p.to_string()).collect::<Vec<String>>(),
+                    })
+                }
             }
         }
         setup_env(&config);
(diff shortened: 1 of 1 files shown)


lapce-proxy/src/terminal.rs
```diff
@@ -307,3 +327,40 @@ fn set_locale_environment() {
...
     std::env::set_var("LC_ALL", locale + ".UTF-8");
 }
+
+#[cfg(target_os = "linux")]
+fn flatpak_get_default_host_shell() -> String {
+    let env_string = Command::new("flatpak-spawn")
+        .arg("--host")
+        .arg("printenv")
+        .output()
+        .unwrap()
```
dzhou121, line 337: would this panic on non-flatpak?

panekj: only if `/.flatpak-info` exists, which is unlikely

panekj: it's only executed in 
```rs
if inside_flatpak {
```

Newbytee: As explained in a comment, checking for `/.flatpak-info` is the de-facto way of checking whether you are inside of a Flatpak container. I don't think there's any reasonable scenario where that file would exist outside of a Flatpak container. Furthermore, if you did create that file on your host for whatever reason, you would break more apps than Lapce.

dzhou121: That's fine then.

If you could fix the fmt and clippy, it's good to merge.

Newbytee:
Can I somehow re-run the CI pipeline? I'm not familiar with GitHub actions.

panekj:
Push new commits to branch or force-push branch

lapce-proxy/src/terminal.rs
```diff
@@ -307,3 +331,52 @@ fn set_locale_environment() {
...
+}
+
+#[cfg(not(target_os = "linux"))]
+fn is_inside_flatpak() -> bool {
+    false // Flatpak is only available on Linux
+}
+
+#[cfg(target_os = "linux")]
+fn is_inside_flatpak() -> bool {
+    use std::path::Path;
```
panekj, line 375: You can use fully qualified path instead of importing it if you use it just once

```rs
std::path::Path::new
```

Newbytee: Personally, I think it's easier to read with the import done separately. But, if this is preferred I can change it.

lapce-proxy/src/terminal.rs
```diff
@@ -307,3 +331,52 @@ fn set_locale_environment() {
...
+
+    const FLATPAK_INFO_PATH: &str = "/.flatpak-info";
```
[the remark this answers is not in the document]
Newbytee: No, it's just personal preference.

sebastian-de:
Tested both as flatpak and native - seems to work as intended.

sebastian-de:
@Newbytee Since users can deny access to org.freedesktop.Flatpak (via Flatseal for example) can you extend the flatpak detection to handle the following case?
- Lapce runs in flatpak container
- `flatpak-spawn --host` is not available

You can test this by running:
```
flatpak install --user https://dl.flathub.org/build-repo/107147/dev.lapce.lapce.flatpakref
flatpak override --user --no-talk-name=org.freedesktop.Flatpak dev.lapce.lapce
flatpak run --user dev.lapce.lapce
```
In this case the terminal doesn't start at all. There should be a fallback to the default (in-container) terminal.

After testing, run `flatpak override --user --reset dev.lapce.lapce`

Newbytee:
> @Newbytee Since users can deny access to org.freedesktop.Flatpak (via Flatseal for example) can you extend the flatpak detection to handle the following case?
> 
>     * Lapce runs in flatpak container
> 
>     * `flatpak-spawn --host` is not available
> 
> 
> You can test this by running:
> 
> ```
> flatpak install --user https://dl.flathub.org/build-repo/107147/dev.lapce.lapce.flatpakref
> flatpak override --user --no-talk-name=org.freedesktop.Flatpak dev.lapce.lapce
> flatpak run --user dev.lapce.lapce
> ```
> 
> In this case the terminal doesn't start at all. There should be a fallback to the default (in-container) terminal.
> 
> After testing, run `flatpak override --user --reset dev.lapce.lapce`

Thanks, that's a good idea. I didn't think about this possibility. I'll fix it as soon as possible. Maybe you could make an issue about this?

PR #1099, opened 2022-09-08
use official tree-sitter-ruby instead of fork

Stanislav-Lapata:


bugadani:
I think I would prefer not pinning a particular revision. `version = ...` might be ineffective (docs aren't clear on it), but it gives us an indication of what lib version we want to use. If a build fails, we can pin the last working revision then. Otherwise, I support the switch to upstream.

bugadani (approved):
Well, apart from my merge commit, which I hate but that's what github gave me, I think this is good to go.

Issue #1119, opened 2022-09-10
Support deleting words in the in-built Terminal


VarLad:
## Is your feature request related to a problem? Please describe.

The editor has the option to delete entire words using Ctrl + Backspace. 
Something similar for the terminal would be awesome too. Most shells support this via Alt + Backspace. 

## Describe the solution you'd like
Allow the in-built terminal to remove entire words with Alt + Backspace

PR #1120, opened 2022-09-10
Consider Alt + Backspace for deletion too

VarLad:
Fixes https://github.com/lapce/lapce/issues/1119


VarLad:
Thank you @bugadani for pointing to the Alacritty code ^^

PR #1122, opened 2022-09-10
Add support for adding support for DLang LSPs

LunaTheFoxgirl:
Adds entries for DLang as needed by https://github.com/LunaTheFoxgirl/lapce-dlang


dzhou121:
Thanks!

Issue #1133, opened 2022-09-11
Terminal/Flatpak: Fallback to default when flatpak-spawn --host is not available


sebastian-de:
brought up here: https://github.com/lapce/lapce/pull/1084#issuecomment-1242808179

Since users can deny access to org.freedesktop.Flatpak (via Flatseal for example) can you extend the flatpak detection to handle the following case?

- Lapce runs in flatpak container
- `flatpak-spawn --host` is not available

You can test this by running:

```
flatpak install --user https://dl.flathub.org/build-repo/107147/dev.lapce.lapce.flatpakref
flatpak override --user --no-talk-name=org.freedesktop.Flatpak dev.lapce.lapce
flatpak run --user dev.lapce.lapce
```

In this case the terminal doesn't start at all. There should be a fallback to the default (in-container) terminal.

After testing, run `flatpak override --user --reset dev.lapce.lapce`

PR #1135, opened 2022-09-11
Check if flaptak-spawn --host works before attempting to use it

Newbytee:
Users may have manually removed the org.freedestkop.Flatpak session
bus talk name from permissions, in which case flatpak-spawn --host
won't work.

Closes https://github.com/lapce/lapce/issues/1133

---

This is not the cleanest solution to this problem, but without pulling in something like zbus I think this is the best we can do.


sebastian-de:
Thanks, works fine!
I now get a default shell with `--no-talk-name=org.freedesktop.Flatpak`.

PR #1149, opened 2022-09-11
A bunch of tiny cleanups

bugadani:
Nothing really important. Moved some rope related code into RopeText so they can be used ourside of a Buffer


lapce-core/src/buffer/mod.rs
```diff
@@ -29,6 +29,11 @@ use crate::{
     word::WordCursor,
 };
 
+mod rope_text;
+
+// TODO(dbuga): don't be lazy here
```
dzhou121, line 34: Handle this todo?

bugadani: Definitely :)

PR #1151, opened 2022-09-11
Make sure flatpak-spawn is found in $PATH on linux

Nykseli:
This pull request fixes a bug where on linux, lapce would not work if `flatpak-spawn` is not found in `$PATH`.


panekj:
Closing in favour of https://github.com/lapce/lapce/pull/1159

PR #1152, opened 2022-09-11
fix: create directory tree before plugin download

panekj:
Fixes plugin installation when `wasm` specifies any kind of directory tree

### DIFF
--- a/lapce-proxy/src/plugin/mod.rs
+++ b/lapce-proxy/src/plugin/mod.rs
@@ -806,6 +806,11 @@ pub fn download_volt(volt: VoltInfo, wasm: bool) -> Result<VoltMetadata> {
         let url = url.join(wasm)?;
         {
             let mut resp = reqwest::blocking::get(url)?;
+            if let Some(path) = path.join(&wasm).parent() {
+                if !path.exists() {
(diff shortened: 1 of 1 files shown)


lapce-proxy/src/plugin/mod.rs
```diff
@@ -806,6 +806,9 @@ pub fn download_volt(volt: VoltInfo, wasm: bool) -> Result<VoltMetadata> {
         let url = url.join(wasm)?;
         {
             let mut resp = reqwest::blocking::get(url)?;
+            if let Some(path) = path.join(&wasm).parent() {
+                fs::DirBuilder::new().recursive(true).create(path)?;
```
dzhou121, line 810: Maybe check if the path exits first?

panekj: > It is considered an error if the directory already exists unless recursive mode is enabled.

panekj: https://doc.rust-lang.org/std/fs/struct.DirBuilder.html#method.create

dzhou121: It feels unnecessary if it exists even if it doesn’t throw an error.

Issue #1125, opened 2022-09-10
Adding a Flutter Batterie


LimaneGaya:
Adding a flutter batterie allowing syntax highlighting ...

Conrad306:
I'll try and work on a plugin for it, im doing so right now. I'll see how it goes and start testing it.

LimaneGaya:
That would be really great if it can happen, it will make me switch from vs code

franklinblanco:
Let's close this Issue

PR #1153, opened 2022-09-11
Adding support for the dart language (help needed)

franklinblanco:
Hello, great codebase you got, it's become my favorite editor in about 2 days of use. Even though there are some bugs it's FOSS & speed > anything else for me. 

I'm here today because I do flutter + rust development and I found a tree_sitter_dart crate. So I followed the EXTREMELY WELL Documented instructions on how to add a language in lapce-core/Language.rs

I did everything. But I'm getting this one problem I can't solve: 
Screen Shot 2022-09-11 at 11 46 56 AM

Followed the web on this one and found this: 
[Answer to the error](https://stackoverflow.com/questions/27895946/expected-fn-item-found-a-different-fn-item-when-working-with-function-pointer)

But couldn't seem to work it out as the answer had primitive function return types and in this case it's a struct return type. tree_sitter::Language.

So, I've rattled a bit and tried to find what I can do, but I'm a bit lost. I see you guys are very active and would appreciate at least some guidance here! Thanks!

Mentioned in #1125

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4931,12 +4931,11 @@ dependencies = [
 
 [[package]]
 name = "tree-sitter"
-version = "0.17.1"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d18dcb776d3affaba6db04d11d645946d34a69b3172e588af96ce9fecd20faac"
+checksum = "dce934b47829db628c3b87c735bb2dcec95c936f1da74acf0baf856ae69347e8"
 dependencies = [
  "cc",
- "regex",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


panekj:
Error means that the library is using incompatible version of `tree-sitter` dependency

franklinblanco:
> Error means that the library is using incompatible version of `tree-sitter` dependency

Holy! Thank you that was so quick, I'm going to fork it and attempt to update it!!

franklinblanco:
Now it's this, guess I'm gonna do some more research before moving on with this
Screen Shot 2022-09-11 at 11 58 22 AM

ghishadow:
try using https://github.com/syntacti/tree-sitter-dart if it works

franklinblanco:
@ghishadow That seemed to compile correctly. I'm going to build it locally and test a dart project. Thank you very much for your help!!!

franklinblanco:
Done! @ghishadow Tested, works just fine, thank you millions for the help, thinking of supporting the project with a small donation, is there a preferred route?

panekj:
> thinking of supporting the project with a small donation, is there a preferred route?

nope

franklinblanco:
@panekj Sorry about that! Forgot to rename a test

PR #1154, opened 2022-09-11
dep: replace tree-sitter-json crate with git repo

panekj:
Replacing crate with git repo because maintainers have no permission to publish new crate versions


bugadani:
Please put a comment in the Cargo.toml to explain this. Someone might want to revert this change otherwise :)

bugadani (approved):

PR #1157, opened 2022-09-11
lapce-proxy: fix flatpak host detection

VolatileMark:
Fix an issue with @Newbytee's flatpak host detection, where if the editor was launched outside a flatpak container, the terminal would fail to open with the following error:

```
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', lapce-proxy/src/terminal.rs:388:10
```

### DIFF
--- a/lapce-proxy/src/terminal.rs
+++ b/lapce-proxy/src/terminal.rs
@@ -381,13 +381,15 @@ fn flatpak_should_use_host_terminal() -> bool {
     const FLATPAK_INFO_PATH: &str = "/.flatpak-info";
 
     // Ensure flatpak-spawn --host can execute a basic command
-    let host_available = Command::new("flatpak-spawn")
+    let host_available = match Command::new("flatpak-spawn")
         .arg("--host")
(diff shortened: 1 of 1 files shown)


Newbytee:
Sorry about this. I have `flatpak-spawn` installed on my host, so I didn't catch it. This looks reasonable.

Newbytee (approved):

panekj:
There is already a PR opened to fix this: https://github.com/lapce/lapce/pull/1151

Newbytee:
I think this approach is cleaner than #1151 personally.

VolatileMark:
> There is already a PR opened to fix this: #1151  

whoops, sorry I missed that. Should I close this one?

bugadani:
I'm not saying this PR is bad, so thanks for posting it, but I think #1159 is a bit nicer. Granted, I might be *slightly* biased with this opinion.

PR #1158, opened 2022-09-11
fix: remove non-existent argument

panekj:


ghishadow (approved):

bugadani:
I'm merging this because we have confirmation it is correct.

PR #1159, opened 2022-09-11
lapce-proxy: fix flatpak host detection

bugadani:
Supersedes #1157 and #1151

This PR
 - cleans up a block comment (a bit of style consistency)
 - only executes the flatpak command if the info file exists
 - does not crash on error

### DIFF
--- a/lapce-proxy/src/terminal.rs
+++ b/lapce-proxy/src/terminal.rs
@@ -380,6 +380,12 @@ fn flatpak_should_use_host_terminal() -> bool {
 
     const FLATPAK_INFO_PATH: &str = "/.flatpak-info";
 
+    // The de-facto way of checking whether one is inside of a Flatpak container is by checking for
+    // the presence of /.flatpak-info in the filesystem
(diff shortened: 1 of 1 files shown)


Newbytee (approved):
Looks great! I like that it doesn't needlessly shell out when `/.flatpak-info` doesn't exist.

panekj (approved):

bugadani:
Let's hope this actually works, as I use Windows and this code is grayed out for me :P

Newbytee:
I can test it, one moment.

panekj:
Logic is sound and clippy didn't complain on Linux so I think it's ok

VolatileMark:
Works on my system (Fedora 36)

Newbytee:
> Works on my system (Fedora 36)

Same here when running on the host, but I'm currently waiting for the Flatpak build to finish so I can test that too.

Newbytee:
Tested both Flatpak-contained with and without the Flatpak talk name, and also host both with and without `flatpak-spawn` installed. All four cases worked as expected! :tada:

PR #1161, opened 2022-09-11
feat: add Svelte & LaTeX

panekj:
Signed-off-by: Jakub Panek <me@panekj.dev>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2559,6 +2559,7 @@ dependencies = [
  "tree-sitter-ruby",
  "tree-sitter-rust",
  "tree-sitter-scss",
+ "tree-sitter-svelte",
  "tree-sitter-swift",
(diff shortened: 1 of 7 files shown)


franklinblanco:
Dang, was just done with Adding svelte, about to create a PR and saw this, good catch. Nice to see such a good OS community

panekj:
> Dang, was just done with Adding svelte, about to create a PR and saw this, good catch. Nice to see such a good OS community

It's better to just mention them in #272 so we can add bunch of them at the same time as each PR causes conflict

zero-edge:
Please forgive me if this is obvious, but how do I use the newly merged Svelte language support?

panekj:
> Please forgive me if this is obvious, but how do I use the newly merged Svelte language support?

You can't because it's not included in 0.2.0

Issue #871, opened 2022-07-25
In the terminal, text is not selectable and copy-paste isn't working


liyun-miao:
In the terminal, I was able to execute commands; however, I wasn't able to select, copy, or paste any content. I would like to be able to select the content in the terminal so that I could copy it to the text editor.
I am using Lapce for Windows (version v0.1.3).

BracketMaster:
Same for MacOS.

vejnar:
Same for Linux.

franklinblanco:
Duplicate of #726

panekj:
Implemented in https://github.com/lapce/lapce/pull/1163

Siddhartha351:
cant even change terminal from windows. its completely blank. help

PR #1163, opened 2022-09-11
terminal: Implement text selection and copy/paste functionality in terminal mode

VolatileMark:
See issue  #871

**// TODO**
- [x] Text selection
- [x] Copy (ctrl+shift+c for Windows/Linux / cmd+shift+c for MacOS)
- [x] Paste (ctrl+shift+c for Windows/Linux / cmd+shift+c for MacOS)
- [x] Full word select (with double left-click)
- [x] Entire line select (with triple left-click)
- [x] Use caret cursor when hovering over the terminal
- [x] Right-click copy/paste functionality

### DIFF
--- a/lapce-core/src/command.rs
+++ b/lapce-core/src/command.rs
@@ -298,6 +298,11 @@ pub enum FocusCommand {
     Rename,
     #[strum(serialize = "confirm_rename")]
     ConfirmRename,
+    /// Terminal selection commands
+    #[strum(serialize = "terminal_copy_selection")]
+    TerminalCopySelection,
+    #[strum(serialize = "terminal_paste_selection")]
+    TerminalPasteSelection,
 }
 
 #[derive(
--- a/lapce-core/src/command.rs
+++ b/lapce-core/src/command.rs
@@ -298,11 +298,6 @@ pub enum FocusCommand {
     Rename,
     #[strum(serialize = "confirm_rename")]
     ConfirmRename,
-    /// Terminal selection commands
-    #[strum(serialize = "terminal_copy_selection")]
-    TerminalCopySelection,
-    #[strum(serialize = "terminal_paste_selection")]
-    TerminalPasteSelection,
 }
 
 #[derive(
--- a/lapce-data/src/terminal.rs
+++ b/lapce-data/src/terminal.rs
@@ -447,6 +447,22 @@ impl KeyPressFocus for LapceTerminalViewData {
                         );
                     }
                 }
+                FocusCommand::TerminalCopySelection => {
+                    if self.terminal.mode == Mode::Terminal {
+                        let mut raw = self.terminal.raw.lock();
+                        let term = &mut raw.term;
+                        if let Some(content) = term.selection_to_string() {
+                            Application::global().clipboard().put_string(content);
+                        }
+                    }
+                }
+                FocusCommand::TerminalPasteSelection => {
+                    if self.terminal.mode == Mode::Terminal {
+                        if let Some(s) = Application::global().clipboard().get_string() {
+                            self.receive_char(ctx, &s);
+                        }
+                    }
+                }
                 _ => return CommandExecuted::No,
             },
             _ => return CommandExecuted::No,
--- a/lapce-data/src/terminal.rs
+++ b/lapce-data/src/terminal.rs
@@ -458,7 +458,8 @@ impl KeyPressFocus for LapceTerminalViewData {
                 }
                 FocusCommand::TerminalPasteSelection => {
                     if self.terminal.mode == Mode::Terminal {
-                        if let Some(s) = Application::global().clipboard().get_string() {
+                        if let Some(s) = 
+                            Application::global().clipboard().get_string() {
                             self.receive_char(ctx, &s);
                         }
                     }
--- a/lapce-data/src/terminal.rs
+++ b/lapce-data/src/terminal.rs
@@ -11,14 +11,15 @@ use alacritty_terminal::{
     Term,
 };
 use druid::{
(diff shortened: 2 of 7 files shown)


lapce-data/src/terminal.rs
```diff
@@ -447,6 +447,23 @@ impl KeyPressFocus for LapceTerminalViewData {
                         );
                     }
                 }
+                FocusCommand::TerminalCopySelection => {
+                    if self.terminal.mode == Mode::Terminal {
+                        let mut raw = self.terminal.raw.lock();
+                        let term = &mut raw.term;
+                        if let Some(content) = term.selection_to_string() {
+                            Application::global().clipboard().put_string(content);
```
bugadani, line 455: There should be a `SystemClipboard` struct somewhere, that could be used to deduplicate these lines.

bugadani (commented):
This is beyond awesome, thank you! Although the commit history will need some cleanup (looking at 6302365 and the merge commits, I would prefer a rebase instead), the implementation is really clean!

VolatileMark:
Thanks! Sorry about the commit history, I'll try cleaning it up

lapce-data/src/terminal.rs
VolatileMark: I could, but I would have to create an instance of the struct, I think (unless there already is one, and I can't see it).  
Actually, I might just do it, it helps readability.

VolatileMark:
Rebased

VolatileMark:
It needs a bit of testing, but I'd say it has all the features it needs.

lapce-ui/src/terminal.rs
```diff
@@ -519,9 +551,55 @@ impl Widget<LapceTabData> for LapceTerminal {
...
+                                term.scroll_display(Scroll::Bottom);
+                            }
+                            None => {}
+                        },
+                    }
+                } else {
+                    term.selection = None;
+                    if self.prev_click != None
+                        && (now - self.prev_click.unwrap()).as_millis()
+                            < DOUBLE_CLICK_TIMEOUT
```
dzhou121, line 583: You can use ```mouse_event.count``` to know if it's a double click, which should be easier.

VolatileMark: Good catch! I should be able to implement triple-click line select then!

lapce-core/src/command.rs
```diff
@@ -298,6 +298,11 @@ pub enum FocusCommand {
     Rename,
     #[strum(serialize = "confirm_rename")]
     ConfirmRename,
+    /// Terminal selection commands
+    #[strum(serialize = "terminal_copy_selection")]
+    TerminalCopySelection,
```
dzhou121, line 303: Any reason you don't want to reuse ClipboardCopy/ClipboardPaste?

VolatileMark: `ClipboardCopy` uses Ctrl+C which the terminal uses to interrupt the current task. `TerminalCopySelection` uses Ctrl+Shift+C which is the shortcut that basically any terminal emulator uses to copy the selection, along with Ctrl+Shift+V (`TerminalPasteSelection`) to paste for consistency.  
I hope my explanation makes sense, English isn't my native language.

dzhou121: Got you.

In that case, I would suggest to do this in the keymaps-nonmac.toml

```
[[keymaps]]
key = "ctrl+shift+c"
command = "clipboard_copy"
mode = "t"

[[keymaps]]
key = "ctrl+shift+v"
command = "clipboard_paste"
mode = "t"
```

And you don't need to add anything for keymaps-mac.toml because ```meta+c``` ```meta+v``` would just work on mac.

VolatileMark: Just tried that, but it doesn't work for some reason. The corresponding match case at `lapce-data/src/terminal.rs:356` is not executed. Any ideas?  

> And you don't need to add anything for keymaps-mac.toml because meta+c meta+v would just work on mac.  

I'm pretty sure the terminal on macOS still uses meta+shift+c and meta+shift+v but don't quote me on that.

VolatileMark: I found the problem! Apparently, it's the `mode = "t"` line that causes the problem. Don't know what `mode = "t"` does, though.

dzhou121: That's because Lapce can be used as a modal editor.

There are a few modes in Lapce, and mode="t" means that key only takes effect when you are in "terminal mode".

VolatileMark: Okay, I found the problem. In `lapce-data/src/keypress/loader.rs` at line 104, there's a check which excludes all the mappings which are not active in insert mode (unless modal editing is enabled).  
Adding an extra condition to check if the mapping is also **not** active in terminal mode fixes the problem, however I don't know if this would break the intended behavior.

franklinblanco:
This is such an important feature, love to see it go forward. Attempting to fix the failing CI

lapce-core/src/command.rs
dzhou121: I think it should check terminal mode as well.

franklinblanco:
Ci is failing because of this:
Screen Shot 2022-09-12 at 1 35 19 PM
`3 | _ => ` can be substituted by `_ => `

franklinblanco:
@VolatileMark `cargo clippy -- -D warnings`
Thank you for your contrib!

VolatileMark:
Thanks!

lapce-core/src/command.rs
VolatileMark: Alright, I'll commit the changes with the extra check for now.

dzhou121:
Thank you so much!

PR #1165, opened 2022-09-11
feat: add command to reveal item in system file explorer

panekj:
Doesn't actually reveal/select item, since it's currently not possible to do it in druid, but it opens directory fine.

### DIFF
--- a/lapce-data/src/command.rs
+++ b/lapce-data/src/command.rs
@@ -205,6 +205,10 @@ pub enum LapceWorkbenchCommand {
     #[strum(message = "Open File")]
     OpenFile,
 
+    #[strum(serialize = "reveal_active_file_in_file_explorer")]
+    #[strum(message = "Reveal Active File in File Explorer")]
+    RevealActiveFileInFileExplorer,
+
     #[strum(serialize = "change_theme")]
     #[strum(message = "Change Theme")]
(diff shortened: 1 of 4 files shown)


panekj:
> > Doesn't actually reveal it
> 
> is this a bug in druid, or a limitation?

It's just not implemented since one could argue it's not in scope of GUI library
Definitely not a bug

bugadani:
> Definitely not a bug

Then I'm not exactly sure what's the point of this PR. I would like to avoid nonfunctional UI elements.

bugadani:
> Doesn't actually reveal it

That suggests this PR doesn't do anything. Is this actually working, but the description is misleading?

panekj:
It's working but it doesn't select item in file explorer

bugadani:
Oh that's fine, apologies then.

panekj:
My point is that this should be done via native platform API (e.g. Windows https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shopenfolderandselectitems) but it's not implemented in druid and I think it would be best if it was there together with `OpenURI` logic. Someday I might implement it there.
For now it's implemented in a "hack way" by passing path to system to open it via default application.

bugadani (approved):

PR #1179, opened 2022-09-12
Window Tab improvements

bugadani:
- Prevent window tabs "sticking" when released
 - Fix undraggable area right of tabs
 - Turn magic numbers into constants
 - Prevent dragged tab overdrawing others
 - Swap tabs if dragged tab's center gets inside another tab (instead of mouse). This might be a controversial change.

Window tab code is a bit confusing as it's not obvious at all, what handles tab drags, for example. There are a number of small bugs around this area, this PR (when finished) attempts fixing those I found.


bugadani:
The code could use some more work, but I'll stop here for now, let's not create a bigger PR than necessary. The PR already implements the fixes for the issues I noticed.

PR #1191, opened 2022-09-13
Tone down inlay hint background color in lapce dark theme

bugadani:
Before:
image

After:
image

I didn't touch the light theme as I think it's less obnoxious there. Disclaimer: I'm not a designer, so let's put this up to a vote.

### DIFF
--- a/defaults/dark-theme.toml
+++ b/defaults/dark-theme.toml
@@ -79,7 +79,7 @@ magenta = "#C678DD"
 "editor.link" = "$cyan"
 
 "inlay_hint.foreground" = "$white"
-"inlay_hint.background" = "#528bFF88"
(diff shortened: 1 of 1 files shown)


bugadani:
Also, I've just randomly changed a few digits until it looked like this. If you have an actual color we should use, feel free to let me know :)

bugadani:
Thanks all, looks like nobody (so far) hates this change. Let's put it up for review :)

pwouik:
Screenshot from 2022-09-14 12-59-50
I personally prefer it even more distinct from code

bugadani:
@pwouik I hear you. We can tune the text color, too, but my goal with this PR was to make inlay hints at least usable on the default dark theme. I would leave further modifications to follow up PRs

Issue #1195, opened 2022-09-13
Improve editor tab close button


CreatorSiSo:
## My Problem

Lapce only shows the close icon when hovering over a tab.

![image](https://user-images.githubusercontent.com/64036709/190028276-61633ed8-0b42-43b2-9f8d-d4e185cd58e8.png)

## Possible Solution

I would like to see the close button on focused or hovered tabs like in a browser or vscode.

![image](https://user-images.githubusercontent.com/64036709/190026081-52a3c264-6d27-4b34-aa29-938498b6c000.png)

I would like to always see the unsaved icon on a tab with unsaved content unless the mouse hovers over the button, then it should show the close icon.

![image](https://user-images.githubusercontent.com/64036709/190026229-96a4a5b6-fb31-4bca-9245-bab1a82448cb.png)
![image](https://user-images.githubusercontent.com/64036709/190026732-4bd18793-fd97-4408-83c1-07b9296d2f64.png)
![image](https://user-images.githubusercontent.com/64036709/190026818-dce25b5f-6e52-4f81-8b1c-819085be9d41.png)
![image](https://user-images.githubusercontent.com/64036709/190026837-89d9c4c4-2da7-41b7-81fb-e14e2de1b10f.png)

## Describe alternatives you've considered

The unsaved icon could also be replaced with the close icon when the tab is hovered and not the actual close button.

PR #1196, opened 2022-09-13
Always show close button on focused tabs.

CreatorSiSo:
Implement changes to tab close button behavior mentioned in #1195.

### DIFF
--- a/lapce-ui/src/editor/tab.rs
+++ b/lapce-ui/src/editor/tab.rs
@@ -605,7 +605,7 @@ pub trait TabRectRenderer {
         ctx: &mut PaintCtx,
         data: &LapceTabData,
         widget_id: WidgetId,
-        i: usize,
+        tab_idx: usize,
         size: Size,
         mouse_pos: Point,
     );
@@ -617,7 +617,7 @@ impl TabRectRenderer for TabRect {
         ctx: &mut PaintCtx,
         data: &LapceTabData,
         widget_id: WidgetId,
-        i: usize,
+        tab_idx: usize,
         size: Size,
         mouse_pos: Point,
     ) {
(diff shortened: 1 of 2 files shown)


bugadani:
Good idea. Should we do this for Window Tabs, too?

ghishadow:
i wish this can become optional or have option to disable this, I hate too much buttons everywhere especially close button. I really like current clean look of tab

panekj:
> > Good idea. Should we do this for Window Tabs, too?
> 
> What exactly do you mean by window tabs?

Screenshot 2022-09-14 at 16 26 35

CreatorSiSo:
> > > Good idea. Should we do this for Window Tabs, too?
> > 
> > 
> > What exactly do you mean by window tabs?
> 
> Screenshot 2022-09-14 at 16 26 35

Woa had never seen them before, but yeah to me it would make sense on them as well.

panekj:
Code is in `lapce-ui/src/title.rs`, tabs can be created via palette with `Create new window tab` command

CreatorSiSo:
Thanks ill have a go at it.

CreatorSiSo:
So the implementation for the window tabs became a bit to complex for my first PR here:

The problem is that I dont know how to get the `current tab index` and `focused tab index` into the `LapceTabHeader::paint()` function (lapce-ui/src/tab.rs:2251). For the editor tabs the `tab index` is passed from the creation loop into the paint function (lapce-ui/src/editor/tab_header_content.rs:417). But for the window tabs `paint()` is part of the Widget trait so I cannot pass another value directly into it. If I was able to do that, I would simply pass in a tab_idx at lapce-ui/src/window.rs:693.

But I am not, which is why I would like to just leave it at this for now.

bugadani (approved):
LGMT, thank you again!

PR #1197, opened 2022-09-14
Auto Assign C-feature or C-bug when new issues are created

TwoThreeTwo:
This feature is a good way to auto assign labels making it easy to filter when new suggestions or bugs are coming through just with a filter selection on GitHub. You will not have to add them your selfs it will automatically add it.

Here is a video describing what it does.

https://user-images.githubusercontent.com/83801755/190030171-32dcdb7f-0ab7-4d49-83e0-696b69e9a938.mp4


panekj:
we use labels for distinguishing issues/features

TwoThreeTwo:
Of course, but having a way to quickly distinguish between a BUG and a FEATURE simply just reading the title as well as the labels can be much faster.

bugadani:
It would be better IMO if you could auto-set the C-bug or C-feature labels. That way the issues can be filtered with a single click. If visual differentiation is that important, we can set those to different colors, or differentiate them in some other way.

TwoThreeTwo:
Oh, yeah I can do that

TwoThreeTwo:
I have now added the:

`C-bug` label for auto assignment when they create an issue
`C-feature` label for auto assignment when they create a feature

I believe these should work, I cant really test it on my fork :c

TwoThreeTwo:
There is also the [FEATURE] and [BUG] included in the title, im not sure if you wanted me to remove that, but it's also there but if you do just send a request and ill do it right away.

bugadani:
I personally don't think we should have [FEATURE] and [BUG] in the title, but this might just be my preference. Maybe others will contribute their opinions to this, too.

TwoThreeTwo:
Yeah I see where you're coming from, I'll remove those

TwoThreeTwo:
It has been removed, if there's anything else you want me to add/delete I can do so.

panekj:
> I personally don't think we should have [FEATURE] and [BUG] in the title, but this might just be my preference. Maybe others will contribute their opinions to this, too.

I would like those removed (:

TwoThreeTwo:
https://github.com/lapce/lapce/pull/1197/commits/eae9de8eab2e2bcf96ae70c28d4650f60d581140

Will be auto assigning labels depending if it's a Feature, or a Bug.

TwoThreeTwo:
I believe it won't do it to every issue, but it will do it to the ones that are created if this is merged.

bugadani:
Thanks, let's see if this works :)

TwoThreeTwo:
👍

PR #1198, opened 2022-09-14
Focus current theme/language in palette

### LINKED ISSUE #642: Change Theme menu should focus on current theme
`Ctrl + P -> "Change Theme"` should have the active selection be the current theme set in the config rather than just the first one.  
This would be a minor improvement.

alichay:
Would close #642 

The implementation here is a little bit weird, but this is less trivial than it seems
* If you just set the index in `PaletteViewData::run`, it gets overridden on the next widget update cycle
* If you fix that, you still need to have the widget auto-scroll to the selected item. The usual scroll code only happens on update, not creation, and it only fires if the old index != the new index, so this also requires special-casing for when there's a default selection.

If someone has any pointers on how this could be cleaned up or made more idiomatic to the codebase (first contributions are always tricky!) I'd love to hear it. Thanks!

([especially this change to lapce-ui/palette.rs](https://github.com/lapce/lapce/blob/14beda90678e6dc0021a5c6bf0e49076e562fc6f/lapce-ui/src/palette.rs#L258) where Druid doesn't implement these shared methods with a trait, so I had to make the function generic over a closure that forwards to the actual method. it's pretty nasty)

### DIFF
--- a/lapce-data/src/palette.rs
+++ b/lapce-data/src/palette.rs
@@ -345,6 +345,7 @@ pub struct PaletteData {
     pub input: String,
     pub cursor: usize,
     pub index: usize,
+    pub has_nonzero_default_index: bool,
     pub items: Vec<PaletteItem>,
     pub filtered_items: Vec<PaletteItem>,
     pub preview_editor: WidgetId,
@@ -462,6 +463,7 @@ impl PaletteData {
             input: "".to_string(),
             cursor: 0,
             index: 0,
+            has_nonzero_default_index: false,
             items: Vec::new(),
             filtered_items: Vec::new(),
             preview_editor,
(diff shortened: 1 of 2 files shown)


lapce-data/src/palette.rs
```diff
@@ -345,6 +345,7 @@ pub struct PaletteData {
     pub input: String,
     pub cursor: usize,
     pub index: usize,
+    pub has_nonzero_default_index: bool,
```
bugadani, line 348: Can we turn `index` into an `Option` instead?

alichay: Index is used regardless of whether there's a default selection or not, so making it an Option would (imo) make everything a little harder to read and less clear.
I mean, even if there's no previous value, the item at index 0 defaults to being selected, right?

bugadani: Thanks :)

lapce-ui/src/palette.rs
```diff
@@ -319,7 +324,9 @@ impl Widget<LapceTabData> for PaletteContainer {
         if old_data.palette.input != data.palette.input
             || old_data.palette.index != data.palette.index
```
dzhou121, line 325: I think if you do

```
old_data.palette.input != data.palette.input
            || old_data.palette.index != data.palette.index
            || old_data.palette.run_id != data.palette.run_id
```

Then you don't need the ensure_item_visible in lifecycle.

I might be wrong. Can you please test.

alichay: Oh yeah that's much cleaner, thanks!

PR #1213, opened 2022-09-14
Use theme color for install plugin button

CreatorSiSo:
The backgound color for the install plugin buttons is currently hard-coded to `rgb(80, 161, 79)` but it should probably use a theme color.
| before | after |
|----------|--------|
| image | image |

I just used the `editor caret` color for now but it would probably make sense to add some sort of `primary button` theme color.

### DIFF
--- a/lapce-ui/src/plugin.rs
+++ b/lapce-ui/src/plugin.rs
@@ -1,9 +1,9 @@
 use crate::{panel::PanelSizing, scroll::LapceScroll};
 use druid::{
     piet::{Text, TextAttribute, TextLayout as PietTextLayout, TextLayoutBuilder},
-    BoxConstraints, Color, Command, Cursor, Env, Event, EventCtx, FontWeight,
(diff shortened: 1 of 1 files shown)


panekj:
I'd prefer to leave it hardcoded green for install/installed, yellow for disabled for now
just like remote button in titlebar

CreatorSiSo:
> yellow for disabled for now

Does it become yellow when the plugin is disabled for you because for me it just stays green.

image

bugadani:
TIL github displays the actual color here, really cool: `rgb(80, 161, 79)` `#123456`

I would prefer creating generic theme colors/color names (like `BUTTON_AVAILABLE`, etc. [come up with better names than my sleepy brain, please]) instead, so we can allow theme authors to choose non-clashing colors, but not changing the colors at this moment.

CreatorSiSo:
Should I create a new issue where we can think about these general theme variables?

panekj:
I don't think there is need for that, you can just prefix action/status with `LAPCE_EXTENSION_BUTTON_*`

PR #1219, opened 2022-09-15
Fixed a bug with the 'Append' command (Fixes #1193)

### LINKED ISSUE #1193: Vim command `a` incorrect behaviour
## Lapce Version

0.2.0

## System information

Ubuntu 22.04.1

## Describe the bug

When `a` (the `Append` command) is pressed from `Normal` mode while the cursor is on an empty line, the cursor will move down to the next line.

## Additional information

Not sure what could be causing this; in the source it seems like the cursor is only moved right, not up or down.

LukaOber:
The `Append` command will no longer move down to the next line when it is used on an empty line.

### DIFF
--- a/lapce-core/src/editor.rs
+++ b/lapce-core/src/editor.rs
@@ -1250,7 +1250,9 @@ impl Editor {
                 vec![]
             }
             Append => {
-                let offset = buffer.move_right(cursor.offset(), Mode::Insert, 1);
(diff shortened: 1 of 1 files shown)


LukaOber:
I might be able to implement a few tests for the modal editing on the weekend. But I think the small change in this pull request can be merged before then.

PR #1234, opened 2022-09-16
Automatic line height

### LINKED ISSUE #1230: Automatic Line Height

When adjusting the font size, line height doesn't change, leaving the fonts overlapping or spaced out. Having to manually calculate the ratio (which is not even a thing that I remember) shouldn't be necessary. Also, this looks like a bug to someone who isn't willing to dig around the settings.

Automatic line height. Maybe what VS Code does:
> Editor: Line Height
> Controls the line height.
>
> Use 0 to automatically compute the line height from the font size.
> Values between 0 and 8 will be used as a multiplier with the font size.
> Values greater than or equal to 8 will be used as effective values.

A suggested ratio, but it's nowhere as good as automatic.

bugadani:
Closes #1230

1.35 might be too tight, so we might want to either make it configurable directly, or provide a few presets. I think this is VSCode's default and I personally find this a good, comfortable spacing.

This PR also removes the assumption that problem panel lines are 25px high.

### DIFF
--- a/lapce-data/src/config.rs
+++ b/lapce-data/src/config.rs
@@ -166,9 +166,9 @@ pub struct EditorConfig {
     #[field_names(desc = "Set the font size in the code lens")]
     pub code_lens_font_size: usize,
     #[field_names(
-        desc = "Set the editor line height. If 0, line height will be calculated automatically"
+        desc = "Set the editor line height. If less than 5.0, line height will be a multiple of the font size."
     )]
-    line_height: usize,
+    line_height: f64,
     #[field_names(desc = "Set the tab width")]
     pub tab_width: usize,
     #[field_names(desc = "If opened editors are shown in a tab")]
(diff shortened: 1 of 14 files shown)


lapce-data/src/config.rs
```diff
@@ -165,8 +165,10 @@ pub struct EditorConfig {
     pub font_size: usize,
     #[field_names(desc = "Set the font size in the code lens")]
     pub code_lens_font_size: usize,
-    #[field_names(desc = "Set the editor line height")]
-    pub line_height: usize,
+    #[field_names(
+        desc = "Set the editor line height. If 0, line height will be calculated automatically"
+    )]
+    line_height: usize,
```
dzhou121, line 171: We can adopt VSCode's way. Use a float here. If smaller than a certain number, it's a multiplier. Greater than a certain number, it's the actual line height value.

bugadani: We can't use a float here, a lot of code assumes integer size (or we have to round at some places, that change is fragile)

dzhou121: We still cast it to integer. It's just not nice that you can't specify a line height like 1.5.

bugadani: I see. We can do that, but we can also do that later :)

lapce-data/src/config.rs
```diff
@@ -226,6 +228,14 @@ pub struct EditorConfig {
 }
 
 impl EditorConfig {
+    pub fn line_height(&self) -> usize {
+        if self.line_height == 0 {
+            (self.font_size as f64 * 1.35).ceil() as usize
+        } else {
+            self.line_height
```
dzhou121, line 235: Another quick win here is to ```self.line_height.max(self.font_size)``` so that text doesn't overlap.

lapce-data/src/config.rs
bugadani: I've decided to give it a try, and it was simpler than I expected.

bugadani: I find 4.9 unusable so we might want to lower the limit, but `¯\_(ツ)_/¯`

PR #1235, opened 2022-09-16
Minor cleanups

bugadani:


Cargo.lock
```diff
@@ -3725,7 +3725,7 @@ dependencies = [
 [[package]]
 name = "psp-types"
 version = "0.1.0"
-source = "git+https://github.com/lapce/psp-types#b7680c844e8faa1b79c800210b4d5526771e7de0"
+source = "git+https://github.com/lapce/psp-types#b1c618e695447e51d6380464b84845f82baf8890"
```
dzhou121, line 3728: What's the difference?

bugadani: it compiles on my machine 😅

bugadani: something with the document_selector wasn't picked up without this

dzhou121: odd, works for me.

bugadani: Yeah I get it, because the old hash is actually the newer merge commit. I'll drop the change but I could not get it to compile for some reason.

panekj: I had issues with it as well but only when on different branch and stopped happening

Issue #1202, opened 2022-09-14
Completion Crash


MinusGix:
**Version**: 0.2  

Log:  

![image](https://user-images.githubusercontent.com/13157904/190167890-b46dc86d-7b38-41c3-b550-9bc9767a909b.png)

The issue is presumably in completion, with the index being bad and so crashing since it tries going out of bounds. We are probably not clearing the index properly somewhere? It would probably be for the best to make the function referenced ( https://github.com/lapce/lapce/blob/master/lapce-data/src/completion.rs#L326 ) just return an `Option<&_>` to avoid any risk of crashing at all.

lena64t:
The same problem, but with backtrace:

Lapce version: 0.2.0 - a41bf067659e36ac73e990c736f53d69c7bf88fb (current master)
Rustc version: 1.65.0-nightly (17cbdfd07 2022-09-13)
OS: Manjaro 22.0.0 (Linux 5.19.7-1-MANJARO) w/ KDE

Backtrace:
```
thread 'main' panicked at 'index out of bounds: the len is 2 but the index is 11', lapce-data/src/completion.rs:326:10
stack backtrace:
   0: rust_begin_unwind
             at /rustc/17cbdfd07178349d0a3cecb8e7dde8f915666ced/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/17cbdfd07178349d0a3cecb8e7dde8f915666ced/library/core/src/panicking.rs:142:14
   2: core::panicking::panic_bounds_check
             at /rustc/17cbdfd07178349d0a3cecb8e7dde8f915666ced/library/core/src/panicking.rs:84:5
   3: lapce_data::completion::CompletionData::current_item
   4: lapce_ui::completion::CompletionContainer::update_documentation
   5: <lapce_ui::completion::CompletionContainer as druid::widget::widget::Widget<lapce_data::data::LapceTabData>>::update
   6: druid::core::WidgetPod<T,W>::update
   7: <lapce_ui::tab::LapceTab as druid::widget::widget::Widget<lapce_data::data::LapceTabData>>::update
   8: <druid::widget::lens_wrap::LensWrap<T,U,L,W> as druid::widget::widget::Widget<T>>::update
   9: druid::core::WidgetPod<T,W>::update
  10: <lapce_ui::window::LapceWindow as druid::widget::widget::Widget<lapce_data::data::LapceWindowData>>::update
[...]

PR #1246, opened 2022-09-17
Unify Palette/Completion with List widget

MinusGix:
This PR adds a `List` widget, which can be used for displaying lists like the palette or the completion. It changes the current implementations to use the `List` widget.  
The general idea is that the `List` widget has some `items` that it renders. This is an `im::Vector` to make it cheap to clone, or replace (such as filtering it from the containing widget). It is given the focus commands (for list movement) by the containing widget. If it receives `ListSelect`, then it sends an event to the containing widget that an item was selected (and it sends that event if an entry was clicked), letting them handle it (and also decide whether to close the list at all).  
With this, it will be easier to make simple lists. I plan on using it to implement a dropdown button element that we can use for the themes in the settings and in another PR change the git-branch selector to use a `List`.  
- I should perhaps just call `completion.run_focus_command` by itself rather than calling it for each list focus command
- This had to make so the `LapceCommand` implemented `PartialEq` so that we could have a `Data::same` implementation.

### DIFF
--- a/lapce-data/src/command.rs
+++ b/lapce-data/src/command.rs
@@ -50,13 +50,13 @@ pub const LAPCE_COMMAND: Selector<LapceCommand> = Selector::new("lapce.new-comma
 pub const LAPCE_UI_COMMAND: Selector<LapceUICommand> =
     Selector::new("lapce.ui_command");
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct LapceCommand {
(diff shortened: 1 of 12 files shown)


MinusGix:
Okay, I think this should be working now. I believe I pinned down the cause of the jump to reference bug I was having, though I'm not 100% sure since I had a hard time ensuring I could reproduce it.

MinusGix:
I believe this should also fix #1202, since I used options for accessing the currently selected option and modified the callers.

PR #1250, opened 2022-09-17
fix: too long socket

panekj:
Fixes issue when `{user}@{host}:{port}` results in too long socket path (max: ~100 characters)
`%C` is a SHA1 hash of `%l%h%p%r`

too long socket paths can happen if someone uses for example gitpod.io, e.g.:
`/Users/pj/.ssh/cm-lapce-lapce-dvqmoe3x0a0@lapce-lapce-dvqmoe3x0a0.ssh.ws-eu64.gitpod.io:22.F2aSCVIZ0IHmNSrh`

using `%C`
`/Users/pj/.ssh/cm_067698ecd110d603843a371f498e878e5d18fbaa=`

### DIFF
--- a/lapce-data/src/proxy.rs
+++ b/lapce-data/src/proxy.rs
@@ -622,15 +622,15 @@ struct SshRemote {
 }
 
 impl SshRemote {
-    #[cfg(target_os = "windows")]
+    #[cfg(windows)]
     const SSH_ARGS: &'static [&'static str] = &["-o", "ConnectTimeout=15"];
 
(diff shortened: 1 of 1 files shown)


coffee-is-power:
i don't think you need to use the whole SHA1, 5 characters are unique enough (like git does)

panekj:
> i don't think you need to use the whole SHA1, 5 characters are unique enough (like git does)

No, we should and if you looked at the code you would see it's not in our control

PR #1251, opened 2022-09-17
Added vims paste before 'P'

LukaOber:
`P` behaves like `p` but instead of inserting after the cursor `P` inserts before the cursor. `P` moves the cursor one to the left before pasting. This move does not occur if the cursor is at the beginning of the line and therefore behaves like vims `P`.

### DIFF
--- a/defaults/keymaps-common.toml
+++ b/defaults/keymaps-common.toml
@@ -384,6 +384,11 @@ command = "paste"
 mode = "nv"
 
 [[keymaps]]
+key = "shift+p"
+command = "paste_before"
+mode = "nv"
+
+[[keymaps]]
 key = "shift+j"
 command = "join_lines"
 mode = "n"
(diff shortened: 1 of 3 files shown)


LukaOber:
The unsuccessful check was due to a network error. It has nothing to do with the added code.

MinusGix:
> The unsuccessful check was due to a network error. It has nothing to do with the added code.

Thanks for mentioning it; I told the check to re-run.

LukaOber:
Sorry for reopening this. I forgot about this PR and deleted my head repository. Is there anything holding this PR back or did it just go under? @MinusGix

panekj:
It's waiting for @dzhou121 to review it

erf:
With this fix, if the cursor is at the beginning of the line it pastes after the first char. It think it should paste at the very beginning of the line. That is how it works in `vim`.

LukaOber:
Okay, I will fix this later today or on the weekend.

PR #1257, opened 2022-09-18
change terminal toggle shortcut

esmaeelE:
change terminal shortcut to ctrl + j as ` not work in many keyboards.


MinusGix:
You should be adding a second keybind rather than replacing the current entry, as Ctrl+` is the common shortcut in other editors I've used.  
I'd suggest checking what other editors use for 'focus terminal'. VSCode uses Ctrl+` with no other defaults, but perhaps other editors have some existing secondary keybind? I don't know if Ctrl+j is some existing common one. Ex: Kate uses F4.

panekj:
It would be preferable to discuss this first in a issue. Closing for now

esmaeelE:
> It would be preferable to discuss this first in a issue. Closing for now

OK thats correct. This is a true premature pull request by me.

PR #1261, opened 2022-09-18
Added Absolute/Relative copy Path,

Jerrody:


Cargo.lock
```diff
@@ -549,6 +549,28 @@ dependencies = [
...
+
+[[package]]
+name = "clipboard-win"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e3a093d6fed558e5fe24c3dfc85a68bb68f1c824f440d3ba5aca189e2998786b"
+dependencies = [
+ "winapi 0.3.9",
+]
+
```
panekj, line 573: Why?

panekj (changes requested):

Cargo.lock
Jerrody: Hm? I just added a `clipboard` crate for the write the content to OS's Clipboard.

panekj: We don't need more dependencies, clipboard is already provided via druid
https://github.com/lapce/lapce/blob/d695914acd737083561eac1bb17b916b53214901/lapce-data/src/document.rs#L58

Jerrody: > We don't need more dependencies, clipboard is already provided via druid
> 
> https://github.com/lapce/lapce/blob/d695914acd737083561eac1bb17b916b53214901/lapce-data/src/document.rs#L58

Oh, okay. Didn't know. Okay, I will fix this.

lapce-data/src/config.rs
```diff
@@ -168,7 +168,7 @@ pub struct EditorConfig {
     #[field_names(
         desc = "Set the editor line height. If less than 5.0, line height will be a multiple of the font size."
     )]
-    line_height: f64,
+    pub line_height: f64,
```
panekj, line 171: Doesn't have to be public

panekj (changes requested):

lapce-data/src/config.rs
Jerrody: > Doesn't have to be public

At this case I get a compilation error.

panekj: What error? I've just built HEAD without any issues

Jerrody: > What error? I've just built HEAD without any issues

I don't know. For the first time I got a compilation error. Anyway, now everything is ok!

PR #1262, opened 2022-09-18
Added Absolute/Relative Copy Path.

Jerrody:


lapce-ui/Cargo.toml
```diff
@@ -42,10 +42,14 @@ xi-unicode = "0.3.0"
...
-open = { version = "3.0.2", features = [] }
+open = "3.0.2"
 
 # lapce deps
-druid = { git = "https://github.com/lapce/druid", branch = "shell_opengl", features = [ "svg", "im", "serde", ] }
+druid = { git = "https://github.com/lapce/druid", branch = "shell_opengl", features = [
+    "svg",
+    "im",
+    "serde",
+] }
```
panekj, line 52: Please remove unnecessary reformatting

panekj (changes requested):

lapce-ui/Cargo.toml
Jerrody: > Please remove unnecessary reformatting
Alright.

lapce-ui/src/explorer.rs
```diff
@@ -686,6 +690,29 @@ impl Widget<LapceTabData> for FileExplorerFileList {
                             menu = menu.entry(item);
                         }
 
+                        menu = menu.separator();
+                        let path_to_file = node.path_buf.clone();
+                        let item = druid::MenuItem::new("Copy Absolute Path")
```
bugadani, line 695: VSCode calls this "Copy Path" and I also think "Absolute" is not really needed here.

Jerrody: Sometimes it's necessary. I think that is good when the user has an variants and can choose.

panekj: It should be:
`Copy path`
`Copy relative path`

bugadani: Sorry I meant the word isn't necessary, so the options should be "Copy Path" and "Copy Relative Path". This is what VSCode calls these, too.

Jerrody: > Sorry I meant the word isn't necessary, so the options should be "Copy Path" and "Copy Relative Path". This is what VSCode calls these, too.

Oh, okay. I missunderstood, well, I do not insist in this case, I just like to be meticulous.
Anyway, if you think that would be better to remove this word, welp.

panekj: I think we just meant the UI text, not internal code

Jerrody: > I think we just meant the UI text, not internal code

Then the uniformity of the code is lost.

PR #1265, opened 2022-09-18
Fix save-delay: Call callback of handle_server_request even though no plugins exist

### LINKED ISSUE #1128: CTRL+s is laggy
## Lapce Version
0.2.0

## System information
Windows 10

## Describe the bug
When I press CTRL+s to save a file, it takes 1 second to save the file (to see the white dot disapearing).

I open this issue because according to the website "any lag on your keystroke will be treated as a bug and get fixed."

tzAcee:
On handle_server_request no method was called to communicate with the server, so there is no callback executed, if there are no plugins installed.
The calls are always (as I can see) based on plugins.

So if there are no plugins, just execute the callback directly, without making a request to the server.

Fixes #1128


dzhou121:
Good fix. Thank you!

PR #1270, opened 2022-09-18
default-border-color: from black to gray

schwrzstrbn:
Gray border color looks better than black in both light and dark themes. It really blends well.

> Catpuccin Mocha
image

> Lapce Dark
image

> Lapce Light
image

### DIFF
--- a/lapce-ui/src/tab.rs
+++ b/lapce-ui/src/tab.rs
@@ -1953,6 +1953,9 @@ impl Widget<LapceTabData> for LapceTab {
                             Target::Widget(data.palette.widget_id),
                         ));
                     }
+                    if data.title.branches.active {
+                        Arc::make_mut(&mut data.title).branches.active = false;
+                    }
                     if data.focus_area == FocusArea::Rename {
                         ctx.submit_command(Command::new(
                             LAPCE_COMMAND,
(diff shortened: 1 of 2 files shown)


bugadani:
I personally prefer darker borders compared to the background color. In fact, I don't think I would like the light theme to change. Even though I don't actually use it, I think it looks pretty good as it is currently.

This might be a touchy subject :)

bugadani:
@schwrzstrbn I did not want you to close the PR, I simply expressed my opinion which can often be wrong. It's very much possible that others like lighter borders better and that there is sufficient interest in this to change the default, especially considering that for example, GitHub's dark theme also uses lighter borders.

schwrzstrbn:
> @schwrzstrbn I did not want you to close the PR, I simply expressed my opinion which can often be wrong. It's very much possible that others like lighter borders better and that there is sufficient interest in this to change the default, especially considering that for example, GitHub's dark theme also uses lighter borders.

It comes down to preference, I guess. The reason why I closed the pr is that, I realized that maybe it was too soon to suggest a change to the default without experimenting first.

schwrzstrbn:
Should I keep it open, just in case?

PR #1281, opened 2022-09-19
implement logic for getting the installation progress

### LINKED ISSUE #1266: Feature Request: Status of Plugin (Un-)Installation

If you install a plugin it is nice to see whether the click on the button successfully went to the GUI and the download started.
Also for bigger plugins in the future it's handy to know how long it will take, till the plugin is installed.

Especially for Uninstallation of plugins Im not sure if the process started or there were some kind of error and I have to spam click to verify (or the uninstallation is just slow, haven't figured it out yet).

Some kind of status when I click the install button.
Could be another window element or the fill percentage of the Install button (in sense of: install button as progess bar): 
image

Looking in the terminal logs what is happening, but there is also not much info.

## Additional context

n/a

fleexo:
- [x] Add logic for getting progress of the installation/uninstallation
- [x] Make plugin download async
- [x] Update the plugin data as soon as "install" call was made
- [x]  Update the plugin data as soon as "uninstall" call was made
- [x] Draw the plugin installation progress data to the UI
- [x] Show if any error occured while installing in the progress bar
- [x] Show if any error occured while uninstalling in the progress bar
- [x] Satisfy Clippy 

Currently looking like that:
The error message will disappear after 3 seconds automatically.
image
I think the design in the UI of the status-bar could be better, but maybe someone else will take a hand on that, the logic is implemented (it's currently just on top of the installed tab).

Fixes #1266

### DIFF
--- a/lapce-ui/src/tab.rs
+++ b/lapce-ui/src/tab.rs
@@ -988,6 +988,8 @@ impl LapceTab {
                                     &volt.display_name,
                                     error.to_string(),
                                 ),
+                            );
+                        }
                     }
                     LapceUICommand::VoltRemoved(volt, only_installing) => {
                         let plugin = Arc::make_mut(&mut data.plugin);


lapce-ui/src/tab.rs
```diff
@@ -945,7 +945,24 @@ impl LapceTab {
                     LapceUICommand::VoltInstalled(volt) => {
                         let plugin = Arc::make_mut(&mut data.plugin);
                         plugin.installed.insert(volt.id(), volt.clone());
-                    }
+
+                        // if there is a value inside the installing map, remove it from there as soon as it is installed.
+                        if plugin.installing.contains_key(&volt.id()) {
+                            plugin.installing.remove(&volt.id());
```
bugadani, line 951: I don't think you need to check existence before removing an element from an IndexMap.

tzAcee: yes you are right, I will change that!

bugadani:
I'm going to cc https://github.com/lapce/lapce/pull/1267 in case it gets merged first, I would suggest reusing the function introduced there.

lapce-ui/src/tab.rs
```diff
@@ -945,6 +946,25 @@ impl LapceTab {
...
+
+                        // if there is a value inside the installing map, remove it from there as soon as it is installed.
+                        plugin.installing.remove(&volt.id());
+                    }
+                    LapceUICommand::VoltInstalling(volt, progress) => {
+                        let plugin = Arc::make_mut(&mut data.plugin);
+
+                        if plugin.installing.contains_key(&volt.id()) {
+                            let elem = plugin.installing.get_mut(&volt.id());
+                            elem.unwrap().set_progress(*progress);
```
bugadani, line 958: This can be written as
```suggestion
                        if let Some(elem) = plugin.installing.get_mut(&volt.id()) {
                            elem.set_progress(*progress);
```

I believe this is a bit cleaner.

tzAcee:
> I'm going to cc https://github.com/lapce/lapce/pull/1267 in case it gets merged first, I would suggest reusing the function introduced there.

No problem, I will just merge the changes into my branch as soon as PR 1267 is on merged master and use the function then

Jerrody:
Great job!

lapce-data/src/plugin/plugin_install_status.rs
```diff
@@ -0,0 +1,42 @@
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum PluginInstallType {
```
bugadani, line 2: This could be `Copy` (with corresponding cleanups), and should use CamelCase naming

lapce-ui/src/tab.rs
```diff
@@ -945,27 +946,76 @@ impl LapceTab {
...
+                        let plugin = Arc::make_mut(&mut data.plugin);
+                        let id = volt.id();
+
+                        // if there is a value inside the installing map, remove it from there as soon as it is installed.
+                        plugin.installing.remove(&volt.id());
+
+                        if !(*only_installing) {
+                            plugin.installed.remove(&id);
+
+                            if plugin.disabled.contains(&id) {
```
bugadani, line 1005: You don't need the `contains` check. [remove](https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.remove) will tell you if the element was present

lapce-ui/src/tab.rs
```diff
@@ -945,27 +946,76 @@ impl LapceTab {
...
+                                plugin.disabled.remove(&id);
+                                let _ = data.db.save_disabled_volts(
+                                    plugin.disabled.iter().collect(),
+                                );
+                            }
+                            if plugin.workspace_disabled.contains(&id) {
```
bugadani, line 1011: Same here

bugadani (changes requested):
I only did a preliminary review, I have pointed out a few very minor opportunities where the code can be simplified

lapce-ui/src/tab.rs
bugadani: Sorry I might have miscommunicated. You still need the `if` to not save data unnecessarily. Like `if plugin.disabled.remove(...) {`

tzAcee: Oh oops, yes of course!

dzhou121:
Sorry can you fix the conflict again and I'll merge.

tzAcee:
> Sorry can you fix the conflict again and I'll merge. 

Yes, I'll fix them as soon as Im home from work!

tzAcee:
@dzhou121 alright fixed the conflicts.
Btw. if there's an error while removing a plugin, the plugin isn't removed anymore from the installed plugins list. That was not the case before (so the folder of the plugin still persisted, while it wasn't showing up in the plugins list). I think that's the behaviour how it should be.

Issue #249, opened 2022-03-24
Feature request: Render whitespace


bugadani:
Just like in VSCode, it would be nice to have a way to display whitespaces.

panekj:
Fixed in https://github.com/lapce/lapce/pull/1284

Issue #302, opened 2022-03-26
Feature request: Some settings items would be better as dropdown selectors


bugadani:
A good example is the theme selector - right now I have to type in the file name exactly, or I'll get the light theme, which is not an ideal user experience, when the set of themes is known in advance.

PhilipDaniels:
Seconded. I was just about to raise an issue saying exactly this.

panekj:
Fixed via https://github.com/lapce/lapce/pull/1938

PR #1284, opened 2022-09-19
Render whitespace

Nykseli:
Adding simple logic for whitespace rendering. Rendering logic is similar to VSCode.
Notably this doesn't take selected text into account. So this doesn't implement the `selection` whitespace rendering option.

Feature request: https://github.com/lapce/lapce/issues/249

### DIFF
--- a/defaults/settings.toml
+++ b/defaults/settings.toml
@@ -28,6 +28,7 @@ error-lens-font-size = 0
 blink-interval = 500 # ms
 multicursor-case-sensitive = true
 multicursor-whole-words = true
+render-whitespace = "none"
 
 [terminal]
 font-family = ""
--- a/lapce-data/src/config.rs
+++ b/lapce-data/src/config.rs
@@ -233,6 +233,10 @@ pub struct EditorConfig {
         desc = "Whether the multiple cursor selection only selects whole words."
     )]
     pub multicursor_whole_words: bool,
+    #[field_names(
+        desc = "How the editor should render whitespace characters.\nOptions: none, all, boundary, trailing."
+    )]
+    pub render_whitespace: String,
 }
 
 impl EditorConfig {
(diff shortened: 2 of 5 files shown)


bugadani (commented):
Thank you, this looks promising!

I have one nit to pick: would it be possible to replace the "push-in-loop" style appending with `string.extend(iter)` instead?

I'm also cc'ing #302 - if we'll ever have dropdowns, we could think about turning the setting into an enum and avoid string comparisons.

Jerrody:
> we could think about turning the setting into an enum and avoid string comparisons.

What about [hashing](https://doc.rust-lang.org/std/hash/trait.Hash.html)?

Nykseli:
> I have one nit to pick: would it be possible to replace the "push-in-loop" style appending with string.extend(iter) instead?

Line 2027:
```rust
for _ in whitespace_buffer.iter() {
    rendered_whitespaces.push(' ');
}
```
And line 2043:
```rust
whitespace_buffer
    .iter()
    .for_each(|c| rendered_whitespaces.push(*c));
```

Can be easily turned into
```rust
rendered_whitespaces.extend(whitespace_buffer.iter());
```

And I agree that it's a lot cleaner that way. Would that be enough? I think removing other "push-in-loops" require rewriting the whole logic.

> we could think about turning the setting into an enum and avoid string comparisons.

I agree with using enums. They are easy to implement, easy to understand and the comparisons are fast. And they even work well if/when the UI starts supporting translations.

bugadani:
> I think removing other "push-in-loops" require rewriting the whole logic.

Not really, I was only referring to those 3-liner snippets you quoted (I saw 3 of those I think). So yup, that's probably enough here :)

Jerrody:
> we could think about turning the setting into an enum and avoid string comparisons.

About this variant of using enum instead of strings.
Exists great crate for this purposes: [strum](https://crates.io/crates/strum).

I want to rewrite little bit a `LapceUICommand` enum to be easy to maintain a command's naming via enum's variant. Without look in many places of code an changing this name.

Nykseli:
Added a commit that uses extend where it makes sense.

> I saw 3 of those I think

There was 3 but one of those was pushing `n` amount of empty spaces (I actually copied that snipped in my previous comment) and I don't think there is cleaner/faster way to add `n` amount of spaces at the end of a string. I might be wrong though.

PR #1285, opened 2022-09-20
fix: Spelling mistake

xiaoxin-sky:
Although `FileWatchNotifer` can't prevent me  understand this struct ，I think use the correct word spelling can make the program more perfect.

### DIFF
--- a/lapce-proxy/src/plugin/mod.rs
+++ b/lapce-proxy/src/plugin/mod.rs
@@ -49,7 +49,7 @@ use crate::buffer::language_id_from_path;
 use crate::directory::Directory;
 
 use self::catalog::PluginCatalog;
-use self::psp::{ClosableCallback, PluginServerRpcHandler, RpcCallback};
+use self::psp::{ClonableCallback, PluginServerRpcHandler, RpcCallback};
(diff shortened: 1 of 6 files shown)


bugadani (approved):

lapce-proxy/src/plugin/mod.rs
```diff
@@ -49,7 +49,7 @@ use crate::buffer::language_id_from_path;
 use crate::directory::Directory;
 
 use self::catalog::PluginCatalog;
-use self::psp::{ClonableCallback, PluginServerRpcHandler, RpcCallback};
+use self::psp::{ClosableCallback, PluginServerRpcHandler, RpcCallback};
```
bugadani, line 52: I'm pretty sure Clonable is correct here

bugadani (changes requested):

lapce-proxy/src/plugin/mod.rs
xiaoxin-sky: sorry, This is my mistake.I will change this error.

bugadani (approved):
Thanks :)

PR #1291, opened 2022-09-20
fix: use link colour for empty editor buttons

panekj:
Should give an idea to users that those can be clicked


bugadani:
A screenshot would be nice to see what changes exactly.

panekj:
current: Screenshot 2022-09-20 at 21 32 18
pr: Screenshot 2022-09-20 at 21 34 07

bugadani (approved):

bugadani:
I like the spirit of the change but they now look out of place a bit.

PR #1314, opened 2022-09-22
Cases get handled correctly when generating keyboard shortcuts from UI

### LINKED ISSUE #1312: Keyboard shortcut generated from UI does not handle case correctly
## Lapce Version

`0.2.0`

## System information

Linux
archlinux 5.19.9-arch1-1
x86_64 GNU/Linux

## Describe the bug

Generating a key-mappings from the UI produce the following entry in `keymap.toml` : 

```toml
[[keymaps]]
command = "move_line_up"
mode = "i"
key = "Ctrl+Shift+U"
```
This does not trigger "move_line_up". 

Do get it working I need to manually edit `keymap.toml`: 
key = "Ctrl+Shift+u"

## Additional information

This commonly includes:

- screenshots
- logs
- theories about what might be going wrong
- workarounds that you used

LukaOber:
Characters now get unshifted when creating keyboard shortcuts from the UI. Fixes #1312

### DIFF
--- a/lapce-ui/src/keymap.rs
+++ b/lapce-ui/src/keymap.rs
@@ -138,7 +138,8 @@ impl Widget<LapceTabData> for LapceKeymap {
             }
             Event::KeyDown(key_event) => {
                 if let Some((_keymap, keys)) = self.active_keymap.as_mut() {
-                    if let Some(keypress) = KeyPressData::keypress(key_event) {
+                    if let Some(mut keypress) = KeyPressData::keypress(key_event) {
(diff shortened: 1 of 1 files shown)


dzhou121:
It's probably better to make the key handling logic to be case insensitive.

LukaOber:
I agree. But I would still add the unshifting for the UI. The keymaps file would then have a consistent lowercase lettering scheme. But for manual editing, a case-insensitive logic would still be beneficial. If you want, I can take a look at that.

Issue #1283, opened 2022-09-19
Allow disabling auto-close-quotes https://github.com/lapce/lapce/pull/1189


dholth:
## Is your feature request related to a problem? Please describe.

It looks like auto-close-quotes was added in https://github.com/lapce/lapce/pull/1189

I will often add quotes to an existing piece of text. Which becomes, with auto-close quotes,

`foobar` -> `"foobar` -> `"foobar""`

Unfortunately I've never come across an editor that does auto-close in a way that works for me. For me it's more of a nuisance to notice the extra edit, delete and correct it than to just type it all out manually.

Add a flag to disable all auto-close of quotes, parenthesis, brackets, etc.

MinusGix:
Having an option to disable this would be useful; but I think this issue could also be used to track the bug mentioned above. Writing a function to check if the current line has an unclosed quote mark *already* would not be hard.

killianc3:
Feature added in #1338.

PR #1338, opened 2022-09-24
Added setting to disable automatic closing of matching pairs

killianc3:
This is an implementation of a parameter to disable the automatic closing, asked by this problem #1283, what do you think ?


panekj:
`auto-closing` is too vague, please rename it to something like `auto-closing-quotes` or similar

PR #1344, opened 2022-09-25
Replace Branch-Selector dropdown with a List

MinusGix:
This replaces the source control branch selector with a `List` widget, which allows: limiting the space it takes, adding more information (such as deviation from HEAD) in the future, and being more consistent.  
image

There could probably be some improvements in the appearance. It has no visual icons of extra information like the palette/completion, and so can appear a bit bland.

### DIFF
--- a/lapce-data/src/command.rs
+++ b/lapce-data/src/command.rs
@@ -500,6 +500,10 @@ pub enum LapceUICommand {
     ShowAlert(AlertContentData),
     ShowMenu(Point, Arc<Vec<MenuKind>>),
     ShowWindow,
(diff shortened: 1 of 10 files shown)


bugadani:
What do you think about placing the branch list related code outside of the title module? I'm starting to become allergic to 1k+ line files personally, and their (title-branch selector) relation is more incidental than logical IMO.

MinusGix:
> What do you think about placing the branch list related code outside of the title module? I'm starting to become allergic to 1k+ line files personally, and their (title-branch selector) relation is more incidental than logical IMO.

I'm not against it.

Issue #253, opened 2022-03-25
Add VIM-like block selection


dialtone:
One of the best features of VIM's modal editing is the ability to select blocks of text for copying or replacement like `yi(`/`ya(`: to copy everything inside/around () or `ci(`/`ca(`: to change everything inside/around (). There's of course more, but even starting with even something simple along these lines would be very nice.

PR #1353, opened 2022-09-25
feat(lsp): implement syntax aware selection

oknozor:
This PR add syntax aware selection (see: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_selectionRange)  

**Example:**
output

Note: I had to implement a `get_selection` function on `Cursor` which might seems strange but was needed to compare the current selection state with the active `textDocument/selectionRange` and reset the document state whenever the selection change. 

Also the current implementation support only single cursor selection but I think multiple cursor selection range might work with the lsp api. This could be implemented in a future PR. 

Let me know if anything needs to be changed.

### DIFF
--- a/lapce-data/src/document.rs
+++ b/lapce-data/src/document.rs
@@ -44,6 +44,7 @@ use xi_rope::{
     Interval, Rope, RopeDelta, Transformer,
 };
 
+use crate::selection_range::SelectionRangeDirection;
 use crate::{
     command::{InitBufferContentCb, LapceUICommand, LAPCE_UI_COMMAND},
     config::{Config, LapceTheme},
@@ -52,6 +53,7 @@ use crate::{
     find::{Find, FindProgress},
     history::DocumentHistory,
     proxy::LapceProxy,
+    selection_range::SyntaxSelectionRanges,
     settings::SettingsValueKind,
 };
 
@@ -416,6 +418,7 @@ pub struct Document {
     pub code_actions: im::HashMap<usize, CodeActionResponse>,
     pub inlay_hints: Option<Spans<InlayHint>>,
     pub diagnostics: Option<Arc<Vec<EditorDiagnostic>>>,
+    pub syntax_selection_range: Option<SyntaxSelectionRanges>,
     pub find: Rc<RefCell<Find>>,
     find_progress: Rc<RefCell<FindProgress>>,
     pub event_sink: ExtEventSink,
@@ -462,6 +465,7 @@ impl Document {
             find_progress: Rc::new(RefCell::new(FindProgress::Ready)),
             event_sink,
             proxy,
+            syntax_selection_range: None,
         }
     }
 
@@ -2670,5 +2674,25 @@ impl Document {
         });
         self.sticky_headers.borrow_mut().insert(line, lines.clone());
         lines
+    }
+
+    pub fn change_syntax_selection(
+        &mut self,
+        direction: SelectionRangeDirection,
+    ) -> Option<Selection> {
+        if let Some(selections) = self.syntax_selection_range.as_mut() {
+            match direction {
+                SelectionRangeDirection::Next => selections.next_range(),
+                SelectionRangeDirection::Previous => selections.previous_range(),
+            }
+            .map(|range| {
+                let start = self.buffer.offset_of_position(&range.start);
+                let end = self.buffer.offset_of_position(&range.end);
+                selections.last_known_selection = Some((start, end));
+                Selection::region(start, end)
+            })
+        } else {
+            None
+        }
     }
 }
--- a/lapce-data/src/editor.rs
+++ b/lapce-data/src/editor.rs
@@ -21,6 +21,7 @@ use crate::keypress::KeyPressFocus;
 use crate::palette::PaletteData;
 use crate::proxy::path_from_url;
 use crate::rename::RenameData;
+use crate::selection_range::SelectionRangeDirection;
 use crate::{
     command::{
         EnsureVisiblePosition, InitBufferContent, LapceUICommand, LAPCE_UI_COMMAND,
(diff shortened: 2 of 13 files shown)


oknozor:
Closes #253

lapce-data/src/document.rs
```diff
@@ -2671,4 +2675,33 @@ impl Document {
...
+            };
+
+            let region = range.map(|range| {
+                let start = self.buffer.offset_of_position(&range.start);
+                let end = self.buffer.offset_of_position(&range.end);
+                SelRegion::new(start, end, None)
+            });
+
+            let mut selection = Selection::new();
+            if let Some(region) = region {
```
bugadani, line 2696: Would `region.map()` be a bit cleaner here? (Maybe not). Also, you create a `selection` one line up that you may not end up using.

Maybe this piece of code would look better as one single big `range.map(...)` or two maps.

To be clear, this is good enough, just a bit inconsistent in style :)

lapce-data/src/editor.rs
```diff
@@ -2213,11 +2216,77 @@ impl LapceEditorBufferData {
...
+                        LAPCE_UI_COMMAND,
+                        LapceUICommand::ApplySelectionRange {
+                            rev,
+                            buffer_id,
+                            direction,
+                        },
+                        Target::Auto,
+                    );
+                }
+                // Other wise, ask the LSP server for `textDocument/selectionRange`
```
bugadani, line 2261:
```suggestion
                // Otherwise, ask the LSP server for `textDocument/selectionRange`
```

lapce-data/src/selection_range.rs
```diff
@@ -0,0 +1,185 @@
...
+        rev: u64,
+        current_selection: Option<(usize, usize)>,
+    ) -> bool {
+        if self.last_known_selection.is_some() {
+            buffer_id == self.buffer_id
+                && rev == self.rev
+                && current_selection == self.last_known_selection
+        } else {
+            buffer_id == self.buffer_id && rev == self.rev
+        }
```
bugadani, line 40: Would something like this be too cryptic? 😅

```suggestion
        if buffer_id == self.buffer_id
                && rev == self.rev
                && (!self.last_known_selection.is_some()  || current_selection == self.last_known_selection)
```

lapce-data/src/selection_range.rs
```diff
@@ -0,0 +1,185 @@
...
+                Some(self.ranges.range)
+            } else {
+                let mut current = self.ranges.parent.as_ref();
+                for _ in 1..index {
+                    if let Some(parent) = current.and_then(|c| c.parent.as_ref()) {
+                        current = Some(parent)
+                    } else {
+                        current = None;
+                        break;
+                    }
```
bugadani, line 78: I'd like a comment explaining this. Also, would the following be equivalent?

```suggestion
                    current = current.and_then(|c| c.parent.as_ref());
                    if current.is_none() {
                        break;
                    }
```

lapce-data/src/selection_range.rs
```diff
@@ -0,0 +1,185 @@
...
+            }
+        })
+    }
+
+    fn count(&self) -> usize {
+        let mut count = 1;
+        let mut range = &self.ranges;
+        while let Some(parent) = &range.parent {
+            count += 1;
+            range = parent
```
bugadani, line 90:
```suggestion
            range = parent;
```

lapce-ui/src/editor.rs
```diff
@@ -1890,62 +1891,159 @@ impl Widget<LapceTabData> for LapceEditor {
...
-                                    LAPCE_UI_COMMAND,
-                                    LapceUICommand::RunCodeAction(action.clone()),
-                                    Target::Widget(editor_data.view_id),
-                                ));
-                                menu = menu.entry(item);
+                match cmd.get_unchecked(LAPCE_UI_COMMAND) {
+                    LapceUICommand::ShowCodeActions(point) => {
+                        let editor_data = data.editor_view_content(self.view_id);
+                        if let Some(actions) = editor_data.current_code_actions() {
+                            if !actions.is_empty() {
```
bugadani, line 1898: I think returning here if empty might end up looking better.

bugadani (commented):
I have a few styling comments, nothing else :)

lapce-ui/src/editor.rs
oknozor: This is not part of the PR, I think it got here because of cargo fmt. 
I could still add an early return, you tell me.

oknozor:
I have updated the PR according to the requested changes

lapce-data/src/selection_range.rs
bugadani: Still the common parts could be extracted into a variable, but no big deal :)

lapce-data/src/selection_range.rs
```diff
@@ -0,0 +1,182 @@
...
+                Some(self.ranges.range)
+            } else {
+                let mut current = self.ranges.parent.as_ref();
+
+                for _ in 1..index {
+                    current = current.and_then(|c| c.parent.as_ref());
+                }
+
+                current.map(|c| c.range)
+            }
```
bugadani, line 78: Much better, but if something like this would typecheck, we could remove one more branch. What do you think?

```suggestion
            let mut current = Some(self.ranges.as_ref());

            for _ in 0..index {
                current = current.and_then(|c| c.parent.as_ref());
            }

            current.map(|c| c.range)
```

oknozor: This wouldn't work without cloning, parent selection are Option<Box<T>> while root node is just T

bugadani: It's a ref, though, but I'll take your word on it

bugadani (approved):

Issue #1296, opened 2022-09-20
Cursor not shown when in sticky header


MinusGix:
**Version**: Current HEAD. 3a09f51.

## Describe the bug
![image](https://user-images.githubusercontent.com/13157904/191372794-cab7ed14-1563-4eb0-bedd-06d1b2815513.png)
When the cursor is moved to the sticky header, it is not rendered. Though, edits can still be made.

PR #1355, opened 2022-09-25
Sticky header fixes

bugadani:
cc #1296 - while this PR doesn't render caret, at least the first line is no longer always overdrawn. Also fixes a weirdness with the shadow.

This PR also optimizes background clearing when multiple sticky lines are visible. Nothing major, but still.

In general, I'm not pleased with how this PR looks but my brain doesn't work this late. Suggestions are welcome.

Before:

1px of overlap above shadow (on a high-DPI display, Surface Laptop Studio):
image

First line is always hidden below a sticky header:
image

After:

First line is no longer sticky except if scrolled:
image
image

Overlap is gone:
image

### DIFF
--- a/lapce-ui/src/editor.rs
+++ b/lapce-ui/src/editor.rs
@@ -1447,16 +1447,8 @@ impl LapceEditor {
             }
         }
 
-        let mut i = 0;
         let total_stickly_lines = sticky_lines.len();
-        if last_sticky_should_scroll {
-            info.last_y_diff = y_diff;
-        } else {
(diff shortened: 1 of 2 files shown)


dzhou121:
Any screenshots to show what does this PR fix?

bugadani:
Added screenshots. I'm pretty sure the +1 px "fix" is simply wrong but I couldn't find the actual reason...

dzhou121:
Thanks. The 1px could be GPU rendering. Not entirely sure.

PR #1357, opened 2022-09-26
Fix proxy hang due to held lock

MinusGix:
#1336 removed the block around variable for the `if let`.  
However, `if let Some(handler) = self.pending.lock().remove(&id)` would drop the lock at the end of the statement, which is after the if statement has ran.    
(The easiest way to trigger the hand for me was to remove or create a file in the file listing)


bugadani:
F me... sorry about that, I thought the temporary lock would be dropped immediately after `remove`. TIL.

PR #1358, opened 2022-09-26
feat(editor): add autosave configuration

### LINKED ISSUE #248: Feature request: Config to enable auto-saving files after a short period of inactivity, or loss of focus
Similar to how settings are saved, we could also save files if the user is done typing. The delay may or may not be configurable.

Auto-save should disable the "file changed" UI hint, to avoid distracting flashing.

oknozor:
This add an `autosave-interval` key to editor config, allowing to set up autosave. 
The config setting default to zero, which disable auto-saving. 

Closes #248

### DIFF
--- a/lapce-ui/src/editor/view.rs
+++ b/lapce-ui/src/editor/view.rs
@@ -42,6 +42,7 @@ pub struct LapceEditorView {
     pub editor: WidgetPod<LapceTabData, LapceEditorContainer>,
     pub find: Option<WidgetPod<LapceTabData, Box<dyn Widget<LapceTabData>>>>,
     cursor_blink_timer: TimerToken,
+    autosave_timer: TimerToken,
     last_idle_timer: TimerToken,
     display_border: bool,
     background_color_name: &'static str,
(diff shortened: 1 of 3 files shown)


bugadani:
Very nice, thank you!

On the other hand, I don't think configurability makes much sense here. A 1s delay should be good enough and allowing 1ms (which basically means save every keystroke) might be a bit harsh on flash storage, as well as probably a bad UX on spinning drives. But maybe I'm wrong :)

panekj:
> On the other hand, I don't think configurability makes much sense here.

I think it's reasonable to provide that config. VSCode also gives option to set it to 1ms

bugadani:
> VSCode also gives option to set it to 1ms

We only need to take what's actually good, though :) But if you think it's reasonable, I won't argue much.

lapce-ui/src/editor/view.rs
```diff
@@ -862,6 +897,16 @@ impl Widget<LapceTabData> for LapceEditorView {
             }
         }
 
+        if data.config.editor.autosave_interval > 0
+            && *data.focus == self.view_id
```
dzhou121, line 901: Shouldn't the timer be triggered when the document is changed rather than focus changed? 

So

 ```
editor_data.doc.rev() != old_editor_data.doc.rev()
```

oknozor: done

PR #1362, opened 2022-09-26
Log update errors

bugadani:


bugadani:
I undid my release related changes because I found out the update fn is not even called anywhere. Windows seems to be stuck on extracting the archive, so let's see why.

Issue #1309, opened 2022-09-22
Close search bar when escape is pressed regardless of current focus


oknozor:
## Is your feature request related to a problem? Please describe.

Currently the search bar (`Ctrl+f`) get closed when hitting the escape key only if it has focus. 

## Describe the solution you'd like

Search bar should close regardless of focus when pressing `esc` or cycling toggle when pressing `Ctrl+f`.

PR #1373, opened 2022-09-27
Improve search-bar workflow.

killianc3:
As described in this issue, #1309 the search bar (Ctrl+f) is only closed by pressing the Esc key if it has focus. So if you write code after opening a search, you have to click again to close it, which is not very usable. 

So I propose this simple modification which solves this problem by allowing you to close the search bar while writing code.

### DIFF
--- a/defaults/keymaps-common.toml
+++ b/defaults/keymaps-common.toml
@@ -292,7 +292,7 @@ mode = "i"
 
 [[keymaps]]
 key = "esc"
-command = "normal_mode"
+command = "clear_search"
 mode = "niv"
 when = "!search_focus && !modal_focus"
 


bugadani (changes requested):
> you have to click again to close it, which is not very usable.

To be fair, as a suboptimal workaround `ctrl-f` then `esc` works, too, no clicking required. 

My concern with this PR is that it breaks existing modal editing workflows as it is right now. Esc is used in modal editing to return to normal mode.

I'm going to mark this PR with "Request changes" so it won't get merged by accident.

killianc3:
Yes, you are right, this pr doesn't make sense, at the time, I didn't know this technique to close the search bar. But to come back to the issue #1309 , we can extend the ability to close the search in other modules, like in the terminal. Do you think this is useful? Or do you have a better idea ?

PR #1385, opened 2022-09-28
test: add code coverage CI action

oknozor:
Add code coverage to CI builds, this probably needs a maintainer to create a codecov account with lapce org  and authorize the github app.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,7 @@ jobs:
           profile: minimal
           toolchain: ${{ matrix.toolchain }}
           override: true
+          components: llvm-tools-preview
 
       - name: Cache Rust dependencies
         uses: Swatinem/rust-cache@v1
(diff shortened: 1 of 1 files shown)


oknozor:
cargo test was replaced with [cargo-llvm-cov](https://github.com/taiki-e/cargo-llvm-cov) which runs the test **and** produce code coverage. 

Additionally we could use nextest as a test runner (see: https://nexte.st/book/test-coverage.html) 
not sure if we want that.

.github/workflows/ci.yml
```diff
@@ -41,11 +42,20 @@ jobs:
...
+          command: llvm-cov
+          args: --all-features --workspace --lcov --output-path lcov.info
+
+      - name: Upload coverage
+        uses: codecov/codecov-action@v2
         with:
-          command: test
-          args: --workspace
+          files: lcov.info
+          fail_ci_if_error: true
```
bugadani, line 58: do we really want to fail CI?

bugadani: But this is the upload workflow 👀

oknozor: oh yes my bad, you are right.

oknozor: I removed it

PR #1386, opened 2022-09-28
Add test & rustdoc to lacpe-core

oknozor:


lapce-core/src/selection.rs
```diff
@@ -50,22 +53,48 @@ impl SelRegion {
...
-    pub fn end(self) -> usize {
-        self.end
-    }
-
+    /// A [`SelRegion`] is considered to be a caret when its start and end position are equal.
+    ///
+    /// **Example:**
+    ///
+    /// ```rust
+    /// # fn main() {
```
MinusGix, line 93: I think you don't need to wrap this in `fn main` if there's no return value (and no usage of `?`).

oknozor:
More documentation will be added in later PRs I think it's best to do this module by module instead of having one gigantic PR.

PR #1389, opened 2022-09-28
Simplify `if` statement.

Jerrody:


bugadani:
While technically correct, I wouldn't introduce a mutable variable that really doesn't have to be mutable. You can fix the weird formatting by extracting the `.len() - 1` part into a variable (called `last_index` for example) so the if can be formatted on a single line instead.

Jerrody:
> While technically correct, I wouldn't introduce a mutable variable that really doesn't have to be mutable. You can fix the weird formatting by extracting the `.len() - 1` part into a variable (called `last_index` for example) so the if can be formatted on a single line instead.

Do you mean this? I didn't understand correcrtly.
```rs
                            let mut new_index = editor_tab.active + 1;
                            let last_index = editor_tab.children.len();

                            if editor_tab.active == last_index - 1 {
                                new_index = 0;
                            }
```

Jerrody:
> Why not just
> 
> ```rust
> let new_index = (editor_tab.active +1) % editor_tab.children.len();
> ```

Pretty good solution :D

Hummel37 (approved):

Hummel37 (approved):

panekj:
While it's simpler (smaller) I don't think it's easier to understand what it does on first glance

panekj:
Closing since it's just a preference change without any impact

PR #1404, opened 2022-09-30
Log panics with full backtrace as errors

bugadani:
This PR captures and logs panics as errors. No backtrace is captured right now but it can be enabled using the `with-backtrace` feature flag and some configuring on the log-panic side.


bugadani:
At present this doesn't seem to work in `--release` builds, which kind of defeats the purpose. This needs some more investigating.

PR #1407, opened 2022-09-30
Fix #1405

### LINKED ISSUE #1405: Lapce won't compile using the most recent commit
## Lapce Version
0.2.0

## System information
Debian Linux Bullseye

## Describe the bug
Lapce won't compile. 61 errors
`
    |
102 |     pub const PANEL_HOVERED: &str = "panel.hovered";
    |                              ^ expected named lifetime parameter
`
are displayed.

## Additional information

This commonly includes:

workarounds that i used:
in lapce-data/src/config.rs:
`&'static str` instead of `& str`

eliasjonsson023:
Lapce won't compile with & str of public constants. Changed & str to &'static str. Now Lapce compiles again

### DIFF
--- a/lapce-data/src/config.rs
+++ b/lapce-data/src/config.rs
@@ -25,87 +25,87 @@ use crate::{
 
 pub use lapce_proxy::APPLICATION_NAME;
 
-const DEFAULT_SETTINGS: &str = include_str!("../../defaults/settings.toml");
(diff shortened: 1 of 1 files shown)


panekj (changes requested):
this change is unnecessary, as it's an issue with your local rust install

bugadani:
Constants have an implied `'static` lifetime. It is unfortunate that removing the lifetimes had some adverse effect, but the code itself is legal and a compiler update and `cargo clean` should solve it.

PR #1414, opened 2022-09-30
Volt activation

dzhou121:


Cargo.lock
```diff
@@ -620,15 +629,34 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
 ]
 
+[[package]]
+name = "config"
+version = "0.13.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "11f1667b8320afa80d69d8bbe40830df2c8a06003d86f73d8e003b2c48df416d"
+dependencies = [
+ "async-trait",
+ "json5",
```
bugadani, line 648: Can we disable the json5 dependency? it brings in an absolute hell of cryptographic crates

bugadani (changes requested):

Cargo.lock
panekj: I agree

Issue #782, opened 2022-07-16
feat: tab indent option


deltanedas:
option to indent using tabs would be cool
instead of space * `tab size`, indent with a tab and display as `tab size` wide

panekj:
If your code uses tabs, lapce automatically indents with tabs

deltanedas:
then to create a new file you have to use another editor first

panekj:
Why? It already uses proper indents even if you create new file from lapce.

deltanedas:
![12:59:25](https://user-images.githubusercontent.com/39013340/179353999-ee4d87ee-8b27-428a-8bfa-5f59434b3fa9.png)
On 0.1.3 at least, it does not.

PR #1419, opened 2022-09-30
Add atomic soft tabs feature

### LINKED ISSUE #1402: Atomic tabs
Atom and VSCode and a few other editors have a feature where you can make soft tabs act as if they are hard tabs, so you can't click in the middle of them or use arrow keys to move into them. It brings *most* of the benefits of hard tabs to those unfortunate enough to have to use soft tabs. The only thing you still don't get is the ability to set the tab width, although now that I think about it I guess you could do that too in theory!

Anyway if someone points me in the rough direction of the code that says "if you press left or click here where should the cursor go" then I can implement this.

[Here's the feature in VSCode with an animation](https://code.visualstudio.com/updates/v1_52#_sticky-tab-stops-when-indenting-with-spaces)

Timmmm:
This adds an editor setting to treat soft tabs (i.e. spaces pretending to be tabs) as if they were real tabs, for the purposes of mouse selection and left/right movement.

This feature is known as "sticky tabs" in VSCode and "atomic tabstops" in Atom.

Fixes #1402

I have tested this manually but not added any unit tests (thought I would get some feedback first before putting in a ton of effort).

Also it uses the tab width setting but this is not synchronised with the indent style, so it may need to be changed a bit when #782 is fixed.

### DIFF
--- a/lapce-data/src/atomic_soft_tabs.rs
+++ b/lapce-data/src/atomic_soft_tabs.rs
@@ -64,5 +64,114 @@ fn count_spaces_from(buffer: &Buffer, from_offset: usize) -> usize {
         }
         space_count += 1;
     }
-    return space_count;
+    space_count
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_count_spaces_from() {
+        let buffer = Buffer::new("     abc\n   def\nghi\n");
+        assert_eq!(count_spaces_from(&buffer, 0), 5);
+        assert_eq!(count_spaces_from(&buffer, 1), 4);
+        assert_eq!(count_spaces_from(&buffer, 5), 0);
+        assert_eq!(count_spaces_from(&buffer, 6), 0);
+
+        assert_eq!(count_spaces_from(&buffer, 8), 0);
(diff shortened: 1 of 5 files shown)


dzhou121:
@Timmmm I think it's a great feature. Thanks for adding it!

lapce-data/src/atomic_soft_tabs.rs
```diff
@@ -0,0 +1,68 @@
...
+fn count_spaces_from(buffer: &Buffer, from_offset: usize) -> usize {
+    let mut cursor = xi_rope::Cursor::new(buffer.text(), from_offset);
+    let mut space_count = 0usize;
+    while let Some(next) = cursor.next_codepoint() {
+        if next != ' ' {
+            break;
+        }
+        space_count += 1;
+    }
+    return space_count;
```
bugadani, line 67: Please remove this `return` keyword here, otherwise LGTM :)

Timmmm:
Awesome! I will add some tests & fix the `return` tonight. :-)

killianc3:
Hey @Timmmm, you should run `cargo fmt --all` to correctly format your code, and then the CI / Rustfmt test will pass 👍

PR #1420, opened 2022-10-01
feat(lsp): add 'codeAction/resolve' support

oknozor:
Add support for lsp `codeAction/resolve` (see: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_codeAction)


lapce-proxy/src/plugin/mod.rs
```diff
@@ -805,6 +805,38 @@ impl PluginCatalogRpcHandler {
         );
     }
 
+    pub fn action_resolve(
+        &self,
+        item: CodeAction,
+        cb: impl FnOnce(Result<CodeAction, RpcError>) + Send + Clone + 'static,
+    ) {
+        let method = CodeActionResolveRequest::METHOD;
+        self.send_request_to_all_plugins(
```
dzhou121, line 814: It would be better to include the plugin_id in the code action response, and then only send to the same plugin_id here.

oknozor: updated

dzhou121: Excellent! Thanks.

PR #1421, opened 2022-10-01
Highlight matching brackets

### LINKED ISSUE #974: Highlight Matching Braces
Does lapce have any support for highlight matching braces,  parens, or square brackets?

zarathir:
This PR adds matching brackets highlighting to Lapce.

image

This is a WIP, but I got the basic functionality working so I thought I'd share this as a draft. 
The basic idea is to check if a bracket is at the cursor, looking for the closing one and highlighting them.

I'm not quite happy with the search of the matching bracket, but am currently out of ideas/inspiration for a different implementation. Maybe someone has a better idea, especially how to handle multiple scopes.

Feel free to discuss and comment.

Closes #974


dzhou121:
There should be some existing code for finding matching brackets that you can have a look/reuse.

nheuillet:
I'm not an expert, maybe @dzhou121 would know better than me, but I'm pretty sure Tree sitter can give us enough information for all scope (bracket, curly brackets, parenthesis, etc.)
All we'd have to do in theory would be to query the tree, then colorize

dzhou121:
> There should be some existing code for finding matching brackets that you can have a look/reuse.

e.g. here

https://github.com/lapce/lapce/blob/master/lapce-data/src/document.rs#L2515

zarathir:
Thanks for the pointer @dzhou121. I got it working.

https://user-images.githubusercontent.com/47316908/193461717-fad68db0-4f2a-40fd-a4e3-d8bb408c05f4.mp4


Should I add more pairs to match? Looks like [Helix](https://github.com/helix-editor/helix/blob/master/helix-core/src/match_brackets.rs#L5) is also matching on `<, >, ', "`. I think matching on `<` and `>` would be nice, but am not sure about the other symbols.

dzhou121:
I think let's keep the current pairs for this PR, and maybe explore other pairs in future PRs?

lapce-ui/src/editor.rs
```diff
@@ -1787,6 +1789,127 @@ impl LapceEditor {
...
+        if screen_lines.lines.is_empty() {
+            return;
+        }
+
+        let cursor_offset = data.editor.cursor.offset();
+
+        if let Some(char_at_cursor) = data.doc.buffer().char_at_offset(cursor_offset)
+        {
+            if is_bracket(char_at_cursor) {
+                if let Some(syntax) = data.doc.syntax() {
```
bugadani, line 1808: I think this would look better if you retrieved the syntax without pushing code to the right. What I have in mind is something like this:

```rust
let syntax = if let Some(syntax) = data.doc.syntax() {
    syntax
} else {
    return;
};

if is_bracket(char_at_cursor) {...
```

This way we could remove a duplicated condition and use the left side of our monitors a bit more :)

lapce-ui/src/editor.rs
```diff
@@ -1787,6 +1789,127 @@ impl LapceEditor {
...
+                                    );
+                                    Self::highlight_char(
+                                        ctx,
+                                        data,
+                                        screen_lines,
+                                        closing_bracket_offset,
+                                    );
+                                    break;
+                                } else {
+                                    continue;
```
bugadani, line 1848: This continue is redundant I think

lapce-ui/src/editor.rs
```diff
@@ -1787,6 +1789,127 @@ impl LapceEditor {
...
+                        Self::highlight_char(ctx, data, screen_lines, cursor_offset);
+                        Self::highlight_char(ctx, data, screen_lines, new_offset);
+                    }
+                }
+            } else {
+                if let Some(syntax) = data.doc.syntax() {
+                    let end_line = *screen_lines.lines.last().unwrap();
+                    let end = data.doc.buffer().offset_of_line(end_line + 1);
+
+                    for (indice, c) in
```
bugadani, line 1821:
```suggestion
                    for (index, c) in
```

bugadani:
Closes #974

nheuillet:
any chance this PR can be expanded to colorized matching pairs, and maybe scope line? I understand if you'd instead leave that for later, but I figured I might suggest it just in case

zarathir:
> I think let's keep the current pairs for this PR, and maybe explore other pairs in future PRs?

Sure 👍.

> any chance this PR can be expanded to colorized matching pairs, and maybe scope line? I understand  > if you'd instead leave that for later, but I figured I might suggest it just in case

I thought about this and I'll leave the PR as is and probably try implementing the colorized brackets/scopes next. It's going to take me some time to research how to best implement this and I think this is already a nice to have feature. Hope that's okay.

lapce-ui/src/editor.rs
zarathir: > This way we could remove a duplicated condition and use the left side of our monitors a bit more :)

Yeah, sorry. As a widescreen user I like to have code in the middle of my monitor to avoid a sore neck, but I get what you mean. I got carried away trying to avoid `.unwrap()`. :)

I'll fix this tonight 👍

nheuillet:
> I thought about this and I'll leave the PR as is and probably try implementing the colorized brackets/scopes next. It's going to take me some time to research how to best implement this and I think the bracket matching is already a nice to have feature. Hope that's okay.

Fair enough! Bracket matching is already amazing. Can't wait to see the next steps :)

lapce-ui/src/editor.rs
```diff
@@ -1782,6 +1785,124 @@ impl LapceEditor {
...
+            None => return,
+        };
+
+        if is_bracket(char_at_cursor) {
+            if let Some(new_offset) = syntax.find_matching_pair(cursor_offset) {
+                Self::highlight_char(ctx, data, screen_lines, cursor_offset);
+                Self::highlight_char(ctx, data, screen_lines, new_offset);
+            }
+        } else {
+            let end_line = *screen_lines.lines.last().unwrap();
```
dzhou121, line 1817: There's existing code for this logic as well.

See https://github.com/lapce/lapce/blob/master/lapce-data/src/document.rs#L2510

Essentially it tries to use tree-sitter to find a bracket first. They only accept a single bracket though, so you might want to extend those functions to work with an array of different brackets, and with some offset boundary as an option so that you only scan the things on your screen.

lapce-ui/src/editor.rs
```diff
@@ -1782,6 +1785,124 @@ impl LapceEditor {
...
+            Some(c) => c,
+            None => return,
+        };
+
+        let syntax = match data.doc.syntax() {
+            Some(syntax) => syntax,
+            None => return,
+        };
+
+        if is_bracket(char_at_cursor) {
```
dzhou121, line 1811: ```is_bracket``` can be changed to use the existing ```matching_pair_direction.is_some()```

JustForFun88:
> Should I add more pairs to match? Looks like [Helix](https://github.com/helix-editor/helix/blob/master/helix-core/src/match_brackets.rs#L5) is also matching on `<, >` ... I think matching on `<` and `>` would be nice ...

It would be very helpful!!! VSCode doesn't highlight them, and it's really painful for fairly complex types, even at the second nesting level. I don't want to use Helix either, because it doesn't show inferred types. Having both inferred types and `<`, `>` highlighting would be great!

zarathir:
> It would be very helpful!!! VSCode doesn't highlight them, and it's really painful for fairly complex types, even at the second nesting level. I don't want to use Helix either, because it doesn't show inferred types. Having both inferred types and `<`, `>` highlighting would be great!

I opened an issue to keep track of stuff to add (#1439), because I think I'll try to get this merged first and keep track of the rest for later.

lapce-ui/src/editor.rs
sokorototo: I think the `if let Some(_) = _ {}` syntax would look much nicer here

bugadani: either of those works

sokorototo: Yeah, just some stylistic differences

dzhou121:
LGTM. Thanks for your amazing work!

PR #1423, opened 2022-10-01
Fix multiple cursor offset after inserting opening pair

### LINKED ISSUE #1374: Multi-cursor with insertion of parentheses moves to wrong cursor position
**Version**: Current HEAD.

When typing with multicursor, typing `(` will sensible autocomplete to `()` at the cursor location.  
However, the cursors after the edit end up at incorrect positions.  
Before:
image
After typing `(`:
image

I believe the issue is that the cursor does not get shifted by both of the inserted parentheses, but only the first one. And it fails to shift the other cursors by both.

NiceSieve:
Code was editing the region before comparing it to the previously inserted region.
Also add doc to matching_pair_direction function.

This closes #1374

### DIFF
--- a/lapce-core/src/editor.rs
+++ b/lapce-core/src/editor.rs
@@ -232,7 +232,7 @@ impl Editor {
                 for region in selection.regions_mut().iter_mut().sorted_by(
                     |region_a, region_b| region_a.start.cmp(&region_b.start),
                 ) {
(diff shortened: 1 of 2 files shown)


NiceSieve:
Am writing a regression test right now, just want the CI checks in parallel :)

PR #1441, opened 2022-10-05
feat(search): add case sensitive search ui button

oknozor:
This add a case sensitive search button to the find box widget. 

output


lapce-data/src/terminal.rs
```diff
@@ -461,6 +461,10 @@ impl KeyPressFocus for LapceTerminalViewData {
                         );
                     }
                 }
+                FocusCommand::ToggleCaseSensitive => {
```
oknozor, line 464: Not sure what to do here, is search working in the terminal right now ?

PR #1455, opened 2022-10-08
Update the preview text in search tab when the actual texts are updated

hbina:


lapce-proxy/src/dispatch.rs
```diff
@@ -306,6 +305,75 @@ impl ProxyHandler for Dispatcher {
...
+                        proxy_rpc.handle_response(
+                            id,
+                            Err(RpcError {
+                                code: 0,
+                                message: "cannot create matcher".to_string(),
+                            }),
+                        );
+                    }
+                });
+            }
```
panekj, line 376: this could probably be:
```suggestion
                        if current_id
                            == WORK_ID.load(std::sync::atomic::Ordering::Relaxed)
                        {
                            proxy_rpc.handle_response(
                                id,
                                Ok(ProxyResponse::AppendSearchResponse {
                                    line_matches,
                                    filepath,
                                }),
                            );
                            return;
                        }
                    }
                    proxy_rpc.handle_response(
                        id,
                        Err(RpcError {
                            code: 0,
                            message: "cannot create matcher".to_string(),
                        }),
                    );
                    return;
                });
            }
```

hbina: :+1:

lapce-proxy/src/dispatch.rs
```diff
@@ -43,7 +43,6 @@ use crate::{
 
 const OPEN_FILE_EVENT_TOKEN: WatchToken = WatchToken(1);
 const WORKSPACE_EVENT_TOKEN: WatchToken = WatchToken(2);
-
 pub struct Dispatcher {
```
panekj, line 46: don't delete the line here

hbina: my OCD in shambles. I will try to include this revert with the Cargo.lock changes removal.

panekj: `git checkout -- Cargo.lock`

PR #1461, opened 2022-10-09
Load shell environment when launching from GUI

TitanNano:
When lapce attempts to launch sub-processes (i.e. a LSP) that depend on environment variables like a customized `PATH` it will fail as long as it has been launched from the GUI (at least on macOS). 

This is due to the fact that i.e. macOS does not load the user's shell environment when launching an app. This can be mitigated by starting a login shell and extracting all environment variables from it. A strategy described in [this comment](https://github.com/atom/atom/issues/18318#issuecomment-439175987) on the Atom Editor project.

I have tested this on macOS, and it resolved all issues of lapce not being able to find LSP executables like `pylsp`. For Linux, I do not know if it is actually necessary or if the correct environment is available to lapce there.

### DIFF
--- a/lapce-ui/src/app.rs
+++ b/lapce-ui/src/app.rs
@@ -39,6 +39,12 @@ pub fn build_window(data: &mut LapceWindowData) -> impl Widget<LapceData> {
 }
 
 pub fn launch() {
+    // if PWD is not set, then we are not being launched via a terminal
+    #[cfg(any(target_os = "macos", target_os = "linux"))]
+    if !std::env::var("PWD").is_ok() {
+        load_shell_env();
+    }
+
     let cli = Cli::parse();
     let pwd = std::env::current_dir().unwrap_or_default();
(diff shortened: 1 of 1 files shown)


panekj (changes requested):
I'm against this type of workaround

TitanNano:
@panekj
a. What do you not like about it?
b. what would you suggest as an alternative solution?

panekj:
a) the fact that it's messing with environment
b) not doing it at all, plugins should correct their behaviour

TitanNano:
> a) the fact that it's messing with environment 
Fair point. What about setting env variables on the `Command` when launching the subprocess?

> b) not doing it at all, plugins should correct their behaviour
I don't see how plugins can resolve this issue. Even when a plugin accepts a `serverPath` to, i.e. `/opt/homebrew/bin/pylsp` the LSP will then not be able to locate the python interpreter, as it has also been installed via Homebrew and is not in the macOS standard `PATH`. 

Lapce not having access to Homebrew installed executables is a high inconvenience for users on macOS.

PR #1469, opened 2022-10-10
fix: canonicalize paths

### LINKED ISSUE #1468: Top workspace indicator bar shows path with relative directory symbols (. and ..)
## Lapce Version

nightly-5e507aa

## System information

Zorin OS 16.1 (Ubuntu 20.04.1)

## Describe the bug

When opened from terminal with relative path, top workspace indicator bar shows path with relative directory symbols (. and ..)

### Steps to reproduce:

1. Create some directories
```bash
mkdir -p /a/b/c/d/e
```

2. Run lapce in b directory like this:
```bash
lapce /a/b/c/..
```
3. Bar on top shows:
image

<details>
<summary><h3>Other examples</h3></summary>
if you open lapce like this:

<code>
lapce /a/b/c/./../c/d/./././..
</code>
<br/>
The bar is also displaying all the dots:

image

This is also the case when opening it like so

image
image
</details>

panekj:
Fixes #1468

### DIFF
--- a/lapce-proxy/src/lib.rs
+++ b/lapce-proxy/src/lib.rs
@@ -67,7 +67,11 @@ pub fn mainloop() {
     let cli = Cli::parse();
     if !cli.proxy {
         let pwd = std::env::current_dir().unwrap_or_default();
-        let paths: Vec<_> = cli.paths.iter().map(|p| pwd.join(p)).collect();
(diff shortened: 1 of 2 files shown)


panekj:
I can't reproduce issue on mac but I think it won't hurt if we resolve all paths completely before passing them anywhere

PR #1471, opened 2022-10-10
feat: add command to un/install lapce from/to PATH

panekj:


dzhou121:
This deserves an entry in changelog

Issue #1474, opened 2022-10-10
Add setting for minimum number of visible line above and below cursor


maxwell8888:
## Is your feature request related to a problem? Please describe.

This is a common user prefence in editors. in VSCode the setting is "editor.cursorSurroundingLines". I personally use it by setting it to a very large value so that the cursor is permenantly fixed in the middle of the editor.

## Describe the solution you'd like

Same behaviour as VSCode. The setting takes an integer value which specifies the number of lines visible above and below the cursor.

## Describe alternatives you've considered

None

PR #1475, opened 2022-10-10
Feat: Add setting Cursor Surrounding Lines (#1474)

maxwell8888:
This is the minimum number of visible lines above and below the cursor. The default is 1 as per the editors current behavior.

- [ x] Added an entry to `CHANGELOG.md` if this change could be valuable to users


dzhou121:
Sorry can you fix conflict for CHANGELOG.md

PR #1477, opened 2022-10-10
fix(search): use esc to close search regardless of current mode

### LINKED ISSUE #1309: Close search bar when escape is pressed regardless of current focus

oknozor:
Closes #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Now use `esc` to close searchbar regarless of the current focus
+
 ### Features/Changes
 
 ### Bug Fixes
--- a/defaults/keymaps-common.toml
+++ b/defaults/keymaps-common.toml
@@ -131,7 +131,7 @@ mode = "i"
 [[keymaps]]
 key = "esc"
 command = "clear_search"
-when = "search_focus"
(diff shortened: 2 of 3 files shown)


panekj:
That change should be noted in changelog, but also I'm not sure if it should happen in modal mode

oknozor:
updated

dzhou121:
I feel like no editor that's doing this?

oknozor:
> I feel like no editor that's doing this?
vscode and  jetbrains are doing this.

dzhou121:
I would prefer normal_mode takes happens before clear_search

oknozor:
Ah I misunderstood I think, you were talking about modal modes. Sure I can change this.

oknozor:
updated so the clear search command is triggered only in normal mode when using modal editing

dzhou121:
Sorry I've released a new version. Can you fix the changelog conflict.

oknozor:
Done

dzhou121:
Change on CHANGELOG.md is gone?

PR #1483, opened 2022-10-12
UI: Fix issue where 1st tab shows the close icon...

hbina:
When there are multiple tabs open using the file explorer. This happens because the `mouse_pos` is initialized to `Point::ZERO` which technically resides inside the rect of the first tab.

Edit: Do you want to include this to `CHANGELOG.md`?

### DIFF
--- a/lapce-ui/src/editor/tab.rs
+++ b/lapce-ui/src/editor/tab.rs
@@ -603,7 +603,7 @@ pub trait TabRectRenderer {
         widget_id: WidgetId,
         tab_idx: usize,
         size: Size,
-        mouse_pos: Point,
+        mouse_pos: Option<Point>,
     );
 }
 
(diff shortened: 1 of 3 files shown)


lapce-ui/src/editor/tab.rs
```diff
@@ -624,8 +624,8 @@ impl TabRectRenderer for TabRect {
         let editor_tab = data.main_split.editor_tabs.get(&widget_id).unwrap();
 
         let rect = Size::new(width, height).to_rect().with_origin(Point::new(
-            self.rect.x0 + (size.height - width) / 2.0,
```
hbina, line 627: I don't really understand what the old code is trying to do here?

lapce-ui/src/editor/tab.rs
```diff
@@ -706,8 +707,10 @@ impl TabRectRenderer for TabRect {
             );
         };
 
-        if is_pristine || self.close_rect.contains(mouse_pos) {
-            if self.rect.contains(mouse_pos) || is_active_tab {
```
hbina, line 710: I took the liberty of removing one check here because `close_rect` is a subset of `rect`.

dzhou121:
Yes please for the changelog. You might want to rebase master first.

PR #1517, opened 2022-10-14
Add support for empty setting items

MinusGix:
This fixes a bug in my PR for handling settings which we don't currently have widgets for, but are still wanting to be rendered in the settings menu as options. (This occurred in plugins)  
This is just how it behaved before, displaying the name and description of the setting but without a way to edit it (without going into the setting.toml file).


oknozor:
Partially addresses #1519

PR #1521, opened 2022-10-14
Show unique paths to disambiguate same file names (Re: #557)

hbina:
Based on the work from https://github.com/lapce/lapce/pull/557.


lapce-ui/src/editor/tab_header_content.rs
```diff
@@ -456,3 +517,114 @@ impl Widget<LapceTabData> for LapceEditorTabHeaderContent {
         }
     }
 }
+
+fn get_truncated_path(full_paths: &[PathBuf]) -> Vec<PathBuf> {
```
dzhou121, line 521: Is this ```PathBuf.parent()```?

hbina: Not quite, you can imagine it as consuming all the given `Path`s in parallel from left and right until it finds a mismatch `Component`. Just a simple algorithm to find a short unique path to display. I can't find the algorithm that VSCode it using so this is the best I can think of.

lapce-ui/src/editor/tab.rs
```diff
@@ -653,6 +653,15 @@ impl TabRectRenderer for TabRect {
             &self.text_layout,
             Point::new(rect.x1 + 5.0, self.text_layout.y_offset(size.height)),
         );
+        if let Some(path_layout) = &self.path_layout {
+            ctx.draw_text(
+                path_layout,
+                Point::new(
+                    rect.x1 + 5.0 + self.text_layout.layout.width() as f64 + 5.0,
+                    path_layout.y_offset(size.height),
```
hbina, line 661: Not entirely sure this is correct because I think its not really aligned to center. IDK how this layout really works tho.

Issue #1529, opened 2022-10-15
Ctrl+Arrow Keys slightly overeager jumping


MinusGix:
## Lapce Version

Current HEAD

## Describe the bug
Ctrl+[Arrow Key] is behaving slightly incorrectly:
```rust
        .other_thing(blah)
        |.thing()
```
(where `|` is the cursor)
Then Ctrl+Left-Arrow,
```rust
        .other_thing(blah|)
        .thing()
```

What it should be doing is:
```rust
        .other_thing(blah)
        |.thing()
```
Ctrl+Left-Arrow
```rust
        .other_thing(blah)
|        .thing()
```
Ctrl+Left-Arrow
```rust
        .other_thing(|blah)
        .thing()
```

PR #1531, opened 2022-10-15
Improve Ctrl+Left-Arrow Keys behavior

killianc3:
As mentioned in the issue #1529, the Ctrl+Arrow keys behave slightly incorrectly, especially with regard to indentation where Ctrl+Left moves the cursor to the top line when it should move the cursor to the beginning of the line and the next time it should move to the top line
 
I don't know now but it looks like CHANGEMOD.md has been automated so I don't know if I should add an entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Features/Changes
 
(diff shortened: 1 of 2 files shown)


panekj:
> I don't know now but it looks like CHANGEMOD.md has been automated

It's not

> so I don't know if I should add an entry.

You should

PR #1549, opened 2022-10-17
Draft: Add paragraph_forward and paragraph_backward commands

yunmikun2:


rust-toolchain.toml
```diff
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "stable"
+channel = "nightly"
```
bugadani, line 2: This seems accidental, please remove

yunmikun2: My bad. Removed.

PR #1565, opened 2022-10-21
Several fixes to the About modal

hbina:
1. Fix an issue wherein the mouse cursor is always set to Pointer
2. Tighten the UI a bit so it doesn't overflow.

### DIFF
--- a/lapce-ui/src/about.rs
+++ b/lapce-ui/src/about.rs
@@ -115,7 +115,7 @@ impl Widget<LapceTabData> for AboutBox {
 }
 
 pub struct AboutBoxContent {
-    mouse_pos: Point,
+    mouse_pos: Option<Point>,
     widget_id: WidgetId,
 
     width: f64,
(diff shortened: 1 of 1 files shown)


lapce-ui/src/about.rs
```diff
@@ -446,9 +442,16 @@ impl Widget<LapceTabData> for AboutBoxContent {
...
             self.width / 2.0 + (row_item.layout.width() as f64 / 2.0)
                 - row_item.layout.width() as f64,
-            rect.y1 - row_item.layout.height() as f64 * 2.0,
+            y,
         ));
 
+        ctx.fill(
+            site_rect,
+            &data.config.get_hover_color(
+                data.config.get_color_unchecked(LapceTheme::TERMINAL_RED),
```
panekj, line 451: It's not terminal so don't use that

panekj (changes requested):

lapce-ui/src/about.rs
hbina: LOL I left it over while debugging. My bad.

PR #1566, opened 2022-10-21
multiple fixes

panekj:

### DIFF
--- a/lapce-ui/src/editor/tab.rs
+++ b/lapce-ui/src/editor/tab.rs
@@ -594,11 +594,9 @@ impl Widget<LapceTabData> for LapceEditorTab {
                     ctx.incr_alpha_depth();
                     ctx.fill(
                         rect,
-                        &data
-                            .config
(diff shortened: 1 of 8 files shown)


lapce-ui/src/editor/tab.rs
```diff
@@ -594,11 +594,9 @@ impl Widget<LapceTabData> for LapceEditorTab {
                     ctx.incr_alpha_depth();
                     ctx.fill(
                         rect,
-                        &data
-                            .config
-                            .get_color_unchecked(LapceTheme::EDITOR_CURRENT_LINE)
-                            .clone()
-                            .with_alpha(0.8),
+                        data.config.get_color_unchecked(
+                            LapceTheme::EDITOR_SPLIT_DROP_BACKGROUND,
```
dzhou121, line 598: How about ```LapceTheme::DRAG_DROP_BACKGROUND``` which could used by panel drag and drop

Issue #1398, opened 2022-09-29
UX issue when opening file from sidebar


snoopdouglas:
## Lapce Version

0.2.0

## System information

Fedora 36

## Describe the bug

When opening a file from the sidebar, the 'Open Editors' section expands immediately, pushing down the item you just clicked:

![image](https://user-images.githubusercontent.com/4892574/193006830-c5bfae4e-79b2-4319-8970-344f91f27fc4.png)

## Additional information

I don't find this often means that I'm accidentally opening the item above the one I originally clicked, but it doesn't feel right. So, I wonder whether we should either:

* Move 'Open Editors' to the bottom of the sidebar
* Only expand it once the mouse has left the sidebar (preferably with an animation so it's more obvious what just happened)
  * In this case, the 'Open Editors' header could 'blink' - and perhaps show a badge - to indicate that it's beginning to acquire items that've not yet been shown, ie.:

![image](https://user-images.githubusercontent.com/4892574/193008232-effd507c-9a32-4714-ad27-e4035a8dcdc8.png)

panekj:
This is exactly same behaviour as in VS Code

superlou:
As someone who habitually double-clicks to open a file, the shift always gets me to open two files. I guess I don't run into it in VSCode because I have open editors disabled:

![image](https://user-images.githubusercontent.com/709695/193026287-d6e8aed4-b727-4611-a4ae-0487afed3e69.png)

snoopdouglas:
Yeah, I'm in the same boat as @superlou. IMO, we shouldn't be aiming for UX-issue-for-issue parity with VS Code. 😉

NiceSieve:
> This is exactly same behaviour as in VS Code

I tried it in VSCode, and see a noticeable delay. Lapce is just too fast compared to VSCode 😉
There is also a second difference: VSCode has "half open" tabs (with italic title) for single clicked files, and "full open" tabs for double clicked or edited files. So the first half tab gives you a delay, and any further half tabs don't move the open editors section. The delay is what gives you that time double click.

vintprox:
It is a bit of a hurdle here switching to new exercises from `rustlings` course. 😆 I want this "Open Editors" panel gone altogether, as a preference.

JonahPlusPlus:
I made a fix for this by moving the "Open Editors" panel to be below the file tree and making it resizable.
I just have to make it so the default size is smaller and make it so that the split reacts when hovering.
[...]

PR #1574, opened 2022-10-22
Open editors list can now be collapsed by clicking it

hbina:
Partially solves: https://github.com/lapce/lapce/issues/1398

### DIFF
--- a/lapce-ui/src/explorer.rs
+++ b/lapce-ui/src/explorer.rs
@@ -588,7 +588,6 @@ impl Widget<LapceTabData> for FileExplorerFileList {
                         file_explorer.active_selected = path.clone();
                         ctx.request_paint();
                     }
-
                     LapceUICommand::FileExplorerRefresh => {
                         data.file_explorer.reload();
                     }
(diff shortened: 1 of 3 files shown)


dzhou121:
Thanks for adding this. Would you like to extend it for all ```PanelSectionHeader```?

panekj:
> Sure, but i should probably add some icons to indicate that its expandable. Something like a down_arrow.svg.

https://github.com/lapce/lapce/blob/master/icons/codicons/chevron-down.svg
https://github.com/lapce/lapce/blob/master/icons/codicons/chevron-right.svg

dzhou121:
I would suggest putting the size 0 logic in PanelSection.

Gerrit0:
Would be nice if this could be made persistent - I'd like to never have open editors show up, this would make me collapse the panel every time I open lapce.

hbina:
@dzhou121 I've gotten around to reimplementing it as you requested. It also comes with some cleanup on the array argument because the first element in that tuple was meaningless. Now it serves the purpose of enabling/disabling collapsing of the content. In this case, the warnings/errors panel cannot be collapsed because they are placed side by side.

dzhou121:
Thanks!

PR #1587, opened 2022-10-23
Add g++ in Ubuntu building from source guide

### LINKED ISSUE #1586: C++ missing when trying to build Lapce on Ubuntu
## Lapce Version

0.2.1

## System information

Ubuntu 22.04.1 LTS

## Describe the bug

Getting `error occurred: Failed to find tool. Is c++ installed?` when trying to build Lapce from source as in https://github.com/lapce/lapce/blob/master/docs/building-from-source.md

## Additional information

image

cglotr:
Fixes #1586

### DIFF
--- a/docs/building-from-source.md
+++ b/docs/building-from-source.md
@@ -8,7 +8,7 @@ It is easy to build Lapce from source on a GNU/Linux distribution. Cargo handles
 
 #### Ubuntu
 ```sh
-sudo apt install cmake pkg-config libfontconfig-dev libgtk-3-dev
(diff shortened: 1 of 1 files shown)


panekj:
We should probably just add `build-essential` package there

cglotr:
I see, so something like this right? `sudo apt install build-essential cmake pkg-config libfontconfig-dev libgtk-3-dev`

But I think just `g++` is better since I think we don't need the other [dependencies](https://packages.ubuntu.com/focal/build-essential)

Issue #1611, opened 2022-10-27
Focus snaps to the end of error lens message when it's too long


xiaoxin-sky:
## Lapce Version 
0.2.1
## System information
macos 11.6.6 
## Describe the bug
I input any thing , the Horizontal scroll bar  scroll to the Diagnosis msg last immediately, when the Diagnosis msg too long (over one window width)
## Additional 

https://user-images.githubusercontent.com/42181585/198323149-f061da3b-74a3-4cf0-9743-479f44aac213.mp4

information

xiaoxin-sky:
The bug has be fix. #1615

PR #1615, opened 2022-10-28
fix(editor): diagnosis cause cursor x calculate error

xiaoxin-sky:
I don't think this should take into account the width of the inline diagnosis when focusing the cursor, otherwise it will cause  #1611 issue，So I delete the diagnosis col.

### DIFF
--- a/lapce-ui/src/editor/view.rs
+++ b/lapce-ui/src/editor/view.rs
@@ -465,6 +465,8 @@ impl LapceEditorView {
     fn cursor_region(data: &LapceEditorBufferData, text: &mut PietText) -> Rect {
         let offset = data.editor.cursor.offset();
         let (line, col) = data.doc.buffer().offset_to_line_col(offset);
+        let inlay_hints = data.doc.line_phantom_text(&data.config, line);
+        let col = inlay_hints.col_after(col, false);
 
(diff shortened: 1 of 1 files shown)


xiaoxin-sky:
**Here is the modified preview**


https://user-images.githubusercontent.com/42181585/198687300-5ec992f1-54ad-4a8b-8a21-79a257565d18.mp4

lapce-ui/src/editor/view.rs
```diff
@@ -465,8 +465,6 @@ impl LapceEditorView {
     fn cursor_region(data: &LapceEditorBufferData, text: &mut PietText) -> Rect {
         let offset = data.editor.cursor.offset();
         let (line, col) = data.doc.buffer().offset_to_line_col(offset);
-        let inlay_hints = data.doc.line_phantom_text(&data.config, line);
-        let col = inlay_hints.col_at(col);
```
dzhou121, line 469: You probably need to change this to ```let col = inlay_hints.col_after(col, false);```

xiaoxin-sky: yes,I'm not check the `inlay_hints` code. I had change `col_at` to `col_after`.This is more suitable.And it work well。

PR #1620, opened 2022-10-30
feat: added show hover keyboard shortcut

m-rutter:
Issue: #1621

Added a new keyboard shortcut, that triggers the hover effect at the cursor location. Very useful for people that use vim-like bindings. The suggested shortcut, `g h`, is what is used by vscode's popular vim plugin, but some other popular defaults might be `ctrl + k`, `k`, `K` among others can be seen in the wild.

As a future improvement we might want to change how mouse events dismiss the hover as currently any mouse event that isn't over the cursor position or hover triggers the hover to become inactive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Features/Changes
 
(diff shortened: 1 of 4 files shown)


dzhou121:
Can you please fix the CHANGELOG conflict.

PR #1631, opened 2022-11-01
Add support for ini files and dotfile highlighting

Insprill:
Adds highlighting support for `.ini` files, and adds internal support for dotfiles. This was required to support `.editorconfig` files which use the ini format.

I'm not sure if my modifications to `LapceLanguage#from_path` are the best way to go about things, so if anyone has a better idea please let me know!

The only issue I'm aware of with this implementation is if a key doesn't have a value, the next line will be highlighted entirely as the previous line's value.


panekj:
This should be `filename`, not `dotfiles`

Insprill:
Currently, it only checks the array if the file is a dotfile.

https://github.com/Insprill/lapce/blob/d6d174e50e761fac468935e38a325cec0385144b/lapce-core/src/language.rs#L924

Should this behavior be changed to match any file then?

panekj:
> Currently, it only checks the array if the file is a dotfile.
> 
> [Insprill/lapce@`d6d174e`/lapce-core/src/language.rs#L924](https://github.com/Insprill/lapce/blob/d6d174e50e761fac468935e38a325cec0385144b/lapce-core/src/language.rs#L924)
> 
> Should this behavior be changed to match any file then?

I would like to NACK or keep on hold this PR until I finish rewriting language core which I started some time ago already

Insprill:
Would you prefer for me to remove the dotfile support and just leave ini highlighting, so dotfiles can be implemented at a later date?

panekj:
> Would you prefer for me to remove the dotfile support and just leave ini highlighting, so dotfiles can be implemented at a later date?

I checked that TS grammar and it doesn't seem to be original INI format but a different version of it based on AWS CLI config

Insprill:
I can't find any resources on an AWS version or how different it might be. I've looked around a bit and don't see any other ini tree sitters either. If it's too different to implement I'm fine with this being closed and taken up again when a better ts is available.

Issue #1169, opened 2022-09-12
Lapce takes control of terminal when opened via one


Ragbot:
## Is your feature request related to a problem? Please describe.

Opening Lapce from the terminal keeps it running in the terminal causing the terminal to be unusable

## Describe the solution you'd like

Maybe create a new process disconnected from the main process and exit off

## Describe alternatives you've considered

creating a shell file and make it open lapce but sadly the same result

## Additional context

none ;-;

panekj:
What version of Lapce are you using? This has been fixed already

Ragbot:
The latest according to the releases page 0.2

Ragbot:
I ttried the nigtly as well but the same issue persists

Ragbot:
![image](https://user-images.githubusercontent.com/64681480/189611468-cd4ad850-96c4-49af-86eb-889b56ba76fb.png)

Ragbot:
![image](https://user-images.githubusercontent.com/64681480/189612664-079d1b61-7e65-4c06-901d-9f17070f49b6.png)
Tried with a different terminal as well same issue

mkunkel:
I also have this issue with Lapce 0.2.0. This occurs when specifying a path, as well as when no path is specified. Also, `lapse . &` does not fix the issue.

Ragbot:
@mkunkel It somehow fixed it for me

panekj:
Fixed via #1646

PR #1646, opened 2022-11-05
feat: fork by default when started from terminal

Arjun31415:


panekj:
Fixes #1169

dzhou121:
Does this affect macOS/windows 'APP' startup?

panekj:
It shouldn't break anything

panekj:
> Can you maybe make it Linux only for now?

Fix in https://github.com/lapce/lapce/pull/1658

Issue #1588, opened 2022-10-23
 activate window num is wrong


humgal:
## Lapce Version
v0.2.1

## System information
macOS

## Describe the bug
in macOS ,we have close window button.if we click close button, window close, this is normal。but click lapce icon on dock ,the window will not activate again。and if you click icon on dock 4 times ,and press commond+Q to quite app, after this ,you open lapce again ,you will find 4 window activate.it's weird.

## Additional information
I guess this bug just because when we click lapce icon once ,the app has new a window in memory or some stored database, but no activate. and press command + Q app don't kill all windows . so this bug appears.

I-Info:
This looks like the `NewWindow` command is not handled after the `ApplicationShouldHandleReopen` event is handled.

https://github.com/lapce/lapce/blob/b843959ac10a23ef6ba9f75be65586e24685dac3/lapce-ui/src/app.rs#L314-L323

Citing the doc of [`druid::DelegateCtx::submit_command`](https://docs.rs/druid/latest/druid/struct.DelegateCtx.html#method.submit_command) :
> Submit a [Command](https://docs.rs/druid/latest/druid/struct.Command.html) to be run after this event is handled.
>
>Commands are run in the order they are submitted; all commands submitted during the handling of an event are executed before the [update()](https://docs.rs/druid/latest/druid/trait.Widget.html#tymethod.update) method is called.

Does this mean the command will never be handled until the widget is updated... 🤔

I-Info:
This rough patch works, but I don't know if it conforms to the spec:
```Rust
Event::ApplicationShouldHandleReopen(has_visible_windows) => {
    if !has_visible_windows {
        // ctx.submit_command(Command::new(
        //     LAPCE_UI_COMMAND,
        //     LapceUICommand::NewWindow(WindowId::next()),
        //     Target::Global,
        // ));
        let cmd = Command::new(
            LAPCE_UI_COMMAND,
            LapceUICommand::NewWindow(WindowId::next()),
            Target::Global,
        );
        self.command(ctx, Target::Global, &cmd, data, _env);
    }
    return None;
}
```

lixiao189:
> This looks like the `NewWindow` command is not handled after the `ApplicationShouldHandleReopen` event is handled.
> 
> https://github.com/lapce/lapce/blob/b843959ac10a23ef6ba9f75be65586e24685dac3/lapce-ui/src/app.rs#L314-L323
> 
[...]

PR #1651, opened 2022-11-06
Fix: new windows would never be created after closing all

I-Info:

Fixed: #1588

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 ### Bug Fixes
 
+- [#1588](https://github.com/lapce/lapce/issues/1588): Fixed an issue where new windows would never be created after closing all windows on macOS.
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 
 ### Bug Fixes
 
+- [#1588](https://github.com/lapce/lapce/issues/1588): Fixed an issue where new windows would never be created after closing all windows on macOS.
```
I-Info, line 12: Sorry for the previous mistake, it's more appropriate to put the link to the pull request here. 😢

I-Info: Is this serious? Should I fix it with a new PR?

I-Info (commented):
Found an inappropriate link in the `CHANGELOG.md`

Issue #1637, opened 2022-11-02
Python indentation always has width of 7


Dirleye:
## Lapce Version

0.2.1

## System information

EndeavourOS Linux

## Describe the bug
Indentations with tab in Python files always have a width of seven, no matter the value in settings.
C and C++ files didn't have the same issue and would follow the value in settings.

## Additional information

This happens with and without the Python plugin installed.

ghishadow:
looks like https://github.com/lapce/lapce/pull/1411/files#diff-d935b0f8af6b77e38d32b19067485d401a736d7af98037433583f51a85f4bcd8 has messed up many language indent values including python https://github.com/lapce/lapce/blob/master/lapce-core/src/language.rs#L624

PR #1666, opened 2022-11-08
fix: use correct python identation

Assistant:

Fixed: #1637


dzhou121:
Python indentation should be 4 spaces according to pep8

Assistant:
> Python indentation should be 4 spaces according to pep8

Changed it to 4 spaces and changed the comment specifying a tab for python to not be misleading

PR #1667, opened 2022-11-08
fix: add schemas, sort base colours, fix light theme

panekj:
* added schema definitions for settings and themes
* sorted base colours as they appear in lapce
* fixed light theme


panekj:
Schemas can include proper values for keys with `.` but I'm too lazy to do that right now since I would have to do it by hand

PR #1686, opened 2022-11-11
add editor and terminal zoom commands

### LINKED ISSUE #785: Add scaling factor support
On some Linux devices, DPI awareness can break. The default scaling could be too small or too large for a screen. Electron allows for a workaround by enabling zoom with `Ctrl + +` and `Ctrl + -`. Maybe there's some way to specify a scaling factor.

Sh4d1:
Signed-off-by: Patrik Cyvoct <patrik@ptrk.io>

Partially fixes #785, while it does not handle scaling, it allows an easy way to change the font's size

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- [#1686](https://github.com/lapce/lapce/pull/1686): Add zoom in/out keybindings for editor and terminal
+
 ### Features/Changes
 
 ### Bug Fixes
--- a/lapce-data/src/command.rs
+++ b/lapce-data/src/command.rs
@@ -457,6 +457,22 @@ pub enum LapceWorkbenchCommand {
     #[strum(message = "Toggle Inlay Hints")]
     ToggleInlayHints,
 
+    #[strum(serialize = "zoom_editor_in")]
+    #[strum(message = "Zoom In on Editor")]
+    ZoomInEditor,
+
+    #[strum(serialize = "zoom_editor_out")]
+    #[strum(message = "Zoom Out on Editor")]
+    ZoomOutEditor,
+
+    #[strum(serialize = "zoom_terminal_in")]
+    #[strum(message = "Zoom In on Terminal")]
+    ZoomInTerminal,
+
+    #[strum(serialize = "zoom_terminal_out")]
+    #[strum(message = "Zoom Out on Terminal")]
+    ZoomOutTerminal,
+
     #[strum(serialize = "restart_to_update")]
     RestartToUpdate,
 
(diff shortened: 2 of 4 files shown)


dzhou121:
Thanks for the PR. Would you like to improve it so that it's an actual zoom?

Sh4d1:
Not sure it is needed, it will get the same result no?

dzhou121:
When you zoom in/out, it should be everything in the UI, e.g. ui font sizes, button sizes, icon sizes, etc.

Sh4d1:
Okay I see, then it's two different functionnalities I guess. I personnaly would prefer to "zoom" only the editor, rather than the whole window. We could merge this one and modify the existing issue to scope only the full UI zoom ? (Also not sure where to begin for zooming the whole UI tbh :smile: )

panekj:
IMO, what you propose doesn't solve any issue and zoom/custom scaling should be implemented properly.
To do that in lapce, we should multiply all values by scale factor for all objects.

JustForFun88:
> IMO, what you propose doesn't solve any issue and zoom/custom scaling should be implemented properly. To do that in lapce, we should multiply all values by scale factor for all objects.

I am not a contributor (although I am thinking of becoming one), but I just want to express the opinion of a simple user. What you suggest is useless in my opinion. To be honest, it has always been annoying that in vscode, when using zoom, it scales not only the code, but everything, even the text in the explorer and the explorer itself (why?).

In my opinion, scaling is much more convenient when only the workspace can be scaled, and the interface remains the same size. This is done, for example, in LibreOffice.

panekj:
> > IMO, what you propose doesn't solve any issue and zoom/custom scaling should be implemented properly. To do that in lapce, we should multiply all values by scale factor for all objects.
> 
> To be honest, it has always been annoying that in vscode, when using zoom, it scales not only the code, but everything, even the text in the explorer and the explorer itself (why?).

Because it's exactly how Chromium behaves and VSCode is built with Chromium/Electron?

> In my opinion, scaling is much more convenient when only the workspace can be scaled, and the interface remains the same size.

This isn't scaling then, it's just font size change.

Sh4d1:
> This isn't scaling then, it's just font size change.

That's the goal of this PR though

panekj:
> > This isn't scaling then, it's just font size change.
> 
> That's the goal of this PR though

image

Sh4d1:
@panekj  not sure I get your screenshot

rochacbruno:
Works great for me, would like to have also mouse scroll support.

+1 for this PR

panekj:
> @panekj not sure I get your screenshot

I don't get your reply earlier either but it's fine since you are not the one that argues about the feature.
IMO, I'm fine with merging this if you rename it to `(Editor|Terminal) font size (increase|zoom in)`/`(Editor|Terminal) font size (decrease|zoom out)`

panekj:
Closing due to inactivity

PR #1690, opened 2022-11-11
add codelens and sticky headers for Dart

zarathir:
This PR adds codelens and sticky headers support for Dart.


panekj:
Could you include comment in changelog file?

zarathir:
Sure!

PR #1698, opened 2022-11-13
Implement lazy-loading ts grammars and runtime language configuration

panekj:
Signed-off-by: panekj <me@panekj.dev>


lapce-core/src/language.rs
```diff
@@ -3,930 +3,1810 @@ use std::{collections::HashSet, path::Path, str::FromStr};
...
-#[derive(Eq, PartialEq, Hash, Clone, Copy, Debug, Display, EnumString)]
+#[derive(Eq, PartialEq, Hash, Clone, Copy, Debug, PartialOrd, Ord, Default)]
+struct CommentProperties {
+    single_line_start: &'static str,
+    single_line_end: &'static str,
+
+    multi_line_start: &'static str,
+    multi_line_end: &'static str,
+    multi_line_prefix: &'static str,
+}
```
claytonrcarter, line 151: A lot of the languages in this file only have a `single_line_start`. Making this an enum w/ "simple"/"conplex" variants would cut down on that boilerplate:
```suggestion
enum CommentProperties {
    /// A simple comment style where all comments are
    /// single line comments beginning with the same token.
    Simple {
            single_line_start: &'static str,
    },
    
    /// A more complicated comment style where comments
    /// may not all begin and end with the same token.
    Complex {
        single_line_start: &'static str,
        single_line_end: &'static str,

        multi_line_start: &'static str,
        multi_line_end: &'static str,
        multi_line_prefix: &'static str,
    }
}
```

And then a lot of the languages can be simplified to eg `CommentProperties::Simple("//")` or `CommentProperties::Simple("#")`

lapce-core/Cargo.toml
```diff
@@ -22,6 +23,8 @@ lapce-rpc = { path = "../lapce-rpc" }
 
 tree-sitter = "0.20.7"
 
+# plaintext grammar, not optional
+tree-sitter-plaintext = { git = "https://github.com/panekj/tree-sitter-plaintext", rev = "38c37a098fea0eec26c6b0aeaa9feaec8bbe93fe" }
```
claytonrcarter, line 27: This seems to be working just fine, but can you clarify why this needs to be a tree-sitter grammar? Is it just to make interop with the rest of the "real" tree-sitter langs easier? Does this mean that all of the `find_tag()`, `find_tag_in_siblings()`, etc methods in `Syntax` will start working for "text" files, or will those still end up using `WordCursor`?

PR #1702, opened 2022-11-15
added queries for svelte

Arjun31415:


lapce-core/queries/svelte/highlights.scm
```diff
@@ -0,0 +1,109 @@
+; src: https://github.com/helix-editor/helix/blob/master/runtime/queries/svelte/highlights.scm
+; licence: https://github.com/helix-editor/helix/blob/master/LICENSE
+
```
panekj, line 3: could it be consistent with other queries and include spdx?

Arjun31415: added it

PR #1723, opened 2022-11-19
Display the keybinding of a command within the palette

### LINKED ISSUE #1683: Please, please in the control pallet after the description add the short-cut.
Hello,

To reinforce the learning of short-cuts so that eventually every programmer is highly productive because they only use the keyboard and short-cuts, 

please add the short-cut, where applicable to the end of each command pallet command, so that you only have to slow search once for it, then you will start using the short-cut.  

Best regards,
João

MinusGix:
  
This PR adds the keybinding for a command as the hint in the palette. Fixes #1683. This will make it easier for users to know the keybinding for a command they might be using often.  
image

It pays attention to what mode the user is in, so it won't recommend completely incorrect keybinds.

----
  
Could be improved with some better alignment of the keybindings?
I think this could also be improved by taking into account `when` the keybind can be used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Features/Changes
 - [#1720](https://github.com/lapce/lapce/pull/1720): Display signature/parameter information from LSP
+- [#1723](https://github.com/lapce/lapce/pull/1723): In the palette, display the keybind for a command adjacent to it
 
 ### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


dzhou121:
Would you mind changing it to be painted at the end of the line and in this style https://github.com/lapce/lapce/blob/master/lapce-data/src/keypress/mod.rs#L45

MinusGix:
Okay, I've made them right aligned and using the same general style.

Gerrit0:
Also fixes #1723

PR #1724, opened 2022-11-19
prevent volt loading on files and hidden file/folders

nheuillet:

### DIFF
--- a/lapce-proxy/src/plugin/wasi.rs
+++ b/lapce-proxy/src/plugin/wasi.rs
@@ -240,7 +240,6 @@ pub fn find_all_volts() -> Vec<VoltMetadata> {
                         return None;
                     }
                     let path = entry.path().join("volt.toml");
-                    println!("PATH: {:?}", &path);
                     load_volt(&path).ok()
                 })
                 .collect()
--- a/lapce-proxy/src/plugin/wasi.rs
+++ b/lapce-proxy/src/plugin/wasi.rs
@@ -233,7 +233,14 @@ pub fn find_all_volts() -> Vec<VoltMetadata> {
             d.read_dir().ok().map(|dir| {
                 dir.filter_map(|result| {
                     let entry = result.ok()?;
+                    if entry.file_type().ok()?.is_file() {
+                        return None;
+                    }
+                    if entry.file_name().to_str()?.starts_with(".") {
+                        return None;
+                    }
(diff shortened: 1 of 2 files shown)


lapce-proxy/src/plugin/wasi.rs
```diff
@@ -233,7 +233,14 @@ pub fn find_all_volts() -> Vec<VoltMetadata> {
...
                 dir.filter_map(|result| {
                     let entry = result.ok()?;
+                    if entry.file_type().ok()?.is_file() {
+                        return None;
+                    }
+                    if entry.file_name().to_str()?.starts_with(".") {
+                        return None;
+                    }
                     let path = entry.path().join("volt.toml");
+                    println!("PATH: {:?}", &path);
```
panekj, line 243:
```suggestion
```

panekj (changes requested):

lapce-proxy/src/plugin/wasi.rs
```diff
@@ -233,6 +233,12 @@ pub fn find_all_volts() -> Vec<VoltMetadata> {
             d.read_dir().ok().map(|dir| {
                 dir.filter_map(|result| {
                     let entry = result.ok()?;
+                    if entry.file_type().ok()?.is_file() {
+                        return None;
+                    }
+                    if entry.file_name().to_str()?.starts_with('.') {
+                        return None;
+                    }
```
panekj, line 241:
```suggestion
                    let entry = entry.metadata()?;
                    if !entry.is_dir() || entry.file_name().to_str()?.starts_with('.') {
                        return None;
                    }
```

panekj (changes requested):

lapce-proxy/src/plugin/wasi.rs
nheuillet: metadata does not contain file_name. I can remove the variable init if you prefer, seeing as it's only used on is_dir

panekj: you committed outdated suggestion

nheuillet: 🤦 Sorry about that

panekj: could probably use `if metadata.is_file()` to not break symlinks

panekj (approved):

Issue #1740, opened 2022-11-23
some Language server completion slow


xiaoxin-sky:
## Lapce Version

0.2.4 (Current HEAD)

## System information

Macos bigsur 11.5

## Describe the bug

Why send empty string when type a char in  [update_completion](https://github.com/lapce/lapce/blob/master/lapce-data/src/editor.rs#L801) function ?This lead to lsp client send two completion ,and this will lead to lsp server response slow.

## Additional information

I collect lsp client send and receive time, and write in log. I use typescript plugin Language server.
### before remove empty string completion request
![tsEmpty](https://user-images.githubusercontent.com/42181585/203462620-ab8b94b2-679d-4eda-87bb-a6374de81e70.jpg)
 
 ### after remove empty string 
![tsnoInput](https://user-images.githubusercontent.com/42181585/203462724-390e770d-313a-45a9-8efa-03b8e6571f4a.jpg)

> Remove the empty requests and the language server speed will increase ~700%

I think the empty completion request is no need. This lead to language response some global completion. 
I'd be happy to fix it

PR #1742, opened 2022-11-23
fix(editor): filter unnecessary  completion request

xiaoxin-sky:
see #1740 , I don't think empty string requests should be sent when the input has a value. Empty requests can cause language server performance problems.
I test typescript plugin and lapce-vue plugin(base tsserver).

### DIFF
--- a/lapce-data/src/completion.rs
+++ b/lapce-data/src/completion.rs
@@ -305,18 +305,17 @@ impl CompletionData {
         self.current_item().map(|item| item.item.label.as_str())
     }
 
-    #[allow(clippy::too_many_arguments)]
     pub fn request(
-        &self,
+        &mut self,
         proxy: Arc<LapceProxy>,
-        request_id: usize,
         path: PathBuf,
         input: String,
         position: Position,
     ) {
+        self.input_items.insert(input.clone(), im::Vector::new());
         proxy
             .proxy_rpc
-            .completion(request_id, path, input, position);
+            .completion(self.request_id, path, input, position);
     }
 
     pub fn cancel(&mut self) {
--- a/lapce-data/src/editor.rs
+++ b/lapce-data/src/editor.rs
@@ -770,7 +770,6 @@ impl LapceEditorBufferData {
                 let start_pos = self.doc.buffer().offset_to_position(start_offset);
                 completion.request(
                     self.proxy.clone(),
-                    completion.request_id,
                     self.doc.content().path().unwrap().into(),
                     "".to_string(),
                     start_pos,
@@ -781,7 +780,6 @@ impl LapceEditorBufferData {
                 let position = self.doc.buffer().offset_to_position(offset);
                 completion.request(
                     self.proxy.clone(),
-                    completion.request_id,
                     self.doc.content().path().unwrap().into(),
                     input,
                     position,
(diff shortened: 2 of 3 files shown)


dzhou121:
The reason we send empty string requests, is to kind of "cache" the completion results locally, so that for the same completion, we can quickly filter locally.

xiaoxin-sky:
> The reason we send empty string requests, is to kind of "cache" the completion results locally, so that for the same completion, we can quickly filter locally.

I think my commit  are not break the cache logic. the cache can work well.
image

dzhou121:
2 requests with one empty string, and the other one with the current input is by design. The empty string request gets all the possible completions which acts as cache. The request with input is for scenarios like auto import in rust-analyzer kicks in only with some input.

xiaoxin-sky:
yeah，I know your thoughts，But every type char  aways send the empty request，this aways lead to the language server calc the almost the same  result. and always slow down the complete.

Can we make an empty request when opening a new file instead of sending an empty request every time we type a char？

dzhou121:
We should do empty request once
https://github.com/lapce/lapce/blob/master/lapce-data/src/editor.rs#L769

One thing that can happen is the response is slow to come back, and we will keep sending the empty requests again and again until we got the response. The way to solve it is to insert a empty Vec for the empty request when we do it.

xiaoxin-sky:
> We should do empty request once https://github.com/lapce/lapce/blob/master/lapce-data/src/editor.rs#L769
> 

Please forgive me for not understanding. Dose [the 769 len](https://github.com/lapce/lapce/blob/master/lapce-data/src/editor.rs#L769) cached global response complete list being used again? Which situation will hit the empty response cache?

In my opinion, First hit the cache must in `CompletionStatus::Started` status, so In this case, the user should be in the state of continuously typing characters or deleting characters,right? 
1. And then if user in the state of continuously typing characters(such as: `abcd`), In my commit ,when user input the second char, it will send empty request, and also send input request('ab'), and then the empty string response will be cached in `input_items`, and the input string also be cached, and user continuously typing the 3rd char,  Because `completion.input_items.contains_key` need to match the full key, so the  `input_items`  can't match 3rd typing('abc'), so need send completion request.  So the response of the empty string in the process is almost hard to hit. 
 2. if user start delete typed chars, Because in the first case，the 'a'、'ab'、'abc' input  had been cached. so it alse not hit empty response.

In the above instructions, assume that the response speed is normal. So It's hard for me to understand that global caching is of great use here. 

> One thing that can happen is the response is slow to come back, and we will keep sending the empty requests again and again until we got the response. 

Of course, the response results did not return in time is another need to solve a problem.

dzhou121:
> In the above instructions, assume that the response speed is normal. So It's hard for me to understand that global caching is of great use here.

The global cache is to display completion results before the response for new input comes back.

xiaoxin-sky:
> The global cache is to display completion results before the response for new input comes back.
>
oh!,I know your design idea. Can I understand that the global is always obtained before each input, and a global list of completion will be displayed before the real content of the input response is returned?
If I understand correctly, then I don't think this is a good idea. On the contrary, the effect is somewhat bad. Because the language need response 2 request，and this will black the 2td request.

dzhou121:
I think the real issue is that we keep sending the requests until the response is back. I'll fix that and you can test if it solves the problem.

xiaoxin-sky:
Ok, I continue to follow your updates. 🚀😄

dzhou121:
Can you please see if https://github.com/lapce/lapce/pull/1746 fixes the issue?

xiaoxin-sky:
> Can you please see if #1746 fixes the issue?

I can see the continuous input and Competion pop-up window will no longer flash.nice. It's kind of a repair.
Although the first empty request can not be removed for design reasons. But better than before

xiaoxin-sky:
I close the pr.

PR #1759, opened 2022-11-29
Update C tree-sitter

Arjun31415:


dzhou121:
Can you please fix the conflict on changelog file

Issue #1762, opened 2022-11-29
Lapce panicks when it tries to load syntax highlighting info for .sh files


TudbuT:
## Lapce Version

![image](https://user-images.githubusercontent.com/48156391/204592137-121917ba-2995-4b08-aa58-291d7f13db94.png) 
(6e4a472)

## System information

Arch Linux

## Describe the bug
Lapce panicks when it tries to load syntax highlighting for .sh files. 
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: QueryError { row: 52, column: 3, offset: 702, message: "expansion_flags", kind: NodeType }': lapce-core/src/language.rs:974

## Additional information

I honestly have no idea why this is happeningm, but the original Err happens in the constructor of HighlightConfiguration or Query

simoniz0r:
I am getting a similar error with version 0.4.0 running on openSUSE Tumbleweed using the pre-compiled tar from the releases page:

```
 syretia  ~  lapce -V
Lapce v0.4.0
 syretia  ~  lapce -w
2024-04-30T05:50:30.722961Z ERROR lapce_core::language: Encountered QueryError { row: 48, column: 3, offset: 520, message: "expansion_flags", kind: NodeType } while trying to construct HighlightConfiguration for Bash
```

Additionally, syntax highlighting doesn't seem to do anything for shell scripts; they display the same as a plain text file would.

Sorry if I should have opened a new issue for this.

PR #1764, opened 2022-11-30
(band-aid) fix opening some file types crashing lapce

TudbuT:

THIS IS NOT A FULL FIX! It only fixes the *crash*, but not the underlying issue, as that is somewhere in the dependencies and I'm unable to find out how to fix it. The error will still be logged, but will not crash lapce.

I might look into the issue further, but suggest adding this band-aid so we at least don't get a crash.

Example of a logged error: `[2022-11-30][08:31:11][lapce_core::language][ERROR] Encountered QueryError { row: 52, column: 3, offset: 702, message: "expansion_flags", kind: NodeType } while trying to construct HighlightConfiguration for bash` - all information from the panic message is still present after this fix.

Shell scripts, and potentially other languages, make lapce panic when opened. This is explained further in the issue this PR closes.

Closes: #1762
Opens: Something like "Lapce does not highlight .sh and .bash files"


dzhou121:
Can you please have a look to fix the grammar/query instead?

TudbuT:
> Can you please have a look to fix the grammar/query instead?
Fixed in #1769

PR #1770, opened 2022-11-30
add terminal tab

dzhou121:


TudbuT:
Great to see this feature coming, split terminals were kind of annoying whenever i needed more than 2

PR #1769, opened 2022-11-30
(properly) fix opening bash files crashing lapce

TudbuT:
- [x] Added an entry to `CHANGELOG.md` if this change could be valuable to users

Better version of #1764

Closes: #1762

PR #1772, opened 2022-12-01
Add error dialog for invalid syntax highlighting schemes

TudbuT:

This is a follow-up to the syntax highlighting issues #1762 #1764 #1768 #1769 #1771.

This PR adds proper and visible failure handling of any future encounter with invalid syntax highlighting .scm files. When such a file is invalid, as soon as an associated file is opened, there will be a dialog saying 
image
The error will also be logged.

All the info from the panic is there, but without the confusion of "Why did lapce just crash???".

### DIFF
--- a/lapce-data/src/document.rs
+++ b/lapce-data/src/document.rs
@@ -428,7 +428,7 @@ impl Document {
                     "Syntax Highlighting failed".to_owned(),
                     ShowMessageParams {
                         typ: MessageType::ERROR,
-                        message: format!("An error occured trying to load syntax highlighting info: {x}. Please report this."),
+                        message: format!("An error occurred trying to load syntax highlighting info: {x}."),
                     },
                 );
                 None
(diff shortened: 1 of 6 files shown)


lapce-data/src/document.rs
```diff
@@ -414,6 +416,26 @@ impl Document {
...
+    ) -> Option<Syntax> {
+        match syntax {
+            Ok(x) => Some(x),
+            Err(HighlightIssue::NotAvailable) => None,
+            Err(HighlightIssue::Error(x)) => {
+                proxy.core_rpc.show_message(
+                    "Syntax Highlighting failed".to_owned(),
+                    ShowMessageParams {
+                        typ: MessageType::ERROR,
+                        message: format!("An error occured trying to load syntax highlighting info: {x}. Please report this."),
```
panekj, line 431:
```suggestion
                        message: format!("An error occurred trying to load syntax highlighting info: {x}."),
```

PR #1776, opened 2022-12-02
fix: hide language/text details on non-editor tabs

panekj:

### DIFF
--- a/lapce-ui/src/status.rs
+++ b/lapce-ui/src/status.rs
@@ -8,7 +8,7 @@ use lapce_core::mode::Mode;
 use lapce_data::{
     command::{CommandKind, LapceCommand, LapceWorkbenchCommand, LAPCE_COMMAND},
     config::{LapceConfig, LapceIcons, LapceTheme},
-    data::{EditorTabChild, LapceTabData},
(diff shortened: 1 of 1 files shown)


lapce-ui/src/status.rs
```diff
@@ -482,6 +482,23 @@ impl Widget<LapceTabData> for LapceStatus {
...
+            // Do not draw file language and cursor/text details for non-file tabs
+            if let Some(editor_tab_id) = editor.tab_id {
+                if let Some(editor_tab) =
+                    data.main_split.editor_tabs.get(&editor_tab_id)
+                {
+                    if let Some(child) = editor_tab.active_child() {
+                        match child {
+                            EditorTabChild::Settings { .. }
+                            | EditorTabChild::Plugin { .. } => {
+                                return;
```
dzhou121, line 494: This could cause a surprise if we'll have something to paint after this.

dzhou121: I meant we should probably extract this paint in a new fn and return there.

panekj: @dzhou121 does that look better?

panekj: Might as well split document language name and details into own functions, but I'm lazy :P

PR #1778, opened 2022-12-04
format keybinding name on settings to Capital case.

ppshobi:
With this PR, the keybindings which doesn't have a message/description will be `Capital case`d, instead of showing the `kebab_cased` name.

This is my first PR, let me know what you guys think
**Before**
Before

**After**
After

### DIFF
--- a/lapce-ui/src/keymap.rs
+++ b/lapce-ui/src/keymap.rs
@@ -274,15 +274,17 @@ impl Widget<LapceTabData> for LapceKeymap {
                         ));
                         let text_layout = ctx
                             .text()
-                            .new_text_layout(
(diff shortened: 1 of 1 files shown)


panekj:
This should be done via strum

ppshobi:
> This should be done via strum

This pr doesn't really affect the strum message functionality. If there is a message it would be displayed

panekj:
> This pr doesn't really affect the strum message functionality. If there is a message it would be displayed

That doesn't change the fact it should be done via strum

dzhou121:
I think this is a good addition when command description isn't there.

ppshobi:
Thanks

Issue #1779, opened 2022-12-04
don't ignore '.gitignore' in "go to file" if no '.git' folder is present


senneh:
## Is your feature request related to a problem? Please describe.

When there is no `.git` folder at the root of your workspace, `.gitignore` will be ignored by `ctrl` `p`.
This happens when you haven't initialized your repository yet.

## Describe the solution you'd like

Make it so `.gitignore` is always respected even if the workspace isn't a .git repository (yet). 

## Additional context

Sure, regular `.ignore` files are still respected. But the magical behavior of "sometimes .gitignore is respected and sometimes it's not" is both confusing and undocumented. Currently we just do whatever ripgrep does, which might not be desirable.

I think it's necessary to consider the sort of behavior we want for "go to file", seeing as there are more quirks like #1605. 

For good consistent standard behavior I think it should:
* Always respect `.gitignore` at the root of your workspace even if it isn't in a git repository.
* Show hidden files, with the exception of `.git` which should always be ignored.
* Do not search parent directories for `.gitignore` files. If ripgrep is in a subfolder of a git repo it will look in a parent directory for `.gitignore` files. However, do we really want Lapce to look for .gitignore files that are outside the workspace? 
* `.ignore` overwrites `.gitignore` (current behavior)
* Respect `.git/info/exclude` and git's global exclude (current behavior)

A walker with this behavior can be configured like 
```rust
let ignore_git = OverrideBuilder::new(&path)
    .add("!.git/")
    .expect("error handling ignored for brevity")
    .build()
    .expect("error handling ignored for brevity");

let walker = WalkBuilder::new(&path)
    .hidden(false)
    .parents(false)
    .require_git(false)
    .overrides(ignore_git)
    .build();

```

Issue #1605, opened 2022-10-25
Don't skip dotfiles in "Go to File"


Pascal-So:
## Is your feature request related to a problem? Please describe.

When I want to edit a hidden file, e.g. `.gitignore`, I can't jump to it with <kbd>Ctrl</kbd><kbd>P</kbd> because no hidden files are listed in this "Go to File" search.

## Describe the solution you'd like

I like the way that VSCode does it: hidden files are shown, files listed inside the `.gitignore` are hidden, contents of the `.git` directory are hidden.

## Describe alternatives you've considered

I tried to quickly implement it myself by just changing [dispatch.rs:581](https://github.com/lapce/lapce/blob/de9655543c17069f7ebe4963b34adf439d4fd637/lapce-proxy/src/dispatch.rs#L581) as follows:

```rust
for path in ignore::WalkBuilder::new(workspace).hidden(false).build().flatten() {
```

but the problem here is that the entire `.git` directory is then listed as well. As far as I can tell there are two possible solutions to this: either we do the filtering ourselves to remove `.git` contents and other unwanted results, or we use an additional ignore file through [`ignore::WalkBuilder::add_ignore`](https://docs.rs/ignore/0.4.18/ignore/struct.WalkBuilder.html#method.add_ignore), but then we would have to store that file somewhere.

Skipping `.git` directly in the `ignore` crate would be much more performant.

Pascal-So:
Just tested this again on version 0.2.5 and it appears to be fixed by #1780, thanks @senneh!

PR #1780, opened 2022-12-04
Configure dir walker for "go to file"

senneh:
see issues #1779 and #1605


dzhou121:
Thanks!

PR #1782, opened 2022-12-05
WIP: History based tab switching

philburr:
This patch adds a setting to change the behavior of the NextEditorTab and PreviousEditorTab to switch to the most recently used tab/least recently used tab.

This is my first PR for this project and so am anxious to get some feedback.  I am (admittedly biased) used to using vscode where ctrl-tab and ctrl-shift-tab switch to most recently used/least recently used tabs as opposed to next/previous in tab index order.

This PR adds a setting to alternate between these behaviors.  However after doing this, I'm thinking a more appropriate solution would be to keep  `NextEditorTab`/`PreviousEditorTab` with existing behavior and add new LapceUICommand commands such as `SwitchMostRecentlyUsedTab`/`SwitchLeastRecentlyUsedTab` and instead of using a editor setting to change behavior, rely on user changing key bindings.

### DIFF
--- a/defaults/settings.toml
+++ b/defaults/settings.toml
@@ -40,6 +40,7 @@ multicursor-whole-words = true
 render-whitespace = "none"
 show-indent-guide = true
 atomic-soft-tabs = false
+historic-next-tab = false
 
 [terminal]
 font-family = ""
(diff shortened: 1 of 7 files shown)


dzhou121:
Thanks for adding this feature! Would you like to change it as you suggested as in additional commands to do it.

PR #1786, opened 2022-12-06
add file menu

ppshobi:


ppshobi:
@panekj any idea why this is failing windows/ubuntu, while `cargo fmt --all` is the one who does this change?

panekj:
It's not `cargo` fault, you didn't add OS gate for import

PR #1791, opened 2022-12-08
highlight scope lines

zarathir:
This PR adds highlighting for scope lines.

image

Functionality is similar to vscode:
 - if the closing bracket is not the first character on the line it paints around it
image
 - and also around indents
image

### Minor changes to bracket matching
This PR also fixes a minor inconvenience which returned the offset of the closing bracket first, if the cursor was on the closing bracket. I think it makes more sense to always return `(opening_bracket, closing_bracket)`, as the other way around may not be what people expect.


dzhou121:
Thanks for the feature! I'm wondering if it should default to off?

zarathir:
I didn't really think about that and just copied it from the bracket highlighting, to be honest. :)
Should I switch it off? I don't really have a preference for these kind of things.

dzhou121:
Yeah I think so. This is the first time I saw this to be honest but rather cool!

zarathir:
Sure thing, will do.
I don't know about other editors, but vscode has this. It's also possible to colorize the brackets and scope lines there, which I think would also be a nice addition to Lapce, but I'll have to get a little more proficient with rust to implement this nicely. Maybe someday :)

PR #1796, opened 2022-12-10
Removing unnecessary allocations and more

JustForFun88:
This pull request does:
- removes a lot of unnecessary memory allocations. For example, previously, multiple allocations of vectors were performed in a loop to store a single character (function `lapce_core::syntax::Syntax::find_matching_pair`);
- removes many indirect function calls. For example calling `to_string()` on a `&' static str`, using `"".to_string()` instead of `String::new()`.

In addition, a complete rewrite of the `Snippet::extract_text` function was performed while preserving the logic of its work. This was done, firstly, to remove the constant copying of immutable vectors in a loop. Secondly, to exclude a possible crash of the program, since accessing an element by index for a string is not a good idea.

### DIFF
--- a/lapce-core/src/syntax/mod.rs
+++ b/lapce-core/src/syntax/mod.rs
@@ -604,11 +604,11 @@ impl Syntax {
                 match highlight {
                     HighlightEvent::Source { start, end } => {
                         if let Some(hl) = current_hl {
-                            if let Some(hl) = SCOPES.get(hl.0) {
+                            if let Some(&hl) = SCOPES.get(hl.0) {
                                 highlights.add_span(
                                     Interval::new(start, end),
                                     Style {
-                                        fg_color: Some(hl.to_string()),
+                                        fg_color: Some(hl.to_owned()),
                                     },
                                 );
                             }
@@ -687,10 +687,16 @@ impl Syntax {
         let node = tree
             .root_node()
             .descendant_for_byte_range(offset, offset + 1)?;
-        let mut chars = node.kind().chars();
-        let char = chars.next()?;
+        let char = node.kind().chars().next()?;
         let char = matching_char(char)?;
-        let tag = &char.to_string();
+
+        // Note: Avoid using `encode_utf8` since it introduce unnecessary branches
+        //
+        // SAFETY: We know for sure that `char` will fit into u8 because the
+        // `match_char` function returns only brackets.
+        let char_buffer = [char as u8];
+        // SAFETY: `char` was valid UTF-8.
+        let tag = unsafe { core::str::from_utf8_unchecked(&char_buffer) };
 
         if let Some(offset) = self.find_tag_in_siblings(node, true, tag) {
             return Some(offset);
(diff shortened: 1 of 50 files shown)


panekj:
what's wrong with `String::new()`?

lapce-core/src/syntax/mod.rs
```diff
@@ -604,11 +604,11 @@ impl Syntax {
                 match highlight {
                     HighlightEvent::Source { start, end } => {
                         if let Some(hl) = current_hl {
-                            if let Some(hl) = SCOPES.get(hl.0) {
+                            if let Some(&hl) = SCOPES.get(hl.0) {
                                 highlights.add_span(
                                     Interval::new(start, end),
                                     Style {
-                                        fg_color: Some(hl.to_string()),
+                                        fg_color: Some(hl.to_owned()),
```
panekj, line 611: I would prefer if we leave all `to_string()`, as all main contributors write code that way

JustForFun88: Okay, I'll revert to the old implementation.

Actually, at first I wanted to use `to_owned` and `clone` everywhere, but I realized that this would limit flexibility. Apparently in this place I replaced the functions by mistake.

But I still think the call to `to_string` on `&'static str` should be removed.

JustForFun88: Done

panekj: I meant all `to_owned()` -> `to_string()`

JustForFun88: Done :-)?

lapce-rpc/Cargo.toml
```diff
@@ -10,7 +10,7 @@ edition = "2021"
 log = "0.4.17"
 parking_lot = "0.11.2"
 anyhow = "1.0.34"
-indexmap = "1"
+indexmap = { version = "1.9.2", features = ["serde"] }
```
JustForFun88, line 13: Added `features = ["serde"]` because it doesn't compile for testing otherwise (#[derive(Serialize, Deserialize)] doesn't work on IndexMap).

JustForFun88:
@panekj  Made requested fixes and improved some of my solutions. I think this pull request is ready for another review.

panekj:
Where is the `CharBuffer` from?

JustForFun88:
> Where is the `CharBuffer` from (and why)?

I wrote it myself as a helper structure to store a one-character `&str` on the stack instead of on the heap. I thought it would be a useful structure both now and later.

And well, since this is a structure for storing a "string slice", then it must be done correctly so that it can be compared with other string structures, converted from one structure to another and vice versa, it must have documentation and tests.

The name may not be very good, but I couldn't think of a better one.

panekj:
Is there anything left to do?

JustForFun88:
> Is there anything left to do?

Yes, I need to clean up the PR by removing things that have already merged. Can't do it right now as I'm busy developing generic traits for containers/collections.

PR #1805, opened 2022-12-12
Fix some LSP bugs causing code actions to not show up correctly

TudbuT:

This will fix the following LSP-related issues: 
- Java LS does not show "Surround with try/catch" and similar issue-related actions
- All language servers don't always update code actions when the document was changed
- Rust LS sometimes highlights a line with a fix suggestion, but is unable to actually provide it in code actions

This PR updates the capabilities and code action requests to fully match LSP spec.

### DIFF
--- a/lapce-data/src/editor.rs
+++ b/lapce-data/src/editor.rs
@@ -303,7 +303,14 @@ impl LapceEditorBufferData {
             self.proxy.proxy_rpc.get_code_actions(
                 path.clone(),
                 position,
-                diagnostics.into_iter().map(|x| x.diagnostic).collect(),
+                diagnostics
+                    .into_iter()
+                    .map(|x| x.diagnostic)
+                    .filter(|x| {
+                        x.range.start.line <= position.line
+                            && x.range.end.line >= position.line
+                    })
+                    .collect(),
                 move |result| {
                     if let Ok(ProxyResponse::GetCodeActionsResponse {
                         plugin_id,
--- a/lapce-data/src/editor.rs
+++ b/lapce-data/src/editor.rs
@@ -289,13 +289,21 @@ impl LapceEditorBufferData {
             let offset = self.editor.cursor.offset();
             let prev_offset = self.doc.buffer().prev_code_boundary(offset);
             // REMOVED(if) this prevents some code actions from showing up correctly, and there is almost no performance penalty for this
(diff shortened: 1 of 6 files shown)


TudbuT:
Wait what, why'd clippy fail here but not on my local PC?

lapce-proxy/src/plugin/mod.rs
```diff
@@ -488,7 +489,16 @@ impl PluginCatalogRpcHandler {
                 start: position,
                 end: position,
             },
-            context: CodeActionContext::default(),
+            context: CodeActionContext {
+                diagnostics: diagnostics
+                    .into_iter()
+                    .filter(|x| {
```
dzhou121, line 495: Maybe it's better to filter the diagnostics in the code actions request.

TudbuT: Done

lapce-data/src/editor.rs
```diff
@@ -288,34 +288,50 @@ impl LapceEditorBufferData {
...
-                                    resp,
-                                },
-                                Target::Auto,
-                            );
-                        }
-                    },
-                );
-            }
+            // REMOVED(if) this prevents some code actions from showing up correctly, and there is almost no performance penalty for this
+            //if self.doc.code_actions.get(&prev_offset).is_none() {
```
dzhou121, line 292: Does this make it not work properly?

TudbuT: The if-statement causes some delayed code actions to be left out, causing some quick fixes to not appear. REMOVED(if) is just the reason why it was removed and what was removed, in this case an if statement.

dzhou121: Is there any way I can test the delayed code actions bit?

TudbuT: Take this rust code:

```rs
fn main() {
    let x = 0;
    x = 1;
    println!(x);
}
```

There are two errors here. Open quick fixes on the variable name, there will be no fix. Now you fix the println one, which has a higher internal priority in RA. Save the file. The bug happens when you open quick fixes on the variable name again, even after you have now fixed the println one and it shows a new error.

TudbuT: The same goes for code actions on already-modified code.

PR #1807, opened 2022-12-12
Code action change

dzhou121:

### DIFF
--- a/lapce-data/src/command.rs
+++ b/lapce-data/src/command.rs
@@ -596,6 +596,11 @@ pub enum LapceUICommand {
         offset: usize,
         resp: CodeActionResponse,
     },
+    CodeActionsError {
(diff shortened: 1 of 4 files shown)


dzhou121:
@TudbuT Would you mind testing this which improves the code actions cache.

TudbuT:
The first of these commits is redundant, as the issue is not there being no code action, its there being wrong code actions. I think we only need to clear the cache on update to be able to enable it again.

dzhou121:
> Oh, that makes sense. We'd also have to clear whenever we get a new EditorDiagnostic, as some LSPs rely on those and wont provide quick fixes otherwise.

Done in latest commit

TudbuT:
Looks good to me, tho as said in the review, that check is a little bit cluttered in my opinion :P

PR #1808, opened 2022-12-13
Use more standardized java indentation

TudbuT:

Java's indentation is almost always 4 spaces, so I think using 4 spaces as the default in Lapce as well makes a lot of sense.


TudbuT:
Check failed due to insufficient disk space(?), not anything I can fix: = note: LINK : fatal error LNK1201: error writing to program database 'D:\a\lapce\lapce\target\debug\build\crc32fast-8dfdeabc22498dee\build_script_build-8dfdeabc22498dee.pdb'; check for insufficient disk space, invalid path, or insufficient privilege

PR #1809, opened 2022-12-14
Dap support

dzhou121:


panekj:
Could you split the `dap_types` into separate crate?

PR #1810, opened 2022-12-14
remove mold requirement for fedora package

Titaniumtown:
Hello! I'm the maintainer of the lapce builds for Fedora.

This PR removes the requirement of mold for compilation and will default to the system compiler. This only applies to builds for packaging on Fedora and related distros. This would allow the package to be build on powerpc as the mold version in the repos don't support powerpc.

### DIFF
--- a/lapce.spec
+++ b/lapce.spec
@@ -8,7 +8,7 @@ URL:            https://github.com/lapce/lapce
 VCS:            {{{ git_dir_vcs }}}
 Source:        	{{{ git_dir_pack }}}
 
-BuildRequires:  cargo perl-FindBin cairo-devel cairo-gobject-devel atk-devel gdk-pixbuf2-devel pango-devel gtk3-devel perl-lib perl-File-Compare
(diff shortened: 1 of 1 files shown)


panekj:
I would prefer to remove the spec completely, it should be maintained in Fedora rather than here

Titaniumtown:
well I simply just run the copr repo. I have no control over that.

panekj:
> well I simply just run the copr repo. I have no control over that. This spec is used for the `-git` variation. For all of the release builds, I have a separate spec for those.

couldn't the spec be stored along release version?

Titaniumtown:
> > well I simply just run the copr repo. I have no control over that. This spec is used for the `-git` variation. For all of the release builds, I have a separate spec for those.
> 
> couldn't the spec be stored along release version?

I guess-? But it's way harder to do `-git` builds with the spec file not in the actual repo being targeted. I tried to do that before submitting the original PR here to add a spec file. From what I remember I couldn't get it working properly. If someone can tell me how to do it, I will.

panekj:
You can use custom build type on COPR, anyway it should not be our responsibility to host the spec file.

Titaniumtown:
ok then

Titaniumtown:
seems there isn't even support for powerpc with cranelift-codegen, so this PR was useless anyways.

PR #1820, opened 2022-12-17
fix: add option to change remote svg icon colour

panekj:


panekj:
No plugin provides settings:
image

Plugins with settings exist:
image

PR #1826, opened 2022-12-17
Support hiding open editors

### LINKED ISSUE #1114: Minimize `open editors` at the side bar as a single line

We already have opened files on the top bar, the side bar one seems unnecessary most of times

We can let `Open Editors` which at the sidebar can be minimize as a single line

## Additional context

The `Open Editors` takes me 25% space in the sidebar
Screen Shot 2022-09-09 at 3 45 50 PM

Gerrit0:

This works, but requires a restart for changes to take effect, which isn't great. I think I'm going about this the wrong way, since it seems that once the panel is created, it's sections aren't expected to change. Is there a better way to do this?

Resolves #1114

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features/Changes
 
+- [#1114](https://github.com/lapce/lapce/issues/1114): Add option to hide open editors
+
 ### Bug Fixes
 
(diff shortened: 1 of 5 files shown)


panekj:
there is https://github.com/lapce/lapce/pull/1574 already

PR #1829, opened 2022-12-18
Rewrite some parts of `LapceSettingsPanel::update_plugins`

JustForFun88:
This pull request does the following:
- Changes the structure of `LapceSettingsPanel::children` to use `hashbrown::HashMap`. This was done because the `std::collections::HashMap::drain_filter` method is not yet stable. Also, this should make things a lot faster with `LapceSettingsPanel::children`, since `hashbrown::HashMap` uses [`aHash`](https://github.com/tkaitchuck/ahash) by default, which is much faster than `SipHash 1-3` hash function [`used`](https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#3142) in the default library (https://github.com/tkaitchuck/aHash/blob/master/FAQ.md#how-is-ahash-so-fast).

- Removes unnecessary creation of an intermediate vector `Vec<LapceSettingsKind>` (for which we have to go through all the keys and clone them).

- Corrects the search for a key in `HashMap` and `IndexMap`, since previously the search for a key in a hash table was performed by a **linear search through an iterator**:
  ``` rust
  data.plugin.installed.keys().contains(&volt_id)
  //...
  self.children.keys().contains(&kind)
  ```

### DIFF
--- a/lapce-ui/src/settings.rs
+++ b/lapce-ui/src/settings.rs
@@ -139,24 +139,23 @@ impl LapceSettingsPanel {
     }
 
     fn update_plugins(&mut self, ctx: &mut EventCtx, data: &LapceTabData) {
-        let drained = self.children.drain_filter(|kind, _| {
+        let mut children_changed = false;
+        self.children.retain(|kind, _| {
             if let LapceSettingsKind::Plugin(volt_id) = &kind {
-                !data.plugin.installed.contains_key(volt_id)
-            } else {
-                false
+                if !data.plugin.installed.contains_key(volt_id) {
+                    if self.active == *kind {
+                        self.active = LapceSettingsKind::Core;
+                        self.switcher
+                            .widget_mut()
+                            .child_mut()
+                            .set_active(self.active.clone(), data);
+                    }
+                    children_changed = true;
+                    return false;
+                }
             }
+            true
         });
-        for (kind, widget_pod) in drained {
-            drop(widget_pod);
-            ctx.children_changed();
-            if self.active == kind {
-                self.active = LapceSettingsKind::Core;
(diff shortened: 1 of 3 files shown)


lapce-ui/src/settings.rs
```diff
@@ -138,27 +139,28 @@ impl LapceSettingsPanel {
...
-                            .set_active(self.active.clone(), data);
-                    }
-                }
+                !data.plugin.installed.contains_key(volt_id)
+            } else {
+                false
+            }
+        });
+        for (kind, widget_pod) in drained {
+            drop(widget_pod);
```
hbina, line 150: Is this drop necessary? Wouldn't it get dropped as drained is dropped anyways?

dzhou121: The actual calling sequence of element added/removed and ```ctx.children_changed()``` doesn't really matter. 

```ctx.children_changed()``` only sets a flag and used for the next paint cycle.

JustForFun88: Yes I understand. I'm generally going to throw out the multiple call to this function in a loop and just have some extra flag or something like that. I'll think about it after work.

JustForFun88: Switched to using `retain` instead of `drain_filter` and `drop`

JustForFun88:
@dzhou121, @hbina ready for another review

JustForFun88:
Can someone rerun `windows-latest` check?

JustForFun88:
This is ridiculous 😅. Now the check fails because `failed to load source for dependency 'tree-sitter-julia'`

panekj:
You have to rebase on master

PR #1838, opened 2022-12-18
Don't use the active editor's view id for opening file diff

MinusGix:

Currently, clicking on a diff in the source control sidebar will cause it to replace the current editor (even if it is unsaved and/or a scratch file). This PR changes the code to not replace the currently opened file.  
  
(Question: Was the behavior of jump_to or this specific command changed some time ago? I don't remember opening the diff doing this a long while ago, but it started doing it sometime in the last couple months I believe)


panekj:
> (Question: Was the behavior of jump_to or this specific command changed some time ago? I don't remember opening the diff doing this a long while ago, but it started doing it sometime in the last couple months I believe)

does git blame say anything?

PR #1855, opened 2022-12-23
feat(editor): Support auto-close for `` and <>

claytonrcarter:

This updates the bracket/quote auto-pairing behavior to include backticks and angle brackets. I find backtick matching really handy when writing markdown, and backtick matching is really nice when writing code with generics in all sorts of languages.
I expanded a test to cover this change; aside from that, I've been using this locally for several days and it has been working as expected. Thanks!


claytonrcarter:
Test failure appears to be a false negative due to network timeouts. I think? I don't believe that I have access to rerun.

lapce-core/src/syntax/util.rs
```diff
@@ -48,6 +50,8 @@ pub fn matching_char(c: char) -> Option<char> {
         ')' => '(',
         '[' => ']',
         ']' => '[',
+        '<' => '>',
+        '>' => '<',
```
JustForFun88, line 54: Shouldn't the `matching_bracket_general` function defined below also be updated?

claytonrcarter: 🤨 How did I miss that? Fixed. Thank you!

JustForFun88:
I think this feature should be improved.
**It creates angle brackets when not needed:**


**Highlights "paired" characters "<" and ">"**:

claytonrcarter:
> I think this feature should be improved.

You're right; angle bracket pairing was half-baked. I've removed it from this PR. I checked several other editors (Doom Emacs, Helix, Typora) and none actually supported angle bracket pairing, but they all supported backtick pairing (which I've left in place in this PR)

Thank you!

panekj (changes requested):
auto-close for grave accent shouldn't be merged since it's used as an escape symbol in certain languages and it would make an annoying behaviour like angle brackets

claytonrcarter:
Thanks for the feedback, @panekj. Which languages use backtick/grave as an escape (I admit that I'm drawing a blank)?

I'm going back to the drawing board to make this better and to make it more configurable by language. I'd like to make sure that I handle them correctly while also allowing pairing in appropriate langs (eg Markdown, JS, sh/bash, php, etc). Thanks!

panekj:
`powershell` uses grave accent as a line continuation character (equivalent to `sh` `\`)
```powershell
Remove-Item `
  -Recurse `
  -Force `
  .
```

panekj:
> I'm going back to the drawing board to make this better and to make it more configurable by language.

That's currently not possible

claytonrcarter:
> `powershell` uses grave accent as a line continuation/escape character (equivalent to `sh \`)

Cool, thank you. I don't see Powershell as a supported language in Lapce, though. I'll try to think about how that should work, because I assume there will be lots of "not officially supported" languages where current behaviors could also be annoying. Lisp comes to mind, because lists of symbols can be made via "single quoting" like `'(a b c)`. These are fairly common in lisp.

> That's currently not possible

I'm aware. I'm working on changing that.

panekj:
> Lisp comes to mind, because lists of symbols can be made via "single quoting" like `'(a b c)`.

To add to that, we also have such rust syntax, where lifetimes are explicitly specified
```rust
fn func_name<'a, 'b>(arg1: 'a type1, arg2: 'b type2) -> (out1::<turbofish1>, out2::<turbofish2>) {
```

PR #1863, opened 2022-12-26
Optimization of `lapce_ui::terminal::TerminalPanel::event`

JustForFun88:


lapce-ui/src/terminal.rs
```diff
@@ -139,30 +140,53 @@ impl Widget<LapceTabData> for TerminalPanel {
...
+            // SAFETY: We have already checked that empty_tabs is not empty
+            let id_to_remove = unsafe { *empty_tabs.get_unchecked(0) };
+            let removed_len = empty_tabs.len();
+
             let terminal = Arc::make_mut(&mut data.terminal);
+            let tabs_order = Arc::make_mut(&mut terminal.tabs_order);
             for tab_id in empty_tabs {
                 self.tabs.remove(&tab_id);
                 terminal.tabs.remove(&tab_id);
-                terminal.tabs_order = Arc::new(
```
JustForFun88, line 154: This version is not able to immediately allocate a vector of the required size, due to the use of a filter

lapce-ui/src/terminal.rs
```diff
@@ -139,30 +140,53 @@ impl Widget<LapceTabData> for TerminalPanel {
...
+            // Therefore, we first store the first `WidgetId` to be removed in the `id_to_remove`
+            // variable, and in the loop itself, instead of deleting, we will equate all the values
+            // to be removed to `id_to_remove`. After the loop, we simply rebuild the vector again,
+            // excluding all elements equal to `id_to_remove`.
+            //
+            // This will be equally effective with one removed element, and with more of them, as
+            // it will allow us to allocate a vector with the desired capacity only once.
+            //
+            // SAFETY: We have already checked that empty_tabs is not empty
+            let id_to_remove = unsafe { *empty_tabs.get_unchecked(0) };
```
bugadani, line 169: I'd like you to replace the unsafe code and the outer if with an `if let Some(id_to_remove) = empty_tabs.get(0)` that does the same without introducing unsafe.

JustForFun88: Indeed, how did I not guess?

JustForFun88:
@bugadani Ready for another review

PR #1866, opened 2022-12-27
use inline format args

danieleades:


panekj:
I don't think it's a necessary change, it only adds noise to git log/blame

danieleades:
> I don't think it's a necessary change, it only adds noise to git log/blame

*Not* making this change adds a lot of noise when linting with clippy, as this is a default lint. This degrades the usefulness of linting for contributors.

If you don't want this change, then this lint should be explicitly suppressed for the whole repo

panekj:
> If you don't want this change, then this lint should be explicitly suppressed for the whole repo

There is no lint that complains about `format!` macro
image

danieleades:
> > If you don't want this change, then this lint should be explicitly suppressed for the whole repo
> 
> There is no lint that complains about `format!` macro image

there sure is. it's not *yet* in the version of clippy you're *currently* pinned too

here are the docs - https://rust-lang.github.io/rust-clippy/master/#uninlined_format_args

try `cargo +nightly clippy -- -D warnings`.

> and I fix clippy after each new release (#1815, #1317)

That's fine, but that only kicks the can down the road.

see #1871

panekj:
> try `cargo +nightly clippy -- -D warnings`.

I know, but we don't use nightly

https://github.com/lapce/lapce/blob/master/rust-toolchain.toml

danieleades:
> > try `cargo +nightly clippy -- -D warnings`.
> 
> I know, but we don't use nightly
> 
> https://github.com/lapce/lapce/blob/master/rust-toolchain.toml

this lint is already in `beta`. You've configured a rolling window in `rust-toolchain.toml` not a pinned version, so you'll get this lint sometime within the next 6 weeks.

see https://github.com/lapce/lapce/pull/1871#issuecomment-1366425887

panekj:
> > > try `cargo +nightly clippy -- -D warnings`.
> > 
> > 
> > I know, but we don't use nightly
> > [`master`/rust-toolchain.toml](https://github.com/lapce/lapce/blob/master/rust-toolchain.toml?rgh-link-date=2022-12-28T07%3A18%3A36Z)
> 
> this lint is already in `beta`. You've configured a rolling window in `rust-toolchain.toml`, not a pinned version, so you'll get this lint sometime within the next 6 weeks.
> 
> see [#1871 (comment)](https://github.com/lapce/lapce/pull/1871#issuecomment-1366425887)

see https://github.com/lapce/lapce/pull/1871#issuecomment-1366433237

PR #1867, opened 2022-12-27
remove explicit iter loops

danieleades:


panekj:
Does this have any meaningful impact such as improved performance or is it only a cosmetic change?

danieleades:
> Does this have any meaningful impact such as improved performance or is it only a cosmetic change?

it's cosmetic. Additionally, there's inherent value in suppressing/addressing things clippy complains about because it makes the linting less noisy, more likely to be used, and therefore more able to generate actual value. I would personally argue that's valuable enough in it's own right, though compared to #1871, this clippy 'noise' is at `clippy::pedantic` level, rather than `clippy::all`, so it's a slightly harder case to make.

JustForFun88:
It seems to me that the changes only increase the compilation time, since now the compiler still has to infer the call to these `iter` and `iter_mut` functions itself.

danieleades:
> It seems to me that the changes only increase the compilation time, since now the compiler still has to infer the call to these `iter` and `iter_mut` functions itself.

Have you done some measurement that shows this? I would find an increase in compilation time here to be *very* surprising

JustForFun88:
> Have you done some measurement that shows this? I would find an increase in compilation time here to be _very_ surprising

Yes, my mistake, you are right 😊. Looks like the `for` loop always results in an implicit call to `IntoIterator`, so both constructs are compile-time equivalent.

lapce-core/src/syntax/util.rs
```diff
@@ -123,12 +123,12 @@ pub fn has_unmatched_pair(line: &str) -> bool {
             }
         }
     }
-    for (_, pair_count) in count.iter() {
+    for (_, pair_count) in &count {
```
Arjun31415, line 126: I believe you need to use `count.values()`

Arjun31415 (approved):

panekj (changes requested):
waiting for next clippy version

panekj:
Closing since those lints don't seem to be default anytime soon

PR #1868, opened 2022-12-27
use 'Self' keyword to refer to own type

danieleades:


panekj:
Same as #1866, I believe it only adds noise without any value

PR #1871, opened 2022-12-28
suppress 'clippy::uninlined_format_args'

danieleades:
this is a new lint in `clippy::all`, which is `warning` level by default. This repo is using a pinned version of clippy, which suppresses the lint *right now*, but not for a whole lot longer.

https://rust-lang.github.io/rust-clippy/master/#uninlined_format_args

This lint adds a lot of noise when linting with clippy and default settings, so it should be addressed or suppressed to reduce linting noise for downstream users.

#1866 addresses the lint, but this solution was not preferred. This PR suppresses the lint instead.

### DIFF
--- a/lapce-data/src/lib.rs
+++ b/lapce-data/src/lib.rs
@@ -1,4 +1,4 @@
-#![allow(clippy::manual_clamp)]
+#![allow(clippy::manual_clamp, clippy::uninlined_format_args)]
 
 pub mod about;
 pub mod alert;
--- a/lapce-proxy/src/lib.rs
+++ b/lapce-proxy/src/lib.rs
@@ -1,4 +1,4 @@
-#![allow(clippy::manual_clamp)]
+#![allow(clippy::manual_clamp, clippy::uninlined_format_args)]
 
 pub mod buffer;
 pub mod dispatch;
--- a/lapce-ui/src/lib.rs
+++ b/lapce-ui/src/lib.rs
@@ -1,4 +1,4 @@
-#![allow(clippy::manual_clamp)]
+#![allow(clippy::manual_clamp, clippy::uninlined_format_args)]
 
 pub mod about;
 pub mod alert;


panekj:
I would prefer to not workaround issues of local user setups by modifying our repo-level lint settings

danieleades:
> I would prefer to not workaround issues of local user setups by modifying our repo-level lint settings

the configured `rust-toolchain.toml` file is used to specify a 'blessed' version of the compiler, known to build successfully and pass all the tests. But it's a very common pattern to use a more recent compiler version for linting and formatting. This is why clippy has an "`msrv`" configuration option (it also parses the `package.rust-version` key from `Cargo.toml` if present) - the purpose of this configuration value is to allow you to run a later version of clippy than your codebase is built against, and suppress lints that would introduce unsupported syntax.

Also, this repo enforces the `stable` compiler, but doesn't pin to a specific version of the stable compiler- it's a rolling window. It's only a matter of time before this lint hits stable and since the github action in this repo uses `-D warnings`, the CI job will start failing the moment it does.

The lint is already in the beta version, so you've got *at most* 6 weeks before it hits stable.

panekj:
> The lint is already in the beta version, so you've got _at most_ 6 weeks before it hits stable.

I'm aware, but I prefer to apply lints that are actually required by Clippy, so when they hit stable (as some lints might change during that time).
As for the rest, we have a specific setup that requires stable compiler, while they might be some quirks to it it's no different than any other rust or non-rust project.
If you you use `nightly` compiler, I'm going to tell you to not do that here because we don't use nightly compiler versions.

danieleades:
> I'm aware, but I prefer to apply lints that are actually required by Clippy, so when they hit stable (as some lints might change during that time).

That would be true if the lint was in nightly, and not yet in beta. Since the lint is in beta, the chance of it changing before hitting stable is extremely slim

> If you you use `nightly` compiler, I'm going to tell you to not do that here because we don't use nightly compiler versions.

i'm using it only for linting, and only then so that I can address lints *before* they break the CI

I'm not sure i understand what your concern about is about adding these three lines, but i guess the other option is to leave this PR open for the next month or so, and *then* merge it.

panekj:
> I'm not sure i understand what your concern about is about adding these three lines

The concern is that it's not required for our project setup and if anything changes, it has to be fixed again.
Also I don't want to suppress lints until it's absolutely necessary (e.g. #1817)

danieleades:
> > I'm not sure i understand what your concern about is about adding these three lines
> 
> The concern is that it's not required for our project setup and if anything changes, it has to be fixed again. Also I don't want to suppress lints until it's absolutely necessary

righto. see you on january 26th.

> (e.g. #1817)

There's no context in this PR so it's difficult to understand the motivation for reverting it. Seems to me, when taken in aggregate, that not using more modern, concise syntax where available simply represents an accumulation of technical debt.

panekj:
> righto. see you on january 26th.

Bye!

> There's no context in this PR so it's difficult to understand the motivation for reverting it.

The context is that I applied a lint that was stable which broke lapce so I had to revert it and disable it.

PR #1879, opened 2022-12-30
Move from `String` to `VoltID`

JustForFun88:
Small, preliminary, yet rather cosmetic improvements that improve the readability of the code and the detection of potential errors (#1881).

The main improvement here is that now `Volts` is identified not just by a `String` (which is fraught with errors), but by a structure with its own separate fields.

Further, in the next pull requests, I expect to go further and remove unnecessary cloning and string fetching in many places, since it is obvious that VoltID can be directly compared with `VoltInfo` or `VoltMetadata` (for example in `HashMaps`) without an intermediate buffer, etc. (for example, we can refactor here https://github.com/lapce/lapce/blob/master/lapce-ui/src/window.rs#L334).

P.S. The pull request may seem like a lot, but I actually just added a few tests.


panekj:
I wouldn't consider calling it `VoldIdText` an improvement

JustForFun88:
> I wouldn't consider calling it `VoltIdText` an improvement

Yes, you are right, at the moment this is more of a cosmetic improvement, although it improves both the readability of the code and the detection of potential errors (#1881).  The main improvement here is that it's not just a string, but a structure with its own separate fields.  Further, in the next pull requests, I expect to go further and remove unnecessary cloning and string allocations in many places, since it is obvious that `VoltIdText` can be directly compared to `VoltInfo` or `VoltMetadata` (as example in `HashMaps`) without an intermediate buffer.

Though I can not absolutely correctly understood you and you mean the name of structure?

panekj:
I meant going from `VoltID` to `VoltIdText`

JustForFun88:
> I meant going from `VoltID` to `VoltIdText`

Went back to using `VoltID`

lapce-proxy/src/plugin/wasi.rs
```diff
@@ -1,3 +1,6 @@
+#[cfg(test)]
+mod tests;
```
dzhou121, line 2: This shouldn't be here?

JustForFun88: At the beginning of the file? I think it's ok. We won’t find it at the very end, because I put the tests in a separate file.

JustForFun88: Here is an example of how a module starts with a test submodule declaration (https://github.com/rust-lang/rust/blob/master/library/std/src/collections/hash/map.rs#L2)

PR #1880, opened 2022-12-30
Add items to MacOS menu bar

joshuagawley:

### DIFF
--- a/lapce-ui/src/app.rs
+++ b/lapce-ui/src/app.rs
@@ -259,8 +259,9 @@ fn macos_window_desc<T: druid::Data>(desc: WindowDesc<T>) -> WindowDesc<T> {
                     },
                     Target::Auto,
                 )))
+                .separator()
                 .entry(
-                    Menu::new("Settings")
+                    Menu::new("Settings...")
                         .entry(MenuItem::new("Open Settings").command(Command::new(
                             LAPCE_COMMAND,
                             LapceCommand {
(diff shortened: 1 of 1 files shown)


lapce-ui/src/app.rs
```diff
@@ -247,159 +247,185 @@ fn macos_window_desc<T: druid::Data>(desc: WindowDesc<T>) -> WindowDesc<T> {
...
+                            MenuItem::new("Hide Lapce")
+                                .command(druid::commands::HIDE_APPLICATION)
+                                .hotkey(SysMods::Cmd, "h"),
+                        )
+                        .separator()
+                        .entry(
+                            MenuItem::new("Quit Lapce")
+                                .command(druid::commands::QUIT_APP)
+                                .hotkey(SysMods::Cmd, "q"),
+                        ))
```
panekj, line 300: This doesn't look right

panekj (changes requested):

joshuagawley:
The first new commit has the code reformatted to reflect the menu hierarchy; the second commit makes minor changes to the app menu to conform more to the Human Interface Guidelines (mainly just adding a couple separators here and there and adding the "Hide Others" and "Show All" commands.

This is what the menu bar looks like after these changes:
Screenshot 2022-12-30 at 11 15 20

(NB: the "Settings..." submenu is taken from VSCode which uses the same concept, and also reflects the structure used in Lapce's existing settings menu):

lapce-ui/src/app.rs
```diff
@@ -247,159 +247,196 @@ fn macos_window_desc<T: druid::Data>(desc: WindowDesc<T>) -> WindowDesc<T> {
...
+                                LAPCE_COMMAND,
+                                LapceCommand {
+                                    kind: CommandKind::Workbench(
+                                        LapceWorkbenchCommand::OpenKeyboardShortcuts,
+                                    ),
+                                    data: None,
+                                },
+                                Target::Auto,
+                            ))
+                            .hotkey(SysMods::Cmd, "k")
```
panekj, line 288: This deviates from our default keymap for macos
https://github.com/lapce/lapce/blob/8044853ff0d4dfd49b6e97815e4db8ca16b6f818/defaults/keymaps-macos.toml#L26-L28

lapce-ui/src/app.rs
```diff
@@ -247,159 +247,196 @@ fn macos_window_desc<T: druid::Data>(desc: WindowDesc<T>) -> WindowDesc<T> {
...
-                    .entry(MenuItem::new("Find").command(Command::new(
-                        LAPCE_COMMAND,
-                        LapceCommand {
-                            kind: CommandKind::Focus(FocusCommand::Search),
-                            data: None,
-                        },
-                        Target::Auto,
-                    ))),
-            )
+                            .hotkey(SysMods::Cmd, "q")
```
panekj, line 309: ditto

panekj: I forgot that macOS platform has this binding no matter what we do, so I guess this one can stay?

PR #1883, opened 2022-12-30
Refactor command.rs

amab8901:

### DIFF
--- a/lapce-data/src/command.rs
+++ b/lapce-data/src/command.rs
@@ -63,40 +63,40 @@ pub const LAPCE_UI_COMMAND: Selector<LapceUICommand> =
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct LapceCommand {
-    pub kind: CommandKind,
     pub data: Option<Value>,
+    pub kind: CommandKind,
 }
 
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub enum CommandKind {
-    Workbench(LapceWorkbenchCommand),
     Edit(EditCommand),
-    Move(MoveCommand),
     Focus(FocusCommand),
     MotionMode(MotionModeCommand),
+    Move(MoveCommand),
     MultiSelection(MultiSelectionCommand),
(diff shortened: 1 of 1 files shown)


panekj:
This seems unnecessary

amab8901:
> This seems unnecessary

the code is currently messy and hard to find around. Refactoring makes it easier for newcomers like myself to find a specific item in a long list, given that I haven't memorized the whole list

panekj:
> > This seems unnecessary
> 
> the code is currently messy and hard to find around. Refactoring makes it easier for newcomers like myself to find a specific item in a long list, given that I haven't memorized the whole list

this just sorts everything and adds noise in git log, I don't see how that hinders ability to use search or go to symbol

amab8901:
Imagine having a dictionary where the words are just put there in random order. It would be kinda hard to get an overview of what's available when learning the codebase

panekj:
but this is not a dictionary, it's a codebase, which you can easily ctrl+f or go by reference/symbol

amab8901:
> but this is not a dictionary, it's a codebase, which you can easily ctrl+f or go by reference/symbol

if I know the name of the item, then I can use ctrl+F. But what if I'm exploring the list to see what's available?

panekj:
Then you can read? Still it doesn't require everything to be sorted

amab8901:
> Then you can read? Still it doesn't require everything to be sorted

yeah I can read, but I think that having it sorted makes it easier for me to navigate it. Let's suppose I'm looking for an item that begins with `Update...`. If it's sorted, then I have them all in one place instead of having to jump around to different parts of the code

amab8901:
this refactoring also sorts items by how many arguments they take in. Let's suppose I'm looking for an item that takes 3 arguments as input. I have them all in one place so that there are fewer items I need to go through

amab8901:
> this just sorts everything and adds noise in git log,

the noise in git log can be solved by git rebase, where you merge this commit into one of its neighbors in the git log

PR #1888, opened 2022-12-31
fix: store lapce data next to lapce for portable install

panekj:

### DIFF
--- a/lapce-core/src/directory.rs
+++ b/lapce-core/src/directory.rs
@@ -6,8 +6,21 @@ use crate::meta::NAME;
 pub struct Directory {}
 
 impl Directory {
+    #[cfg(not(feature = "portable"))]
     fn project_dirs() -> Option<ProjectDirs> {
         ProjectDirs::from("dev", "lapce", &NAME)
+    }
+
+    /// Return path adjacent to lapce executable when built as portable
+    #[cfg(feature = "portable")]
+    fn project_dirs() -> Option<ProjectDirs> {
+        if let Ok(current_exe) = std::env::current_exe() {
+            if let Some(parent) = current_exe.parent() {
+                return ProjectDirs::from_path(parent.join("lapce-data"));
+            }
+            unreachable!("Couldn't obtain current process parent path");
+        }
+        unreachable!("Couldn't obtain current process path");
     }
 
     // Get path of local data directory
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,13 +13,19 @@ concurrency:
 
 jobs:
   build:
-    name: Rust on ${{ matrix.os }}
+    name: Rust on ${{ matrix.os }} (${{ join(matrix.features, ',') }})
     if: github.event.pull_request.draft == false
     needs: [fmt, clippy]
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-latest, windows-latest, macos-latest ]
+        os: [ ubuntu-latest, macos-latest ]
+        include:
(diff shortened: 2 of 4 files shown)


lapce-core/src/directory.rs
```diff
@@ -6,10 +6,26 @@ use crate::meta::NAME;
...
+    #[cfg(any(
+        not(target_os = "windows"),
+        all(target_os = "windows", not(feature = "portable"))
+    ))]
     fn project_dirs() -> Option<ProjectDirs> {
         ProjectDirs::from("dev", "lapce", &NAME)
     }
 
+    /// Return path adjacent to lapce executable when built as portable
+    #[cfg(all(target_os = "windows", feature = "portable"))]
```
sh7dm, line 18: Why is it Windows-only? Linux binary can work fine as well. Maybe just search for data dir as I did?

panekj: no

sh7dm: What's wrong with portability being cross-platform? It can be useful for people who use a USB drive to carry the whole dev setup between computers.

panekj: This is not a supported scenario

sh7dm: And how does supporting this on all platforms interfere with something? Maybe just make it an option for all builds while enabling by default in the `Windows Portable` builds only as the expected behavior would be?

panekj: > And how does supporting this on all platforms interfere with something?

Some platforms are specifically designed to be incompatible with the concept of portable executables

> Maybe just make it an option for all builds while enabling by default in the `Windows Portable` builds only as the expected behavior would be?

We have feature gates for specific platforms known to work

sh7dm: > Some platforms are specifically designed to be incompatible with the concept of portable executables

Rust Linux binaries are forwards-compatible and pretty much portable. The usecase I suggested is, for example, sharing the configured editor setup between Linux and Windows machines (keeping two executables together with a `data` directory). Anyway, if you think that feature introduces maintenance issues, feel free avoiding it.

> We have feature gates for specific platforms known to work

Didn't fully understand. I meant that any build should use data directory when provided with, but Windows Portable build should default to this behaviour and create a directory anyway. That should not introduce any regression on other platforms and even Windows Installable builds, so 100% fine to do, because when you don't place a data directory alongside the executable, it runs as before.

panekj: > Rust Linux binaries are forwards-compatible and pretty much portable.

I'm sorry but lol. Linux (and probably *BSD too) are the only platforms exclusively hostile towards any kind of portability and the fact that something built on one OS works on another is miracle.
We build Lapce for Linux on Ubuntu 18.04 so the dylibs are old enough for each distro to work.

> Didn't fully understand. I meant that any build should use data directory when provided with, but Windows Portable build should default to this behaviour and create a directory anyway. That should not introduce any regression on other platforms and even Windows Installable builds, so 100% fine to do, because when you don't place a data directory alongside the executable, it runs as before.

I don't want that behaviour

sh7dm: > We build Lapce for Linux on Ubuntu 18.04 so the dylibs are old enough for each distro to work

Exactly. Have also been waiting long for Rust to support linking against older glibc. This way (just build on an old system) works just fine, the point of Rust is to build everything together that way.

VSCode (from which the portable feature originates) supports this feature on all supported platforms.

> I don't want that behaviour

Well, your choice as a maintainer is to decline the feature, I only suggested it

PR #1893, opened 2023-01-01
Make clippy happy

amab8901:


panekj:
Clippy is not happy, for a good reason

amab8901:
oh, haha it was happy here locally after I removed those lines. Idk why the outome is different on the CI

PR #1895, opened 2023-01-01
feat: select name only when renaming a file

Domerpig:

Adds a settings under the core section that makes it possible to only select the name of a file when renaming. I am not completely sure if "core" is the right section for it, but the others fit even worse. 
The option is on by default.


panekj:
Why do we need a setting for that?

Domerpig:
The todo stated, that a configrutation option could be added for this. If the setting is not needed, I can remove it.

PR #1930, opened 2023-01-06
Added primary buffer support to the editor when compiled for linux

### LINKED ISSUE #834: Middle mouse button copy and paste under X11 not working
One convenient feature under X11 is that I can select text in one application (without any explicit copy command) and paste it into another application by clicking the middle mouse button.
This uses the so called "PRIMARY" "[selection](https://www.x.org/releases/X11R7.6/doc/xorg-docs/specs/ICCCM/icccm.html#peer_to_peer_communication_by_means_of_selections)"  in contrast to the "CLIPBOARD" "selection" (I've put selection in quotes here, as it refers to the selections from the X11 Inter-Client Communication Conventions Manual) used for the explicit crtl-c ctr-v copy and paste.

Currently you can neither copy text from, nor paste text into lapce using this mechanism.

https://specifications.freedesktop.org/clipboards-spec/clipboards-latest.txt has some more details on the commonly used behaviour:
 - selecting but with no explicit copy should only set PRIMARY, never CLIPBOARD
 - middle mouse button should paste PRIMARY, never CLIPBOARD
 - explicit cut/copy commands (i.e. menu items, toolbar buttons) should always set CLIPBOARD to the currently-selected data (i.e. conceptually copy PRIMARY to CLIPBOARD)
 - explicit cut/copy commands should always set both CLIPBOARD and PRIMARY, even when copying doesn't involve a selection (e.g. a "copy url" -option which explicitly copies an url without the url being selected first)
 - explicit paste commands should paste CLIPBOARD, not PRIMARY
 - a selection becoming unselected should never unset PRIMARY

I'm not sure if this is issue with lapce itself, or rather with druid or druids X11 backend.

floriankramer:
This pr adds support for the primary clipboard which is present in most linux systems. This allows for pasting text selected in any application using the middle mouse button. This fixes #834.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2649,6 +2649,7 @@ dependencies = [
  "directories",
  "dmg",
  "druid",
+ "druid-shell",
  "flate2",
  "fs_extra",
  "fuzzy-matcher",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,6 +75,9 @@ branch = "shell_opengl"
 # path = "../../druid/druid"
 features = ["svg", "im", "serde"]
 
+[workspace.dependencies.druid-shell]
+git = "https://github.com/lapce/druid"
+branch = "shell_opengl"
 
 [profile.release-lto]
 inherits = "release"
--- a/lapce-core/src/command.rs
+++ b/lapce-core/src/command.rs
@@ -75,6 +75,12 @@ pub enum EditCommand {
     Paste,
     #[strum(serialize = "paste_before")]
     PasteBefore,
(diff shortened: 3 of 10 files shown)


floriankramer:
Currently this pr only works for the editor window itself, and the way text is copied into the primary buffer is not that nice. The copy is triggered on mouseup and after keydown events. A much nicer way of doing this could be an integration with cursor, which stores the current selection. That'd be a (possibly) larger change though, and I don't know the architecture of lapce well enough to be sure it'd be a good way of prettying this pr up.

Issue #1910, opened 2023-01-02
Spaces in path break definition lookup


SkaneroOo:
## Lapce Version

0.2.5

## System information

Windows 11

## Describe the bug
If path to file containing definition contains spaces, lapce replaces them with %20 and tries to open file in location, that doesn't exist

## Additional information

While trying to access definition of EntityPosition struct, instead of moving cursor few lines higher, lapce opened new tab with file, that doesn't exist. Original file path is in terminal, opened path is at the top of right tab. 
When I've removed space in directory name, everything worked fine.

![Screenshot (5)](https://user-images.githubusercontent.com/40175161/210282856-8f07dfad-f7ba-43aa-8ee8-0c60d0ed2537.png)

TheIronBorn:
Is anyone else working on this? I think I could manage this if not. I welcome any tips though

panekj:
Fixed via f2355aa0b8839aacc5fa684db2c8b75b5258e251

PR #1932, opened 2023-01-06
Fix GotoTypeDefinition that contains percent-encoded characters

hbina:
Simply use the provided function by Url. Do you want to propagate this upwards?

Solves: https://github.com/lapce/lapce/issues/1910

### DIFF
--- a/lapce-data/src/proxy.rs
+++ b/lapce-data/src/proxy.rs
@@ -919,22 +919,7 @@ impl Remote for WslRemote {
 // This function strips the additional / from the beginning, if the first segment is a drive letter.
 #[cfg(windows)]
 pub fn path_from_url(url: &Url) -> PathBuf {
-    let path = url.path();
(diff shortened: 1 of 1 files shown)


lapce-data/src/proxy.rs
```diff
@@ -914,30 +914,6 @@ impl Remote for WslRemote {
     }
 }
 
-// Rust-analyzer returns paths in the form of "file:///<drive>:/...", which gets parsed into URL
-// as "/<drive>://" which is then interpreted by PathBuf::new() as a UNIX-like path from root.
-// This function strips the additional / from the beginning, if the first segment is a drive letter.
-#[cfg(windows)]
-pub fn path_from_url(url: &Url) -> PathBuf {
```
dzhou121, line 921: I think this is for when you connect to a Linux remote from a Windows machine. The proper way to fix it is probably using Url throughout Lapce.

PR #1936, opened 2023-01-07
fix: do not ignore result when cursor is at the beginning of the match

RoPP:


dzhou121:
This change will make it not jump to the next match if the cursor is at the beginning of the match.

PR #1939, opened 2023-01-08
Register buffer when saving-as a new file

### LINKED ISSUE #1750: Newly created file cannot be saved
## Lapce Version

nightly-50216c4

## System information

Debian Sid

## Describe the bug
A newly created file can be saved only once. Further modifications do not get saved.
Steps to reproduce the problem
1. Open a new file in lapce.
2. Type in some text and save it.
3. Make further modifications and try to save it using the key binding.
4. The changes are not saved.
5. Close the editor tab using the key binding or the close button on the tab.
6. The prompt to save the file appears, but clicking on save does not save the changes. Also the editor tab does not get closed.

At this point, closing lapce itself works, and file modification prompt does not show up. 
Opening lapce again and then modifying the file and saving it works.

MinusGix:

Fixes #1750   
The issue was that the new file was not registered, thus causing future attempts to update the file to error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 ### Bug Fixes
 - [#1911](https://github.com/lapce/lapce/pull/1911): Fix movement on selections with left/right arrow keys
+- [#1939](https://github.com/lapce/lapce/pull/1939): Fix saving/editing newly saved-as files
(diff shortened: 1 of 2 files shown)


dzhou121:
Good catch!

PR #1940, opened 2023-01-08
cargo: migrate to workspace.dependencies

panekj:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -39,17 +39,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "ahash"
-version = "0.8.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bf6ccdb167abbf410dcb915cabd428929d7f6a04980b54a11f26a39f1c7f7107"
-dependencies = [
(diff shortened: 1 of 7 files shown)


panekj:
⚠ DO NOT MERGE ⚠

panekj:
`textwrap` and `dyn-clone` are transitive deps, they got updated by cargo but the bump is small and I don't want to deal with cargo anymore so I leave them like that
moved all shared deps under `workspace.dependencies`, now we can control version for all of them in single place, it also revealed few inconsistencies, fixed in this pr
I also removed few properties under `package` and set to inherit them from `workspace.package`

panekj:
I also replaced all branch/versions for git submodules in tree-sitter grammars to locked revisions, otherwise it often is bumped accidentally by people who try to add new grammar and I have to fix it for them
(and because cargo updated them for me too, grrr...)

panekj:
`workspace.dependencies` is a rust/cargo 1.64 feature, on which we already depend, so no `rust-version` bump needed

panekj:
locked tree-sitter grammars to specific revision should also prevent our builds fail when someone force pushes the branch

PR #1944, opened 2023-01-09
feat: add branches filter

tk103331:

add branches filter

Input UI:
image

filter result: 

image

### DIFF
--- a/lapce-data/src/data.rs
+++ b/lapce-data/src/data.rs
@@ -789,6 +789,13 @@ impl LapceTabData {
             event_sink.clone(),
         );
         main_split.add_editor(
+            title.branches.filter_editor,
+            None,
+            LocalBufferKind::BranchesFilter,
+            &config,
+            event_sink.clone(),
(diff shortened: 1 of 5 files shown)


runiorr:
Hey, I think there's a bug when we write in the filter input, close it and open again.

If I write "terminal", it shows only one branch. Then I click away to minimize and click on the branches icon again, "terminal" is still there, but now all branches have appeared, ignoring the filter.

I don't know which should be the normal behavior, keep the filter on or clean the input field.

I managed to workaround by moving your code inside the "Event::Timer" to "LapceUICommand::ShowGitBranches".
But its generating the filter again every time it's clicked. I don't know how to save on the state to avoid this redundancy.

panekj:
> Hey, I think there's a bug when we write in the filter input, close it and open again.

Please create issue instead of replying in merged PRs

tk103331:
Maybe the filter should be cleared in ShowGitBranches

PR #1946, opened 2023-01-09
Add more info to .desktop file

m00nwtchr:
Added mime-types (allows opening lapce from e.g. a file manager/browser, "Open With"), new window action, and generic name

### DIFF
--- a/extra/linux/dev.lapce.lapce.desktop
+++ b/extra/linux/dev.lapce.lapce.desktop
@@ -11,9 +11,9 @@ Icon=dev.lapce.lapce
 Exec=lapce
 Terminal=false
 MimeType=text/plain;inode/directory;
-Actions=new-empty-window;
+Actions=new-window;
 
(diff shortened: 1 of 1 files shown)


extra/linux/dev.lapce.lapce.desktop
```diff
@@ -5,7 +5,15 @@ Type=Application
...
 
 Icon=dev.lapce.lapce
 Exec=lapce
 Terminal=false
+MimeType=text/plain;inode/directory;
+Actions=new-empty-window;
+
+[Desktop Action new-empty-window]
+Name=New Empty Window
+Exec=lapce --new %F
```
panekj, line 18: That doesn't seem to be "empty"?

m00nwtchr: It depends on the context the action is used in, but i guess just "new window" would be better (I based this on the vscode desktop files)

Issue #1951, opened 2023-01-10
hover triggered when renaming symbols


tk103331:
## Lapce Version

Version 0.2.5

## System information

macOS 13.1

## Describe the bug

When I am renaming a symbol, the mouse moves over the input box, the hover behind the input box will be triggered

## Additional information

<img width="712" alt="image" src="https://user-images.githubusercontent.com/4404609/211442675-db2dd0ab-e7f9-4966-affe-116366006554.png">

PR #1954, opened 2023-01-10
fix: trigger hover when renaming

tk103331:

disable trigger hover when renaming，only trigger hover when focus on editor, fix: https://github.com/lapce/lapce/issues/1951

image


dzhou121:
This will disable hover for inactive tab when you have split. 

The real issue for #1951 I think is the MouseMove event is caught by the editor when rename is there.

tk103331:
check mouse within rename when update_hover

PR #1965, opened 2023-01-11
ci: fetch only locked dependencies

panekj:


dzhou121:
What is the benefit of this, just for my understanding?

panekj:
It should not grab latest transitive or direct dependencies which are not locked to full version.

PR #1971, opened 2023-01-12
Change cursor up/down behavior for the first/last line

alvra:
Currently, pressing up/down while on the first/last line does nothing.
This PR changes that so the cursor goes to the start/end of the line, as is common in other editors.


alvra:
Thank you :)

dzhou121:
Thank you for putting this in!

PR #1972, opened 2023-01-12
Implement file duplication

alvra:
This PR implements duplicating files in the fs tree context menu.

### DIFF
--- a/lapce-data/src/explorer.rs
+++ b/lapce-data/src/explorer.rs
@@ -1,6 +1,5 @@
 use std::{
     collections::HashMap,
-    ffi::OsString,
     path::{Path, PathBuf},
     sync::Arc,
 };
@@ -49,7 +48,7 @@ pub enum Naming {
         /// The folder that the file/directory is being created within
         base_path: PathBuf,
         /// The name of the file being duplicated
-        name: OsString,
+        name: String,
     },
 }
 impl Naming {
@@ -449,7 +448,7 @@ impl FileExplorerData {
         list_index: usize,
         indent_level: usize,
         base_path: PathBuf,
-        name: OsString,
+        name: String,
     ) {
         self.cancel_naming();
         self.naming = Some(Naming::Duplicating {
@@ -464,7 +463,7 @@ impl FileExplorerData {
             .local_docs
             .get_mut(&LocalBufferKind::PathName)
             .unwrap();
-        Arc::make_mut(doc).reload(Rope::from(name.to_string_lossy()), true);
+        Arc::make_mut(doc).reload(Rope::from(name), true);
 
         // TODO: We could provide a configuration option to only select the filename at first,
         // which would fit a common case of just wanting to change the filename and not the ext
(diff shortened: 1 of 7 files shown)


panekj:
I don't think adding duplication is a proper thing to do, I'd rather see `Copy`/`Paste`/`Cut` actions like in standard file explorers

alvra:
What would be the difference between file duplication and copy? Do you mean copy to clipboard?

panekj:
> What would be the difference between file duplication and copy? 

Duplication implements very specific action which can be replaced by clipboard actions

> Do you mean copy to clipboard?

Yes

dzhou121:
I personally feel it's a good addition to copy/paste/cut, because I often do copy and paste which is essentially duplicate.

panekj:
> I personally feel it's a good addition to copy/paste/cut, because I often do copy and paste which is essentially duplicate.

It's not bad per se, but I would like to avoid adding it to mouse context menu. As we develop we will be adding more actions which could be more useful and cluttering context menu with too many actions feels wrong.

alvra:
True, duplication is more specific than copy+paste. Personally I use duplicate much more often and having it as a single option is helpful to me. New files often share code with its siblings so I probably use duplicate more even than "new empty file".

dzhou121:
> > I personally feel it's a good addition to copy/paste/cut, because I often do copy and paste which is essentially duplicate.
> 
> It's not bad per se, but I would like to avoid adding it to mouse context menu. As we develop we will be adding more actions which could be more useful and cluttering context menu with too many actions feels wrong.

I don't feel it would be too much with one more menu item in the copy/paste/cut section(in the future).

panekj:
> New files often share code with its siblings so I probably use duplicate more even than "new empty file".

We should probably hide `New ...` when targeting files like VSCode and show it on directories/empty space only

dzhou121:
> > New files often share code with its siblings so I probably use duplicate more even than "new empty file".
> 
> We should probably hide `New ...` when targeting files like VSCode and show it on directories/empty space only image image

Yeah in a separate PR maybe.

panekj:
> > > I personally feel it's a good addition to copy/paste/cut, because I often do copy and paste which is essentially duplicate.
> > 
> > 
> > It's not bad per se, but I would like to avoid adding it to mouse context menu. As we develop we will be adding more actions which could be more useful and cluttering context menu with too many actions feels wrong.
> 
> I don't feel it would be too much with one more menu item in the copy/paste/cut section(in the future).

If we're not going to implement custom context actions from plugins, then yes, it should stay fine but that's quite a limitation IMO

lapce-data/src/explorer.rs
```diff
@@ -38,12 +39,25 @@ pub enum Naming {
...
+    Duplicating {
+        /// The index that the file being created should appear at
+        /// Note that when duplicating, it is not yet actually created.
+        list_index: usize,
+        /// Indentation level
+        indent_level: usize,
+        /// The folder that the file/directory is being created within
+        base_path: PathBuf,
+        /// The name of the file being duplicated
+        name: OsString,
```
panekj, line 52: This should be a `String` rather than `OsString`, we don't operate on raw platform strings (at least not yet)

dzhou121:
Can you please resolve the CHANGELOG.md conflict?

PR #1979, opened 2023-01-13
Feat increase diff extend lines

tk103331:

feature: support increase diff extend size

image

click the area will increase the diff extend size, now wil plus 5 lines.

image


dzhou121:
Thanks so much for working on this. It will be a very good feature. There is one problem that it will expand all diff sections.

tk103331:
This problem is fixed, the screenshot above may be confused, I will update later.

dzhou121:
Yeah it works now. Amazing work. I'll merge it now.

There is one thing I noticed which can be improved in the future. The history extend will be reset when you do an edit.

dzhou121:
You might want to store it in the document so that it doesn't get reset.

PR #2008, opened 2023-01-16
Bump Rust version to 1.65

JustForFun88:
Needed for #2007. One more step towards removing an unnecessary vector here (https://github.com/lapce/lapce/blob/master/lapce-ui/src/window.rs#L335) and further in others places.


panekj:
We've `let`-`else` syntax introduced somewhere already so we should bump anyway

PR #2020, opened 2023-01-16
Limit global search to 1 worker thread

hbina:
Instead of spawning a new thread for every keystroke.

### DIFF
--- a/lapce-proxy/src/dispatch.rs
+++ b/lapce-proxy/src/dispatch.rs
@@ -802,9 +802,12 @@ impl Dispatcher {
         let (global_search_tx, global_search_rx) = unbounded::<()>();
 
         let local_proxy_rpc = proxy_rpc.clone();
-        thread::spawn(move || {
-            start_global_search_job(local_proxy_rpc, global_search_rx)
-        });
+        let _ = thread::Builder::new()
+            .name("global search worker".into())
+            .spawn(move || {
+                start_global_search_job(local_proxy_rpc, global_search_rx)
+            })
+            .unwrap();
 
         Self {
             workspace: None,
--- a/lapce-proxy/src/dispatch.rs
+++ b/lapce-proxy/src/dispatch.rs
@@ -341,25 +341,16 @@ impl ProxyHandler for Dispatcher {
                                                 // (such as in minified javascript)
                                                 // Note that the start/end are column based, not absolute from the
                                                 // start of the file.
-                                                let match_start = mymatch
+                                                let display_range = mymatch
                                                     .start()
-                                                    .saturating_sub(100);
-                                                let match_end = line
-                                                    .len()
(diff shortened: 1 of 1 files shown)


lapce-proxy/src/dispatch.rs
```diff
@@ -1265,3 +1227,76 @@ fn git_get_remote_file_url(workspace_path: &Path, file: &Path) -> Result<String>
         file.strip_prefix(workspace_path)?.to_str().unwrap()
     ))
 }
+
+fn start_global_search_job(
+    proxy_rpc: ProxyRpcHandler,
+    global_search_rx: Receiver<()>,
+) {
+    loop {
```
hbina, line 1235: Can someone explain why this loop is necessary here? It doesn't work if you remove it. Yet people keep telling that it should work without it. The documentation says that the channel will yield `None` if the sender is empty AND disconnected. But our sender here lives as long as `Dispatcher` is so I don't see why it should die.

hbina: I am stupid :brain: that is supposed to be a `continue` not a `break`.

lapce-proxy/src/dispatch.rs
```diff
@@ -1265,3 +1227,72 @@ fn git_get_remote_file_url(workspace_path: &Path, file: &Path) -> Result<String>
...
+                                    let mymatch =
+                                        matcher.find(line.as_bytes())?.unwrap();
+                                    // Shorten the line to avoid sending over absurdly long-lines
+                                    // (such as in minified javascript)
+                                    // Note that the start/end are column based, not absolute from the
+                                    // start of the file.
+                                    line_matches.push((
+                                        lnum as usize,
+                                        (mymatch.start(), mymatch.end()),
+                                        line.chars()
```
JustForFun88, line 1272: As with the previous pull request, there is a clear error here.  Regex returns the start of the match in bytes, not characters (https://docs.rs/grep-regex/0.1.10/grep_regex/struct.RegexMatcher.html#method.find).  Take a slice in the range mymatch.start(), mymatch.end(), it will be valid utf-8.

JustForFun88: And use chars iter after taking the slice.

dzhou121:
I would prefer an approach for a new search to cancel the last search rather than waiting for the last search to finish. 

similar to here https://github.com/lapce/lapce/blob/master/lapce-core/src/syntax/highlight.rs#L459

hbina:
@dzhou121 I've updated implementation to be a lot simpler using worker IDs. Each thread is assigned an ID and regularly checks if the atomic u64 matches itself. If not, it cancels its own job. Only the latest job should survive and do the whole search.

dzhou121:
Looks very good!

Issue #2031, opened 2023-01-17
Double-click file to open


Netroxen:
## Is your feature request related to a problem? Please describe.

Sometimes accidental clicks or navigating the UI results in opening files.

## Describe the solution you'd like

Implement a double-click setting allowing the user to set whether a file should be opened using a single or double-click.

## Describe alternatives you've considered

A right-click menu item called "Open Document".

PR #2033, opened 2023-01-18
Implement basic double click delay

hbina:
Fixes https://github.com/lapce/lapce/issues/2031

The state machine is kinda shite. Does anyone have a prior art for something like this?

### DIFF
--- a/lapce-ui/src/explorer.rs
+++ b/lapce-ui/src/explorer.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, path::Path, sync::Arc};
+use std::{collections::HashMap, path::Path, sync::Arc, time::SystemTime};
 
 use druid::{
     menu::MenuEventCtx,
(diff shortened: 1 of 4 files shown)


lapce-ui/src/explorer.rs
```diff
@@ -690,7 +692,50 @@ impl Widget<LapceTabData> for FileExplorerFileList {
                 let file_explorer = Arc::make_mut(&mut data.file_explorer);
                 let index = ((mouse_event.pos.y + self.line_height)
                     / self.line_height) as usize;
+
                 if mouse_event.button.is_left() {
+                    // Check double click delay
+                    let f = || -> bool {
```
dzhou121, line 698: mouse_event has `count` which can be used to check double click or trigger click.

hbina: Awesome. This should be a LOT simpler now.

Issue #1949, opened 2023-01-09
Rename symbol in right click menu


ArniDagur:
Currently, you can rename a symbol using the `F2` keybinding. However, I want to be able to right click a symbol to rename it.

I briefly looked into implementing this myself. I see that there is a `Rename` member inside of `FocusCommand`. However, I don't know why it is not displayed when I right click, as "Go to Type Definition" is.

tk103331:
try add MenuItem like this:

https://github.com/lapce/lapce/blob/242fa766d0904f8048845ad3284a8514404d5bb8/lapce-ui/src/editor.rs#L310-L320

PR #2045, opened 2023-01-22
Add 'Rename Symbol' option on right-click

ArniDagur:
Fixes https://github.com/lapce/lapce/issues/1949

Implementation suggested by https://github.com/lapce/lapce/issues/1949#issuecomment-1376658407. I have confirmed that with this patch I can right click to rename a symbol.

### DIFF
--- a/lapce-ui/src/editor.rs
+++ b/lapce-ui/src/editor.rs
@@ -314,6 +314,15 @@ impl LapceEditor {
                 MenuKind::Item(MenuItem {
                     desc: None,
                     command: LapceCommand {
+                        kind: CommandKind::Focus(FocusCommand::Rename),
+                        data: None,
+                    },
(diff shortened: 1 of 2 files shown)


lapce-ui/src/editor.rs
```diff
@@ -311,6 +311,15 @@ impl LapceEditor {
...
                 MenuKind::Separator,
+                MenuKind::Item(MenuItem {
+                    desc: None,
+                    command: LapceCommand {
+                        kind: CommandKind::Focus(FocusCommand::Rename),
+                        data: None,
+                    },
+                    enabled: true,
+                }),
+                MenuKind::Separator,
```
ArniDagur, line 322: Added separator to match the right click item organization in VSCode.

Issue #1872, opened 2022-12-28
Implement opening single files at a specific line number


VarLad:
## Is your feature request related to a problem? Please describe.

As a follow up to https://github.com/lapce/lapce/pull/1003 it'd be nice if Lapce could open a single file at a specific file number. My particular use case being, some languages support opening specific line at which an error occurs when a function is run in a REPL (Lapce terminal in our case). That in turn could provide a powerful experience when using the Lapce editor with those languages.

## Describe the solution you'd like

Providing arguments `lapce +10 /path/to/file` should open that file on the 10th line. 
This behavior is taken from the Vim editor.

cc: @dzhou121 (since the linked PR is their implementation :))

PR #2052, opened 2023-01-23
Parse into file, line, column

amab8901:

Addresses https://github.com/lapce/lapce/issues/1872


panekj:
Implemented via #1964

PR #2060, opened 2023-01-26
Encode language_id as usize instead of String

hbina:
Pros:
1. This means that it is 3x cheaper to move this thing around
2. Remove the need to wrap this in Option, which removes 1 usize of flag

Cons:
1. It might be harder to debug
2. We need to maintain the mapping between str <-> LanguageId. This can be fixed by introducing Other(String) variation.


panekj:
I will be moving language id to language syntax so this change is not necessary

PR #2061, opened 2023-01-26
Actually logs LogMessage request from plugins

hbina:
Found while trying to debug plugins. Will help with that significantly if we can have the logs.

### DIFF
--- a/lapce-data/src/proxy.rs
+++ b/lapce-data/src/proxy.rs
@@ -179,16 +179,16 @@ impl CoreHandler for LapceProxy {
                 message: LogMessageParams { message, typ },
             } => match typ {
                 MessageType::ERROR => {
-                    log::error!("{}", message)
+                    log::error!("{message}")
                 }
                 MessageType::WARNING => {
-                    log::warn!("{}", message)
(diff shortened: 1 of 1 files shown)


lapce-data/src/proxy.rs
```diff
@@ -175,7 +175,23 @@ impl CoreHandler for LapceProxy {
...
+                }
+                MessageType::WARNING => {
+                    log::warn!("{}", message)
+                }
+                MessageType::INFO => {
+                    log::info!("{}", message)
+                }
+                MessageType::LOG => {
+                    log::debug!("{}", message)
+                }
```
panekj, line 192:
```suggestion
                MessageType::ERROR => {
                    log::error!("{message}")
                }
                MessageType::WARNING => {
                    log::warn!("{message}")
                }
                MessageType::INFO => {
                    log::info!("{message}")
                }
                MessageType::LOG => {
                    log::debug!("{message}")
                }
```

panekj:
this should include plugin in target

hbina:
No idea what that means :3?

PR #2068, opened 2023-01-28
fix: remove remote web url copy command

panekj:

Git doesn't have any guarantees about web interface so the command should not be provided by Lapce


dzhou121:
It doesn't always work but useful for Github, Gitlab etc?

panekj:
> It doesn't always work but useful for Github, Gitlab etc?

Yes, but because of that I think it should not be available as built-in command in Lapce

panekj:
Specifically, only GitHub and GitLab work because they accept URL of `scheme://host/namespace/blob/refspec/filepath`, but other Git web-frontends don't work (Gitea, CGit, etc.)

dzhou121:
Shall we leave it for now. It's quite useful sometimes.

panekj:
Closing in favour of error popup

Issue #2039, opened 2023-01-20
Git does not detect changes when `.git` is not directly in workspace


IsawU:
## Lapce Version

0.2.5

## System information

Fedora 37

## Describe the bug

Git tab does not detect any changes if `.git` directory is not immediately inside the opened workspace directory.

## Additional information

I have a single Git repo for _frontend_ and _backend_ and I open _both_ in separate windows.

I tested both when `.git` is 1 level above workspace and 1 level deep in workspace and in both situations, no changes were detected.

IsawU:
vscode recently changed their behavior in that they do not automatically open repositories in parent directories, but instead require a simple user interaction to open them.

The reasoning is:

> Unfortunately, we have seen that this behaviour can also cause confusion. One such scenario is when one opens a loose file in the home folder, and there is a Git repository in the home directory. In this case the repository from the home directory and can cause confusion as there are lots of changes in the "Source Control" view. Also discarding those changes results in data loss.

You can read the full details of this change [here](https://github.com/microsoft/vscode/issues/172010) and decide if it makes sense to implement something along these lines.

PR #2069, opened 2023-01-28
fix: find repository if it's not in current workspace path

panekj:

Fixes https://github.com/lapce/lapce/issues/2039


dzhou121:
CHANGELOG is conflicting :)

PR #2076, opened 2023-01-29
Added fetch command

jath03:


lapce-proxy/src/dispatch.rs
```diff
@@ -1233,6 +1241,20 @@ fn git_get_remote_file_url(workspace_path: &Path, file: &Path) -> Result<String>
...
+        let mut remote = repo.find_remote(remote_name.unwrap())?;
+        let refspecs: Vec<_> = remote
+            .refspecs()
+            .map(|r| r.str().unwrap().to_string())
+            .collect();
+        remote.fetch(&refspecs, None, None)?;
+    }
+    Ok(())
+}
+
```
panekj, line 1257:
```suggestion
fn git_fetch(workspace_path: &Path) -> Result<()> {
    let repo = Repository::discover(workspace_path)?;

    for remote_name in repo.remotes()?.iter() {
        if let Some(remote) = remote_name {
            let mut remote = repo.find_remote(remote)?;
            let refspecs: Vec<String> = remote
                .refspecs()
                .filter_map(|r| Some(r.str()?.to_string()))
                .collect();
            remote.fetch(&refspecs, FetchOptions::default(), None)?;
        }
    }
    Ok(())
}
```

lapce-proxy/src/dispatch.rs
```diff
@@ -1233,6 +1241,21 @@ fn git_get_remote_file_url(workspace_path: &Path, file: &Path) -> Result<String>
...
+    let repo = Repository::discover(workspace_path)?;
+
+    for remote_name in repo.remotes()?.iter() {
+        if let Some(remote) = remote_name {
+            let mut remote = repo.find_remote(remote)?;
+            let refspecs: Vec<String> = remote
+                .refspecs()
+                .filter_map(|r| Some(r.str()?.to_string()))
+                .collect();
+            remote.fetch(&refspecs, FetchOptions::default(), None)?;
```
panekj, line 1254:
```suggestion
            remote.fetch(&refspecs, Some(&mut FetchOptions::default()), None)?;
```

jath03:
I'd assume leaving that field None would provide the default options.
https://docs.rs/git2/latest/git2/struct.Remote.html#method.fetch

panekj:
Could be, from the library I don't see default settings being equal to `None`

lapce-proxy/src/dispatch.rs
```diff
@@ -1227,6 +1238,34 @@ fn git_get_remote_file_url(workspace_path: &Path, file: &Path) -> Result<String>
...
+        let mut callbacks = RemoteCallbacks::new();
+        callbacks.credentials(|_url, username_from_url, _allowed_types| {
+            Cred::ssh_key(
+                username_from_url.unwrap(),
+                None,
+                std::path::Path::new(&format!(
+                    "{}/.ssh/id_rsa",
+                    env::var("HOME").unwrap()
+                )),
+                None,
```
panekj, line 1259: this isn't right:
1. you should not use `HOME` envvar
2. it doesn't have to be RSA key
3. key doesn't have to exist with name `id_{key_type}`
4. key doesn't have to exist as a file, it could be smartcard

jath03: I am very aware that this isn't a perfect solution.  I was hoping to discuss some of these things.
I'm assume we are going to have to make the key file configurable?  The user might not even be using ssh.  I'll be looking into this more, but wanted to see if you or others had any ideas.

panekj: > I'm assume we are going to have to make the key file configurable?

There should be nothing configurable, it's up to git to handle everything

panekj: credential helper will handle whole authentication, there is no reason to handle auth any other way these days

jath03: I was under the impression that credential helper only dealt with password-based auth, not ssh.
https://docs.rs/git2/latest/git2/struct.Cred.html#method.credential_helper

panekj: It depends on the credential helper, but you can add ssh agent on credential helper failure

PR #2077, opened 2023-01-29
Fixed copy remote file url

jath03:
N/A


panekj:
See #2075

PR #2087, opened 2023-01-31
Enable plugins to act as LSP themselves

hbina:
As of now, plugins are merely there to spawn the actual LSP by way of `host/startLspServer` command and then it does _nothing_ because it never registered any capabilities and lapce itself never cared about the response from the init request. This changes that.

1. Use volt activation filters for the document filters.
2. Extract out the client capabilities struct to a function and pass it to the plugin as well.

This does require the plugin themselves to reply to the host init request like here: https://github.com/lapce/lapce-plugin-rust/pull/17

### DIFF
--- a/lapce-proxy/src/plugin/wasi.rs
+++ b/lapce-proxy/src/plugin/wasi.rs
@@ -22,17 +22,19 @@ use lapce_rpc::{
 };
 use lapce_xi_rope::{Rope, RopeDelta};
 use lsp_types::{
-    request::Initialize, ClientCapabilities, InitializeParams,
+    notification::Initialized, request::Initialize, DocumentFilter,
+    InitializeParams, InitializeResult, InitializedParams,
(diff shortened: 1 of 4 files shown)


lapce-proxy/src/plugin/wasi.rs
```diff
@@ -181,30 +183,37 @@ impl PluginServerHandler for Plugin {
...
-                    root_uri,
-                    capabilities: ClientCapabilities::default(),
-                    trace: None,
-                    client_info: None,
-                    locale: None,
-                    initialization_options: configurations,
-                    workspace_folders: None,
-                },
+        let root_uri = workspace.map(|p| Url::from_directory_path(p).unwrap());
+        if let Ok(value) = self.host.server_rpc.server_request(
```
hbina, line 189: I suppose this can be turned into a non-blocking request by spawning another thread and wrapping `server_capabilities` inside an `Arc<Mutex<T>>`

PR #2091, opened 2023-01-31
Fix compilation error because psp-types is updated

hbina:
Perhaps its time to lock to a specific rev?


panekj (changes requested):
it is locked

PR #2100, opened 2023-02-02
Fixed plugin not replying to initialization request to totally hang the plugin thread

hbina:
This will make plugins that don't respond to initialization request to not hang the plugin itself

### DIFF
--- a/lapce-proxy/src/plugin/psp.rs
+++ b/lapce-proxy/src/plugin/psp.rs
@@ -6,6 +6,7 @@ use std::{
         Arc,
     },
     thread,
+    time::Duration,
 };
 
 use anyhow::{anyhow, Result};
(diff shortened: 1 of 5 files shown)


dzhou121:
I think it's fine to require plugin only start to work after initialization. This is how LSP works so we should follow that too.

lapce-proxy/src/plugin/wasi.rs
```diff
@@ -183,37 +186,37 @@ impl PluginServerHandler for Plugin {
...
-            None,
-            false,
-        ) {
-            let result: InitializeResult = serde_json::from_value(value).unwrap();
-            self.host.server_capabilities = result.capabilities;
-            self.host.server_rpc.server_notification(
-                Initialized::METHOD,
-                InitializedParams {},
+        let server_capabilities = self.host.server_initialisation.clone();
+        thread::spawn(move || {
```
hbina, line 193: All the changes here are necessary to make offloading the init request here to a separate thread.

lapce-proxy/src/plugin/psp.rs
```diff
@@ -308,12 +310,15 @@ impl PluginServerRpcHandler {
             check,
             ResponseHandler::Chan(tx),
         );
-        rx.recv().unwrap_or_else(|_| {
-            Err(RpcError {
-                code: 0,
-                message: "io error".to_string(),
+        // TODO: add some kind of logging/visual cue for this timeout issues?
+        // TODO: customizable timeout durations
+        rx.recv_timeout(Duration::from_secs(10))
```
hbina, line 315: It does make the code unreasonable verbose. So perhaps having a timeout here is enough? Perhaps _every_ `recv` needs to be `recv_timeout`? @dzhou121

PR #2101, opened 2023-02-02
Revert "Properly instantiate the WASM blob and let it live inside the thread"

panekj:
Revert lapce/lapce#2090 since it breaks plugins

### DIFF
--- a/lapce-proxy/src/plugin/wasi.rs
+++ b/lapce-proxy/src/plugin/wasi.rs
@@ -449,14 +449,14 @@ pub fn start_volt(
         }
     })?;
     linker.module(&mut store, "", &module)?;
+    let handle_rpc = linker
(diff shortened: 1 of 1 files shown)


dzhou121:
@hbina Is this what "broke" plugins?

hbina:
I think this is what broke plugins https://github.com/lapce/lapce/blob/master/lapce-proxy/src/plugin/wasi.rs#L189-L216
Since all plugins so far never responded to the this request so the plugin thread simply hangs waiting for the respond (it worked previously because it was on a separate thread).
It was introduced here https://github.com/lapce/lapce/pull/2087
It can also be fixed by offloading the work to a separate thread again https://github.com/lapce/lapce/pull/2100
The minimum fix is to use `recv_timeout` instead of `recv` with a reasonable timeout like 5 seconds.

PR #2112, opened 2023-02-05
Fixed a deadlock when trying to initialize LSPs

hbina:
The deadlock was introduced in this PR here when the thread is removed. A better alternative to this PR: https://github.com/lapce/lapce/pull/2100 by using `server_request_async`. That other PR will still cause a deadlock, it just terminates the blocked request. This one should instead allow all requests to be served properly.


dzhou121:
I've changed it to go through the mainloop to update it without a lock. 

https://github.com/lapce/lapce/pull/2114

hbina:
Basically what I did except you use the `handle_rpc` to propagate the change. Not entirely familiar with the facilities.

Issue #2108, opened 2023-02-04
Reflect compilation results on project tree


nappa85:
A code edit can produce a lot of errors also outside current file, e.g. changing method parameters would break every method invocation.
At the moment, project tree doesn't reflect this state, e.g. writing file name in red on errors, yellow on warnings, etc...
I would be great also if folders would have the "highest raking" color, this way we can see errors also inside collapsed folders.
E.g. a folder containing a file with errors and a file with warnings should be red.
Also a folders should be recursive, e.g. a folder containing a red folder should be red itself.

PR #2126, opened 2023-02-09
Reflect diagnostic issues in the file explorer

hbina:
First implementation to fix issue here: https://github.com/lapce/lapce/issues/2108

~~To properly implement the feature with parent folders taking the worst severity, the painting algorithm would probably need a rewrite. It is currently painting the folders first then the files/folders inside it recursively.
So a solution might be to grab all the interested folders, precalculate how they should be drawn, THEN actually draw them.~~

Edit: I have updated the implementation so that the any issues in a file will be reflected in all of its ancestors. The ancestors will take on the worst of the issues as expected. Can't be arsed to figure out a better color tho. It looks horrible.


lapce-ui/src/explorer.rs
```diff
@@ -245,16 +259,14 @@ fn draw_name_input(
...
-        Naming::Renaming { .. } => {
-            name_edit_input.paint(ctx, data, env);
-        }
         Naming::Naming { .. } | Naming::Duplicating { .. } => {
-            name_edit_input.paint(ctx, data, env);
             // Skip forward by an entry
             // This is fine since we aren't using i as an index, but as an offset-multiple in painting
             *i += 1;
         }
+        _ => {}
```
dzhou121, line 269: I would change this to  ```Naming::Renaming { .. }```

PR #2127, opened 2023-02-09
Extended double-click options

Netroxen:
Extends the double-click option with the following options:
- None (single-click default)
- file (file-only double-click mode)
- all (file and folders double-click mode)

Changes:
- Field definition: `bool` -> `String`


lapce-ui/src/explorer.rs
```diff
@@ -687,52 +689,72 @@ impl Widget<LapceTabData> for FileExplorerFileList {
...
+                        double_click_mode = 1;
+                    }
+                    "all" => {
+                        double_click_mode = 2;
+                    }
+                    _ => {
+                        double_click_mode = 0;
+                    }
+                }
+
```
panekj, line 703: this should be enum type rather than int

Netroxen: Please see latest commit - is this what you mean?

hbina: No, `EditorConfig` will be directly deserialized from the file so you can instead automatically generate the conversion using

```rust
#[derive(Serialize, Deserialize)]
enum ClickModes {
    #[serde(rename = "single")]
    SingleClick,
    #[serde(rename = "file")]
    DoubleClickFile,
    #[serde(rename = "all")]
    DoubleClickAll,
}
```

and put the enum directly,

```rust
#[field_names(
        desc = "Use a double click to interact with the file explorer.\nOptions: file, all or Nothing (Default)."
    )]
    pub double_click: ClickModes,
```

and set the default to "single".

This removes the "magic constants" from plaguing the codebase and keep
There doesn't seem to be any error handling yet on the  `ReloadConfig` command (it simply fallback to the default config on failure).

Netroxen: Ahhh thanks for the information, as you can probably tell, I'm relatively new to Rust - so this really helps 💪

Ideally a drop-down menu would be even nicer, but unfortunately that was a little too complicated for me at this stage.

hbina: Yea creating the dropdown menu is quite tricky. Feel free to ping me if you have any more questions.

Netroxen: Alright, I have changed it as requested - please take a look.

Netroxen:
@panekj can be re-reviewed unless I missed something?

PR #2131, opened 2023-02-09
fix: dont overwrite symlink

### LINKED ISSUE #2122: Saving to a symlinked file overwrites the link
## Lapce Version

nightly-eaa1b57 (`lapce-nightly-bin` AUR package)

## System information

Arch Linux, x64

## Describe the bug

Saving a file that is a symlink to a different file overwrites the file and removes the symlink (and thus does not update the linked-to file).

Intended behaviour is to preserve/follow the link and update the original file.

## Additional information

- Created test files:
image

- Edited in Lapce:
image

- `bar` is no longer a link:
image

panekj:

Fixes #2122
Closes #2129


hbina:
I think the buffer hashmap should also be canonicalized.

PR #2160, opened 2023-02-16
Added a panel for stdio for both LSP/PSP

hbina:
1. It only seperate stdin and stdout right now regardless from which PSP/LSP it comes from.
2. Some performance consideration? How to make this better? Maybe allow user to configure the size of the buffer?
3. Some filters?
4. More seperation like request, response, notification, errors?


hbina:
Debugging with just the log is actually painful tho. This is still in draft stage, but I can imagine making queries like "what is the response for this request?" "Is this plugin still alive?". Another alternative is to improve the state of logging in Lapce by having something like structured logging which makes the log at least searchable with some grep and jq. But as I said, painful.

panekj:
Painful but that's what I already planned months ago and it doesn't limit itself to just LSP

hbina:
What did you plan?

Issue #1481, opened 2022-10-11
Ability to use Lapce from inside toolbox images


VarLad:
## Is your feature request related to a problem? Please describe.

I use Fedora SilverBlue and as such, my primary workflow involves making separate podman containers for separate stuff. 
Unfortunately, I have to reinstall Lapce's dependencies every time I make a new container. 

## Describe the solution you'd like

It would be nice if Lapce could display the containers and we could choose and enter into containers from the editor. This could be a feature built into Lapce or a plugin. This is a Linux specific feature.

JacobDeuchert:
The team from vscode published the Dev Containers as a open [Spec](https://containers.dev). If this feature is getting implemented at some point i think it would make sense to consider implementing the feature with the spec in mind. Will make it easier for teams to have the same configuration files for vscode and lapce.

PR #2165, opened 2023-02-18
feat: implement custom remote backend

panekj:


thomasaarholt:
Will this PR allow vscode-like "attach to running container" style connections. Those are used for connecting into docker containers. I see you're closing #1481 - are there more steps required to enable connecting to containers?

PR #2166, opened 2023-02-18
fix: build WSL backend only for Windows

panekj:


panekj:
image

PR #2167, opened 2023-02-18
ci: do not build lapce twice

panekj:
- changed `llvm-cov` to use `nextest`
- removed building lapce since it wastes time and resources as it's rebuilt again for coverage + tests
- removed whitespace


panekj:
image

panekj:
before change:
image

Issue #2213, opened 2023-03-05
Allow folding file content in global search results


hakro:
Hi !

It would be nice to be able to fold / unfold the results of global search by clicking on the file name.
This is very useful when there are many results to look at a time : 

![2023-03-05_23-06](https://user-images.githubusercontent.com/3624853/222988840-5027766c-b467-4e1e-bbc6-0c335ef5a1b9.png)

Cheers !

tk103331:
@hakro Good idea,  I implemented this feature

  see https://github.com/lapce/lapce/pull/2217

hakro:
@tk103331 , that's great, thanks a lot.
I'll give it a test today and let you know.

Cheers!

PR #2217, opened 2023-03-07
feat: allow folding file content in global search results

tk103331:
Allow folding file content in global search results
image

image

Issue: https://github.com/lapce/lapce/issues/2213


hakro:
I just tested this, it's looking great !
Thank you

Issue #2212, opened 2023-03-05
Highlight all occurrences of the selected variable / text in the current document


hakro:
It would be really useful to have all occurrences of the selected text highlighted.

![2023-03-05_23-01](https://user-images.githubusercontent.com/3624853/222988464-22779fce-0547-48bd-aff1-6432dc4c8b01.png)

This helps spot where the variable is used in the document.

Thank you !

tk103331:
@hakro  hello hakro，I just implemented this feature.

see https://github.com/lapce/lapce/pull/2222

hakro:
It's Christmas everyday. Thank you very much :)
I'll give it a test today.

PR #2222, opened 2023-03-09
feat: highlight all occurrences of the selected text

tk103331:
feature: highlight all occurrences of the selected text

Fixes https://github.com/lapce/lapce/issues/2212

image

There is also a setting to disable this feature：
image

### DIFF
--- a/lapce-ui/src/editor.rs
+++ b/lapce-ui/src/editor.rs
@@ -1532,11 +1532,8 @@ impl LapceEditor {
             let end_line = *screen_lines.lines.last().unwrap();
             let start = data.doc.buffer().offset_of_line(start_line);
             let end = data.doc.buffer().offset_of_line(end_line + 1);
-            data.doc.update_selection_find(
-                &selection,
-                start_line,
-                end_line,
-            );
+            data.doc
+                .update_selection_find(&selection, start_line, end_line);
             for region in data
                 .doc
                 .selection_find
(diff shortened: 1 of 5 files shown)


lapce-ui/src/editor.rs
```diff
@@ -1484,6 +1485,128 @@ impl LapceEditor {
...
+        if !selection.is_empty() {
+            let cursor_offset = data.editor.cursor.offset();
+            let start_line = *screen_lines.lines.first().unwrap();
+            let end_line = *screen_lines.lines.last().unwrap();
+            let start = data.doc.buffer().offset_of_line(start_line);
+            let end = data.doc.buffer().offset_of_line(end_line + 1);
+            data.doc.update_selection_find(
+                &selection,
+                0,
+                data.doc.buffer().last_line(),
```
hbina, line 1538: I think this might cause a lag in a very large file. Why not search only the lines within `screen_lines`?

hbina:
I think it should not try to find whitespaces?

lapce-ui/src/editor.rs
tk103331: Here is the reference 'find' logic (I know that it is a full document search for counting).
I tried to search in "screen_lines", but I didn't find a way to repaint. Can you give me some suggestions?

tk103331: Sorry, I made a mistake. When "screen_lines" changes, because the selection does not change, the search logic is not executed. 
But at the same time, I want to consider the situation that "screen_lines" and "selection" are unchanged, and I should not execute the search logic.
I'm looking for a good way to solve this problem

hakro:
I just tested this. Looking great !
Thx

PR #2228, opened 2023-03-11
Add StartupWMClass to desktop entry

panekj:


Toromambo (commented):
thx for the tip!

PR #2245, opened 2023-03-14
feat: support for selecting WSL distribution

tk103331:
Feature: Support for selecting WSL distribution

image

Open Folder label

image

Openned Folder label
image

Recent Folder label
image


MinusGix:
This PR is for the pre-Floem version of Lapce, which we recently changed `master` to, it would be nice if it could be rewritten.  
I don't think the changes would be all that different in `lapce-app`.

tk103331:
Close due to new UI

PR #2246, opened 2023-03-14
Remove duplicate binding for Shift-Delete

lnicola:

#2071 added this binding, but it's redundant with the existing one (for cut). Worse, the result is that two lines are removed instead of a single one.

I think it's better to cut instead of deleting because that's what every editor I've seen does. And `Cut` without a selection, already cuts the whole line, as it does everywhere else.

### DIFF
--- a/defaults/keymaps-nonmacos.toml
+++ b/defaults/keymaps-nonmacos.toml
@@ -112,11 +112,6 @@ command = "delete_word_forward"
 mode = "i"
 
 [[keymaps]]
-key = "shift+delete"
-command = "delete_line"
-mode = "i"
-
-[[keymaps]]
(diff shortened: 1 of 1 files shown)


lnicola:
CC @phw

phw:
> And `Cut` without a selection, already cuts the whole line, as it does everywhere else.

"Cut" and "Delete lines" are not identical. "Cut" cuts the selection, whereas delete lines does delete all selected lines (also partially selected). But as the `delete_line` command you added handles this all is good for me, don't need a default key binding.

And the `delete_line` action works well for me, thanks a lot for this :) I usually have it bound to `Ctrl+Shift+K` .

PR #2115, opened 2023-02-07
ci: add aarch64 (arm64) builds

panekj:
Fixes #2110

PR #2256, opened 2023-03-18
vendor dependencies and add artifacts checksum

### LINKED ISSUE #2110: [Feature]: Please provide downloads for Linux aarch64
Please build for aarch64-unknown-linux-gnu triplet in future GitHub CI and releases. RISC architectures are getting more common these days.

Although I don't use arm (I use 64 bit aarch64), it would be cool to add it too because others may have, say, a Raspberry Pi.

### LINKED ISSUE #2247: Please provide vendor tarballs for a release
tl;dr: Modern languages make packagers problems. There is an easy solution, look below!

Long version:
Most package systems are not build to fit modern statically linked languages that have their own package managers that manage dependencies. Most of them come from an era where developers released tarballs an their own website. A packager would write a package that fetches this tarball, set some options and compiled it. 
With the modern languages package managers adapted by adding possibilities to fetch those dependencies from the internet, doing Git checkouts instead of downloading a tarball etc. 
However, how those languages are handled in package managers is still open to debate. 
For example there are package managers that just fetch the source and run `cargo build`, some package each dependency independently and manage everything their own. For Rust packages on Gentoo we decided that we utilize Cargo, but we do not allow cargo to fetch anything from the internet during build.
We therefore have a variable with a list of crates. All the crates on that list are fetched and put into the right place, then we patch the cargo config of the package and only then invoke cargo build (with --offline)

This mechanism worked nice for a while until we ran into problems with the same mechanism we use with packages written in Go-Lang. For some packages that list (which is just a space-separated string) was getting longer than what the Linux Kernel would allow for a variable. (https://bugs.gentoo.org/721088) Also for every artifact that is downloaded during build we check the checksum. The checksum is saved during packaging. Those lists with checksums also were getting too big for Go-Lang packages. 
Therefore the deprecated that mechanism for Go-Lang packages. On Rust it is not that bad, and we kept it for now, but for packagers it is still quite much work to figure out the dependency crates and especially with crates from Github that lapce uses heavily. Because we need to get the commit hash of every crate from Github that lapce uses every time a new version of lapce is released. 

I would like to ask you to release a vendor tarball alongside ever release. A vendor tarball is a tar archive that contains all of the dependency crates. 
It is easily generated by running `cargo vendor` in the repository and then tarring that created folder with `tar -czvf vendor.tar.gz vendor`.
The resulting tarball could be released as a release artifact. 

If you are using GitHub actions to manage your releases, this could be easily done by running a few extra jobs of checking out the repo, running cargo vendor and tarring that vendor directory. 

This is just a few extra lines in a GitHub job for you, but it would help me tremendously to package your software and make it easily available for Gentoo users.

panekj:

---

- [ ] Added cross-compilation for aarch64
- [ ] Added checksums to releases
- [ ] Added vendor tarballs to releases
- [ ] Added Debian package
- [ ] Added RPM package

---

Closes #2115
Fixes #2110
Fixes #2247


coolcoder613eb:
+1

coolcoder613eb:
Why did you close it?

panekj:
> Why did you close it?

Because it's outdated and it's not going to be implemented

coolcoder613eb:
I want the aarch64 cross-builds :(

Issue #2235, opened 2023-03-13
Create Dev Container to have an stable contributing environment


FarzinVatani:
## Is your feature request related to a problem? Please describe.

New contributors need to install bunch of libraries and maybe they develop on windows and it may be a little hard to work around.

## Describe the solution you'd like

So a universal development environment like docker development container is a good idea to start developing fast.

panekj:
Lapce is a graphical application so running it in container is not supported, it also puts maintainence burden on us to keep devcontainer up-to-date.

nikaro:
I think the idea of this issue was more to support something like [VSCode DevContainer](https://containers.dev). This is a really nice things as it allows to declaratively configure your development environment and share it easily. No need to install LSP servers, linters, compilers, and whatever on your machine. It installed/started when you start your IDE.

PR #2261, opened 2023-03-20
Development environment for lapce using devcontainer

akhildevelops:
I felt it would be great to have a devcontainer easing the pain for lot many contributors to setup environment.

Few of the them even mentioned for the need of it.https://github.com/lapce/lapce/issues/2235

Here's the devcontainer.json file that can be used to setup development environment in docker and build lapce.

Running the binary creates a seperate gui window that uses host's X11 server.

Reference: http://fabiorehm.com/blog/2014/09/11/running-gui-apps-with-docker/

image

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,32 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
+// README at: https://github.com/devcontainers/templates/tree/main/src/ubuntu
+{
+	"name": "Ubuntu",
+	// Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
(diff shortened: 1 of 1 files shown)


panekj:
I would like to not depend on specific host configuration and hacks

akhildevelops:
What do you mean by hacks ?

panekj:
It requires X11?

akhildevelops:
Yup

panekj:
That's a hack, that's a very specific host dependency

akhildevelops:
I believe it also works on wayland with few changes, lemme check and come back.

akhildevelops:
For further reference: https://github.com/mviereck/x11docker/wiki/How-to-provide-Wayland-socket-to-docker-container

akhildevelops:
Closing this as it cannot be built for majority of display servers

PR #2262, opened 2023-03-20
Upgraded dependencies

ThePseudo:
I tried to update some of the dependencies (the ones that do not create any problem), as some may contain bugfixes and so on.

Thank you


panekj:
Thanks but we have dependabot to do that and you introduce major changes (e.g.: clap 3 -> 4)
We keep track of most dependencies that cause issues so it's not necessary to update all of them

Issue #2267, opened 2023-03-21
Folding search results does not work correctly in some cases


hakro:
## Lapce Version

0.2.7

## System information

PopOS 22

## Describe the bug
Under some circumstances, folding the global search results does not work as expected.

Here is a quick video to show that : 

https://user-images.githubusercontent.com/3624853/226652509-70a51f41-f3bc-4f31-b376-2fc8268f4506.mp4

Basically, when I try to to fold the 4 search result (render_canvas_cull.h)
But when I click on the filename, instead of folding it, it unfolds the file above.

Also, when I want to click on the search result of the 3 file, it folds the result instead of opening the file. This happens whether or not the file is already opened in the editor.

Most of the time, it works fine thought.
Folding the search results was added recently here : https://github.com/lapce/lapce/pull/2217

Cheers

PR #2272, opened 2023-03-22
fix: search result folding incorrectly

tk103331:
Fix https://github.com/lapce/lapce/issues/2267

### DIFF
--- a/lapce-ui/src/search.rs
+++ b/lapce-ui/src/search.rs
@@ -333,10 +333,6 @@ impl SearchContent {
 
         let mut i = 0;
         for (path, matches) in data.search.matches.iter() {
-            if matches.len() + 1 + i < n {
-                i += matches.len() + 1;
-                continue;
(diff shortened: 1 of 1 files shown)


tk103331:
https://user-images.githubusercontent.com/4404609/227404888-ea11ba31-2557-486d-912e-490827171f68.mp4

hbina:
I can reproduce it here

https://user-images.githubusercontent.com/6733660/227407596-63a9d90b-b39f-47d4-9ff8-38949ef57752.mp4

tk103331:
@hakro @hbina 

I found the mistake. The logic of processing folding in paint and mouse_down is inconsistent, so the item found by clicking the position is not the actual displayed item.
This bug may be triggered when using the scrollbar to scroll to some positions.

I've fixed the logic issues in paint and also done some tests. But maybe some more testing is needed.

Issue #2268, opened 2023-03-21
Allow completely hiding the "Open Editors" Panel


hakro:
Hey !

It would be great to be able to show/hide this panel via a configuration flag : 

![2023-03-21_16-27](https://user-images.githubusercontent.com/3624853/226655979-4d31f9e1-1406-4f9a-91e1-e6c9ad5cdd5b.png)

IMHO, its usefulness is quite limited, given that the files are already shown in the tabs over the editors. And in my case, it's an annoyance to have, even if it can be folded. It still takes some space, and sometimes opens automatically for some reason.

I think the idea to have it comes form VSCode/ium. But even there, it's possible to hide any of the panel on the left side, by clicking on the 3 dots : 

![2023-03-21_16-31](https://user-images.githubusercontent.com/3624853/226656865-4cfba7d5-2953-4333-ac81-b99d254aa7be.png)

Cheers !

PR #2273, opened 2023-03-22
allow hide open editors

tk103331:
feat: https://github.com/lapce/lapce/issues/2268

image

image

### DIFF
--- a/defaults/settings.toml
+++ b/defaults/settings.toml
@@ -13,6 +13,7 @@ code-lens-font-size = 2
 line-height = 1.5
 tab-width = 4
 show-tab = true
+hide-open-editors = false
 show-bread-crumbs = true
 scroll-beyond-last-line = true
 cursor-surrounding-lines = 1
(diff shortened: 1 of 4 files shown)


hbina:
This requires a restart to take effect right?

tk103331:
restart or reopen workspace

hakro:
Thank you so much @tk103331 
Just a few nitpicks that come to mind.
- It would probably be better to consistently keep a positive configuration action, rather than a negative one. Meaning, have a config that says "Show open editors" that defaults to true, instead of "Hide open editors" that defaults to false. This would be a good thing to actually have all across the configurations.
- Is there a way to avoid having to restart the editor or the workspace? 

Cheers !

hakro:
It's probably better to mention that a restart is required somewhere then.
Either on a small toast (that can even be reused by other configs that require restarting), or directly as a `(Requires restarting Lapce)` mention next to each config.

Cheers

tk103331:
Close due to new UI

PR #2280, opened 2023-03-22
Scala Treesitter Support

gautamprikshit1:
- [ ] Added scala treesitter support


panekj:
Closing since we can't add more grammars for now

PR #2281, opened 2023-03-22
editor: highlight multiple selections and words on cursor

azizk:
Hi!

I reworked the `paint_selection_find` function so that:
- Multiple cursor regions are taken into account.
- `data.doc.selection_find` is used directly because a `FindProgress` isn't really necessary for an implicit search like this which is also limited to the visible lines.
- The "active" selection under the cursor isn't filled with a color as it's distracting and makes only sense with a `Ctrl+F` search.
- The phantom text (to the right) isn't included in the rectangular highlight (also fixed for `Ctrl+F`).
- Selecting backwards is working (`start` and `end` are swapped).
- Renamed variables for readability.

Notes/Issues:
- I'm still learning Rust, so there may be beginner mistakes.
- The code for getting multiple regions and iterating over them seems correct to me, but it doesn't actually highlight all selections. Only those of the first cursor. Not experienced enough to debug this.
- I left the curly braces block in there after removing the if clause, so that the diff remains readable.

### DIFF
--- a/lapce-core/src/cursor.rs
+++ b/lapce-core/src/cursor.rs
@@ -276,8 +276,8 @@ impl Cursor {
         RegisterData { content, mode }
     }
 
-    /// Return the current selection start and end position for a
-    /// Single cursor selection
+    /// Returns the current selection start and end position for a
+    /// single cursor selection.
     pub fn get_selection(&self) -> Option<(usize, usize)> {
         match &self.mode {
             CursorMode::Visual {
@@ -285,11 +285,27 @@ impl Cursor {
                 end,
                 mode: _,
             } => Some((*start, *end)),
+
             CursorMode::Insert(selection) => selection
                 .regions()
                 .first()
                 .map(|region| (region.start, region.end)),
+
             _ => None,
+        }
+    }
+
+    /// Returns the [`Selection`] of the current mode.
+    pub fn as_selection(&self, buffer: &Buffer) -> Selection {
+        match &self.mode {
+            CursorMode::Normal(offset) => Selection::caret(*offset),
+
+            CursorMode::Visual { start, end, .. } => Selection::region(
+                (*start).min(*end),
+                buffer.next_grapheme_offset((*start).max(*end), 1, buffer.len()),
+            ),
+
+            CursorMode::Insert(selection) => selection.clone(),
         }
     }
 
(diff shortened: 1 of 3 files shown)


azizk:
More fixes and improvements:

- Completely removed unneeded `update_selection_find` function.
- Call `paint_selection_find` after `highlight_scope_and_brackets` so the highlight rectangle is drawn over the vertical scope lines when overlapping.
imagen
- No need to call `.occurrences().regions_in_range()` as the `selection_find.update_find()` function is already called with the first and last screen line offsets.
- Remember the search strings to search for and the `match_whole_words` setting for each string.
- Actually take the `case_matching` setting from the finder box widget.
- No need to `for line in &screen_lines.lines` since we know the range and can iterate over `start_line..=end_line` instead.
- Draw the highlight rectangle with rounded corners.
- Refactored code into a more functional style.

azizk:
TODO: I saw that the view is painted frequently and the search is run each time, so maybe we should cache the selection search results so that redraws are as fast as possible?

azizk:
- Moved code for getting the selection ranges to `as_selection()` in `impl Cursor`.
- No need to call `selection_find.update_find()` with `include_slop: true`. Instead calculate the search range from the longest selected string to search for.

azizk:
I'm not sure but I think I found a bug in `Find::update_find()`. When you search for an Umlaut with `whole_words = true` the app crashes:

```rust
selection_find.search_string = Some("ä".to_string());
selection_find.whole_words = true;
selection_find.update_find(buffer.text(), start, end, false);
```

Snippet from logs:

```
[2023-03-27][02:26:59][panic][ERROR] thread 'main' panicked at 'byte index 904 is not a char boundary; it is inside 'ä' (bytes 903..905) of `...ä...`[...]': /home/aziz/.cargo/registry/src/github.com-1ecc6299db9ec823/lapce-xi-rope-0.3.1/src/rope.rs:663
   0: <backtrace::capture::Backtrace as core::default::Default>::default
             at /home/aziz/.cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.66/src/capture.rs:410:9
```

azizk:
I really had to implement these continuous multi-line highlights:

imagen

:grin:

Cool thing is you'll be able to use the same functions for normal `Ctrl+F` searches!

I'd love to do one final thing: render the shape with rounded the corners. I think it will look more pleasant that way.

azizk:
This has been sitting around for too long. The code can't be merged anymore.

dzhou121:
Sorry for not reviewing it earlier.

azizk:
No worries. Maybe you can salvage one or two useful things. :+1:

PR #2287, opened 2023-03-23
Cargo: added --wait flag and explaining note

azizk:
It took me a while to figure out why I didn't get any console output when starting lapce. A note like this would've been helpful. :slightly_smiling_face:


panekj:
It doesn't say anything about starting development

panekj:
I believe this is a non-issue and that such note isn't useful especially since it's hidden in Cargo manifest. Tt will be better to describe default behaviour of lapce in docs at some point

PR #2305, opened 2023-03-29
feat:add plugin settings folding icon

tk103331:
feature: add plugin settings folding icon

image

image


tk103331:
Close due to new UI

PR #2315, opened 2023-04-01
fix: `ce` motion respects whitespace

### LINKED ISSUE #1848: vim `ce` implementation is incorrect and consumes whitespace after the end of the word
## Lapce Version

v0.2.5

## System information

windows 10

## Describe the bug

If the cursor is at the start of a line `foo bar`, and the user presses `ce`

### Expected outcome

The string `foo` is deleted, the user is in edit mode with the line containing ` bar`.

### Actual outcome

`foo ` is deleted the line contains only `bar`.

## Additional information

The analogous vim delete command, `de`, respects whitespace, so probably could look at that for inspiration for how to fix `ce`

jvhoven:

The `de` motion respects whitespace as this is compromised of two different keymaps: `d` for `DeleteWord` and `e` for `WordEndForward`, where the `WordEndForward` implementation accounts for whitespace. 

I've introduced a new EditCommand called `DeleteWordEndAndInsert` and slightly tweaked the current implementation to alter the selection logic when `DeleteWordEndAndInsert` is used.

Fixes #1848

### DIFF
--- a/lapce-core/src/editor.rs
+++ b/lapce-core/src/editor.rs
@@ -1353,13 +1353,18 @@ impl Editor {
                 cursor.mode = CursorMode::Insert(selection);
                 vec![(delta, inval_lines, edits)]
             }
-            DeleteWordAndInsert => {
+            DeleteWordAndInsert | DeleteWordEndAndInsert => {
                 let selection = {
                     let mut new_selection = Selection::new();
                     let selection = cursor.edit_selection(buffer);
 
                     for region in selection.regions() {
-                        let end = buffer.move_word_forward(region.end);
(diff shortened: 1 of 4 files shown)


lapce-core/src/editor.rs
```diff
@@ -1353,13 +1353,18 @@ impl Editor {
                 cursor.mode = CursorMode::Insert(selection);
                 vec![(delta, inval_lines, edits)]
             }
-            DeleteWordAndInsert => {
+            DeleteWordAndInsert | DeleteWordEndAndInsert => {
```
jvhoven, line 1356: This is the nicest way I could find to avoid code duplication, hopefully this is alright

PR #2344, opened 2023-04-15
feat(editor): bracket colorization

bbrabbitt:
This PR implements a basic bracket highlighter by building an AST of brackets and highlighting them accordingly. I originally want to follow [vscode's way](https://code.visualstudio.com/blogs/2021/09/29/bracket-pair-colorization) to build it but I don't have enough time to optimize it with a 2-3 tree. Nonetheless Rust is fast enough to parse most files in realtime in my test. This is still WIP as there are some minor problems to solve.

Screenshot 2023-04-15 at 3 15 45 AM

known issues:
- [x] first character of the file is incorrectly highlighted
- [x] some files still have unhighlighted brackets
- [x] imbalanced parentheses may cause following brackets highlight incorrectly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1492,7 +1492,7 @@ checksum = "98de4bbd547a563b716d8dfa9aad1cb19bfab00f4fa09a6a4ed21dbcf44ce9c4"
 [[package]]
 name = "floem"
 version = "0.1.0"
-source = "git+https://github.com/lapce/floem?rev=c02c55fe730c5333be0533970274b4380e930025#c02c55fe730c5333be0533970274b4380e930025"
+source = "git+https://github.com/lapce/floem?rev=7981a0ad810ee3b74406391aa7066228e022c059#7981a0ad810ee3b74406391aa7066228e022c059"
 dependencies = [
  "bitflags 2.4.0",
  "crossbeam-channel",
@@ -1519,12 +1519,12 @@ dependencies = [
 [[package]]
 name = "floem_reactive"
 version = "0.1.0"
-source = "git+https://github.com/lapce/floem?rev=c02c55fe730c5333be0533970274b4380e930025#c02c55fe730c5333be0533970274b4380e930025"
+source = "git+https://github.com/lapce/floem?rev=7981a0ad810ee3b74406391aa7066228e022c059#7981a0ad810ee3b74406391aa7066228e022c059"
 
 [[package]]
 name = "floem_renderer"
 version = "0.1.0"
-source = "git+https://github.com/lapce/floem?rev=c02c55fe730c5333be0533970274b4380e930025#c02c55fe730c5333be0533970274b4380e930025"
+source = "git+https://github.com/lapce/floem?rev=7981a0ad810ee3b74406391aa7066228e022c059#7981a0ad810ee3b74406391aa7066228e022c059"
 dependencies = [
  "cosmic-text",
  "peniko",
@@ -1534,7 +1534,7 @@ dependencies = [
 [[package]]
 name = "floem_vger"
 version = "0.1.0"
-source = "git+https://github.com/lapce/floem?rev=c02c55fe730c5333be0533970274b4380e930025#c02c55fe730c5333be0533970274b4380e930025"
+source = "git+https://github.com/lapce/floem?rev=7981a0ad810ee3b74406391aa7066228e022c059#7981a0ad810ee3b74406391aa7066228e022c059"
 dependencies = [
  "anyhow",
  "floem_renderer",
--- a/lapce-app/Cargo.toml
+++ b/lapce-app/Cargo.toml
@@ -51,7 +51,7 @@ bytemuck = "1.8.0"
 tokio = { version = "1.21", features = ["full"] }
 futures = "0.3.26"
 clipboard = "0.5.0"
-floem = { git = "https://github.com/lapce/floem", rev = "c02c55fe730c5333be0533970274b4380e930025" }
+floem = { git = "https://github.com/lapce/floem", rev = "7981a0ad810ee3b74406391aa7066228e022c059" }
 # floem = { path = "../../workspaces/floem" }
 config = { version = "0.13.2", default-features = false, features = ["toml"] }
 structdesc = { git = "https://github.com/lapce/structdesc" }
--- a/lapce-app/src/about.rs
+++ b/lapce-app/src/about.rs
@@ -99,64 +99,61 @@ pub fn about_popup(window_tab_data: Rc<WindowTabData>) -> impl View {
     let logo_size = 100.0;
 
     exclusive_popup(window_tab_data, about_data.visible, move || {
-        stack(move || {
-            (
-                svg(move || (*config.get()).logo_svg()).style(move |s| {
-                    s.size_px(logo_size, logo_size).color(
-                        *config.get().get_color(LapceColor::EDITOR_FOREGROUND),
-                    )
-                }),
-                label(|| "Lapce".to_string()).style(move |s| {
-                    s.font_bold().margin_top_px(10.0).color(
-                        *config.get().get_color(LapceColor::EDITOR_FOREGROUND),
-                    )
-                }),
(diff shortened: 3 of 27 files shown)


dzhou121:
Can I suggest 2 things:

1) You can query brackets from tree-sitter which is parsed for the file anyway. You can have a look at the code for highlighting the current bracket scope. 

2) Only parse the file content that is in view for the ones without tree-sitter grammar.

bbrabbitt:
> Thanks for this work. It's excellent. Can I suggest 2 things:
> 
> 
> 
> 1) You can query brackets from tree-sitter which is parsed for the file anyway. You can have a look at the code for highlighting the current bracket scope. 
> 
> 
> 
> 2) Only parse the file content that is in view for the ones without tree-sitter grammar. 

Thanks for the advice, I was also thinking about tree sitter when I wrote the parser but I'm not very familiar with tree sitter. I'll look at the bracket scope code.

lapce-data/src/document.rs
```diff
@@ -49,6 +49,240 @@ use lsp_types::{
...
+    LeftCurly,
+    RightCurly,
+    Pair,
+    Code,
+    Dummy,
+}
+
+#[derive(Clone, Debug)]
+enum BracketParserMode {
+    Paring,
```
TerminalFi, line 67: Typo ?

bbrabbitt:
I add tree sitter parser as primary parser for bracket colorization and use my parser as a backup. I also add four theme colors and an option to turn on/off the bracket colorization. This PR is ready for review and merge.

defaults/settings.toml
```diff
@@ -49,6 +49,7 @@ double-click = "single"
 move-focus-while-search = true
 diff-context-lines=3
 scroll-speed-modifier=1
+colorize-brackets = true
```
panekj, line 52:
```suggestion
brackets-pair-colorization = true
```

lapce-data/src/document.rs
```diff
@@ -2217,6 +2236,40 @@ impl Document {
...
+                                config.get_color_unchecked(LapceTheme::TERMINAL_BLUE)
+                            }
+                            "bracket.color2" => config
+                                .get_color_unchecked(LapceTheme::TERMINAL_YELLOW),
+                            "bracket.color3" => config
+                                .get_color_unchecked(LapceTheme::TERMINAL_MAGENTA),
+                            "bracket.unpaired" => {
+                                config.get_color_unchecked(LapceTheme::TERMINAL_RED)
+                            }
+                            _ => unreachable!(),
```
panekj, line 2261: this doesn't seem right, `TERMINAL` shouldn't be used in document

bbrabbitt: I know it looks weird, but I can't find a good way to fallback to theme's basic color if the theme doesn't set bracket colorization colors. The basic color seems not loaded to the config.

bbrabbitt: Fixed

panekj: What I meant is that this shouldn't be matched by string, it should it's own theme key (since you added them to default configs anyway)

bbrabbitt: Well the problem is if I download a theme without these keys, then there will be no color for brackets since `get_style_color` will not fallback to default theme.

panekj: That is fine

bbrabbitt: Fine, I removed it.

lapce-core/src/syntax/mod.rs
```diff
@@ -68,6 +69,300 @@ pub enum Error {
...
+                    self.bracket_pos = bracket_pos;
+                }
+            } else {
+                self.code = code.chars().collect();
+                self.cur = 0;
+                self.parse();
+                let mut pos_vec = vec![];
+                Self::highlight_pos(
+                    &self.ast,
+                    &mut pos_vec,
```
panekj, line 200: Why pass a vec instead of just returning it?

lapce-data/src/document.rs
```diff
@@ -470,6 +471,11 @@ impl Document {
         };
         let mut selection_find = Find::new(0);
         selection_find.case_matching = CaseMatching::Exact;
+        let config = LapceConfig::load(&LapceWorkspace::default(), &[]);
+        let parser = BracketParser::new(
+            "".to_string(),
+            config.editor.bracket_pair_colorization,
+        );
```
panekj, line 478: This is wrong. It should load config from calling function

lapce-core/src/syntax/mod.rs
bbrabbitt: Since this is a recursive parser. Passing a vec can save my time to fight against the borrow checker. Trust me I tried to return when I wrote another similar parser and it was a nightmare.

defaults/dark-theme.toml
```diff
@@ -61,6 +61,11 @@ yellow = "#E5C07B"
 "variable.other.member" = "$green"
 "tag" = "$blue"
 
+"bracket.color.1" = "$blue"
+"bracket.color.2" = "$yellow"
+"bracket.color.3" = "$purple"
+"bracket.unpaired" = "$red"
+
```
panekj, line 68: This is still missing from config file

defaults/light-theme.toml
```diff
@@ -67,6 +67,11 @@ yellow = "#C18401"
 "variable.other.member" = "$green"
 "tag" = "$blue"
 
+"bracket.color.1" = "$blue"
+"bracket.color.2" = "$yellow"
+"bracket.color.3" = "$purple"
+"bracket.unpaired" = "$red"
+
```
panekj, line 74: This is still missing from config file

lapce-data/src/document.rs
bbrabbitt: You mean pass from the new function of Document?

defaults/dark-theme.toml
bbrabbitt: The parser lives in core so it just returns String like tree-sitter. There is no need to access via LapceTheme config.

panekj: That doesn't matter though? It still has to be part of the config

bbrabbitt: Do you mean lapce-data/config.rs? Other syntax colors are also not included though?

Seven0492:
Sorry for the ping, but is this ready to be merged yet? Because it seems like it has stalled.

I'd love for this to be included by default.

lapce-data/src/config.rs
```diff
@@ -454,7 +454,7 @@ pub struct EditorConfig {
     #[field_names(
         desc = "Set the cursor blink interval (in milliseconds). Set to 0 to completely disable."
     )]
-    pub blink_interval: u64, // TODO: change to u128 when upgrading config-rs to >0.11
```
Seven0492, line 457: Why remove the comment? TODO's are important, so much so they're highlighted by default on most terminal text/code editors.

panekj: This comment is rightfully removed

defaults/dark-theme.toml
panekj: Now that I'm looking at it, it's correct, my bad

Nyabinary:
Would this work with the new Floem based Frontend?

panekj:
> Would this work with the new Floem based Frontend?

You can just look at conflicts for resolution

bbrabbitt:
> Sorry for the ping, but is this ready to be merged yet? Because it seems like it has stalled.
> 
> It's two months old now.
> 
> I'd love for this to be included by default.

I was busy in the last few months. Now the biggest blocker for me is to find a better way to get the current state of the colorization switch. Currently I load the config again using `LapceConfig::load` in document.rs ,which is not the best solution but I haven't come up a more idiomatic way to do this.

Nyabinary:
Couldnt fix the issue I assume?

bbrabbitt:
> Couldnt fix the issue I assume?

No, I just made a mistake in merging all the changes.

PR #2366, opened 2023-05-12
Always solve relative paths (not symlinks) in cli arguments

akazakarin96:
This will fix an issue where using `.` will cause lapce-proxy to crash because it is trying to create a Url like `./Cargo.toml` which is wrong.


akazakarin96:
Superseded by https://github.com/lapce/lapce/pull/2367

PR #2388, opened 2023-05-21
Add option to hide files from .gitignore (Fixes #811)

### LINKED ISSUE #811: [feature] blacklist for file explorer tab
In VS Code there's a way to blacklist certain directories/files i.e `.git`, `target`, etc from showing up in the file explorer tab and it would be very helpful if that was added to lapce
Either as a global config or as a workspace specific on
If it was per-workspace, I think being able to automatically generate the blacklist from a .gitignore file would also be nice (but not necessary)

MatthewScholefield:
This is a draft PR (unfinished) to an option to hide files based on a `.gitignore` at the project root (Fixes #811).

Remaining work:
 - [ ] Trigger update when config value changed
    - Right now when the config value is changed, it doesn't update the file explorer. I need to find out how to subscribe to this change and trigger an update. Any pointers are welcome

Possible future work:
 - Support nested `.gitignore`  files
 - Support custom ignore patterns

### DIFF
--- a/defaults/settings.toml
+++ b/defaults/settings.toml
@@ -29,6 +29,7 @@ highlight-selection-occurrences = true
 highlight-scope-lines = false
 autosave-interval = 0
 format-on-autosave = true
+hide-gitignored-files = false
 enable-inlay-hints = true
 inlay-hint-font-family = ""
 inlay-hint-font-size = 0
--- a/lapce-data/Cargo.toml
+++ b/lapce-data/Cargo.toml
@@ -55,6 +55,7 @@ bytemuck = "1.8.0"
 # For parsing markdown data, such as in hovers
 pulldown-cmark = "0.9.1"
 url = "2.3.1"
+ignore = "0.4.20"
 
 [target.'cfg(target_os="macos")'.dependencies]
 dmg = "0.1.1"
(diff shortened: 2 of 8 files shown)


lapce-data/Cargo.toml
```diff
@@ -55,6 +55,7 @@ bytemuck = "1.8.0"
 # For parsing markdown data, such as in hovers
 pulldown-cmark = "0.9.1"
 url = "2.3.1"
+ignore = "0.4.20"
```
panekj, line 58: this should be same version as in `lapce-proxy`

defaults/settings.toml
```diff
@@ -29,6 +29,7 @@ highlight-selection-occurrences = true
 highlight-scope-lines = false
 autosave-interval = 0
 format-on-autosave = true
+hide-gitignored-files = false
```
panekj, line 32:
```suggestion
exclude-gitignore = false
```

lapce-data/src/explorer.rs
```diff
@@ -107,6 +111,22 @@ impl FileExplorerData {
...
+                    if let Some(e) = err {
+                        log::warn!("Problem parsing gitignore: {}", e);
+                    }
+                    Some(gitignore)
+                } else {
+                    None
+                }
+            } else {
+                None
+            },
```
panekj, line 129: it would be better to take out that if before `Self`

PR #2407, opened 2023-05-27
tree-sitter: Add the .slint UI language

flukejones:


MinusGix:
This needs to add itself to `lapce-app/Cargo.toml`, similar to how it was done in `lapce-ui/Cargo.toml`. Also the changelog shouldn't be in the 0.2.8 section since it wasn't sent out in 0.2.8, it should go in unreleased.

panekj:
Closing since we can't add more grammars for now

PR #2412, opened 2023-05-28
Move text layout logic out of Document

MinusGix:

This PR breaks the text layout logic and movement logic out of the `Document` structure.  
Text layouts and functions related to points / offsets are moved to the `EditorViewData` structure that is put within `EditorData`.  
This is necessary for some linewrapping configurations, such as when the editor should wrap at the editor width and thus there should be separate lines. As well, having the views like this would allow other alternative representations of the same document more easily (git views?)  
  
This also moves the movement logic in `Document` out into its own file and simplifies what pieces of logic are actually needed. This will make it easier to test.  
Various `Buffer` functions had their implementations moved to `RopeText`, since they did not depend on the details of `Buffer`.

------

~~I started on this before the change to add `EditorContentView`. I like that structure, but I don't think I can just put the text layout cache into `EditorContentView` as it needs to be accessible from anything that needs the functions using it.   
Having `EditorView` on `EditorData` lets us just swap out existing uses of `self.doc`/`editor.with_untracked(|e| e.doc)` with `self.view`/`..`.  
I'm leaning towards just having them `EditorView` and `EditorContentView` be separate, but I might be missing a nicer solution that lets us merge them. If we have them separate, I feel like more distinct names might be called for. Maybe `EditorView` -> `EditorLines`? Eh.  
Any suggestions on a good way to merge them, or whether they just shouldn't be merged at all, would be appreciated.~~

### DIFF
--- a/lapce-app/src/config/editor.rs
+++ b/lapce-app/src/config/editor.rs
@@ -190,4 +190,13 @@ impl EditorConfig {
             self.error_lens_font_size
         }
     }
+
+    /// Returns the tab width if atomic soft tabs are enabled.
+    pub fn atomic_soft_tab_width(&self) -> Option<usize> {
+        if self.atomic_soft_tabs {
+            Some(self.tab_width)
+        } else {
+            None
+        }
+    }
 }
--- a/lapce-app/src/doc.rs
+++ b/lapce-app/src/doc.rs
@@ -1,36 +1,22 @@
-use std::{
-    cell::RefCell,
-    collections::{HashMap, HashSet},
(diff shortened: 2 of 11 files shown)


MinusGix:
I renamed `EditorView` -> `EditorViewData` and `EditorContentView` -> `EditorView`.  
Minor: the `editor_content_view` constructor seems like it should be renamed as well, but we already have `editor_view` as a function in the same file.  

It might be good to move sticky headers to `EditorViewData` eventually, but it isn't necessary.

PR #2441, opened 2023-06-09
Use Listener for commands

MinusGix:

Uses the listener ~~from https://github.com/lapce/floem/pull/55~~ for command receiving.  
This simplifies some of the code, removing the need to use `Some` in all the locations, and makes it more obvious what the command fields are for.  

~~The CI build will fail since this needs to be updated to point at the right commit!~~


dzhou121:
I think ```Listener``` probably doesn't belong in Floem. I feel like it will confuse Floem user with event listeners.

MinusGix:
Okay, I've changed it.

Issue #2417, opened 2023-06-01
Reimplement existing features on Floem Lapce


MinusGix:
The Floem (UI library) version of Lapce will be merged into the `master` branch soon. Once that occurs, bringing it completely up to feature parity with the druid version of Lapce is needed. There is a variety of missing features, some of which were skipped over and some of which have happened since the fork, that need to be reimplemented.  
  
**Easy**: Features that are just a matter of porting code over
[ ] Completion Lens
[ ] IME phantom text
  - might be blocked by backend?
**Medium:** May need a good amount of changes and work
?  
**Hard:** May be intricate to implement, even with the existing implementation. Either due to backend differences
[ ] Diff Editor
  - Should be separate from the document instead.

TitanNano:
Wayland support is also no longer a thing.

WeetHet:
And basically nothing works on macos anymore

MinusGix:
> And basically nothing works on macos anymore

Can you be more specific? I believe dzhou uses the editor on macos, so editing works (etc). Is there a specific feature that's missing? A specific integration with MacOS?

WeetHet:
> > And basically nothing works on macos anymore
> 
> Can you be more specific? I believe dzhou uses the editor on macos, so editing works (etc). Is there a specific feature that's missing? A specific integration with MacOS?

1. No macOS menu bar integration, meaning no way to open a folder/create new file as
2. Shortcuts don't seem to work
3. LSP also is not working at all somehow
4. Closing last window closes the app — not macOS way of doing things
5. And probably a lot more stuff I didn't found in the first thirty seconds of using nightly build

MinusGix:
Thanks. I'll add those to the list and discuss them.

> 3. LSP also is not working at all somehow

Which specific LSP?

WeetHet:
> Thanks. I'll add those to the list and discuss them.
> 
> > 3. LSP also is not working at all somehow
> 
> Which specific LSP?

Rust

dsweb19778:
![Screenshot_20230604_214730](https://github.com/lapce/lapce/assets/40462602/95fea256-657f-449a-a48c-ed93e56cc844)
after updating to the new nightly version all is broken nothing working setting dont open shortcuts also 
Os : Linux Manjaro Kde

panekj:
[...]

PR #2458, opened 2023-06-15
fix:reimplementing EnableModal and DisableModal command

tk103331:

Reimplementing EnableModal and DisableModal command


MinusGix:
This has the issue that it won't change the current cursor mode, I believe.   
Our previous impl uses some logic when modal-mode changes to then update the cursor's mode in some reasonable way.  
https://github.com/lapce/lapce/blob/master/lapce-ui/src/editor/view.rs#L972
Which uses  
https://github.com/lapce/lapce/blob/master/lapce-core/src/editor.rs#L1411 for `InsertMode`
https://github.com/lapce/lapce/blob/master/lapce-core/src/editor.rs#L1411 for `NormalMode`  
  
`Document` has its `do_edit` function which would work for this. I presume it would need to be applied to all open editors (all open window tabs and their editors?)

tk103331:
I am referring to the original implementation，just call `LapceConfig::update_file`.

https://github.com/lapce/lapce/blob/3d921f662da9ea885e3d50b29ab39ce2ab59a59b/lapce-data/src/data.rs#L1461-L1478

MinusGix:
The original had implicit side-effects of updating the cursor when the modal mode changed which we don't have currently.  
  
You don't have to implement those side-effects in your PR, but it does mean that the cursor will behave weirdly when changing modal mode. If you want, I can just have a separate issue for updating the cursor.

tk103331:
I understand what you mean, it should be better to open another issue.

PR #2459, opened 2023-06-15
fix:reimplementing logging and OpenLogFile command

tk103331:

fix:reimplementing logging and OpenLogFile command


panekj:
We will most likely not reimplement same logging system

tk103331:
The task in issue https://github.com/lapce/lapce/issues/2417 :


image

I think if you want to use other solutions, you should write clearly here.

tk103331:
duplicated
https://github.com/lapce/lapce/pull/2466

Issue #2326, opened 2023-04-06
Global search do not add selected text to search input


Aukstkalnis:
## Lapce Version

v0.2.7

## System information

Linux Fedora 37.

## Describe the bug
When I select tex, which I want to search throughout entire project and press "ctrl+shift+f" Lapce opens search panel but do not paste selected text in search input. If I use VSCode, for example, it pastes selected text and not toggles search panel. 

Maybe this was intentionally develop, but this is not convenient. cttl+shift+f should open search panel and paste selected text. If new text is selected and  cttl+shift+f is pressed, then previous text should be released. If no text is selected and cttl+shift+f is pressed,  and search panel is already visible it should only put cursor to search input field, and if not visible, then it should open search panel and put cursor to search input field.

## Additional information
Interestingly, when I press ctrl+f selected text is pasted in local search input field and in global search input field.

PR #2465, opened 2023-06-18
feat: update search input when showing search panel

mky2:
Addressing issue #2326. Implementation is copied from existing one in https://github.com/lapce/lapce/blob/640179a0a572e4f194d2b12f0128bb2eb79bf131/lapce-data/src/editor.rs#L1673-L1680


panekj:
lapce-data is not used anymore in lapce, all logic is part of lapce-app

mky2:
Thank you for letting me know, I will close this request then.

PR #2467, opened 2023-06-18
feat: add terminal profiles

### LINKED ISSUE #980: Terminal Settings for paths with spaces fail
On Windows, I was trying to set my terminal to git bash which is located in the `Program Files` directory. The code for configuring the shell splits the setting on spaces which: https://github.com/lapce/lapce/blob/master/lapce-proxy/src/terminal.rs#L59, this causes the terminal pane to crash.

### LINKED ISSUE #1155: [Win11] Terminal Shell Option Shows No Effect
## Lapce Version
Version 0.2.0

## System information
Windows 11 Pro 22H2 Build 22621.590

## Describe the bug
Tried changing terminal shell, but no change occured. Option changed "Shell (Set the terminal shell)"

Attempted:
"C:\Program Files\Git\git-bash.exe"
C:\Program Files\Git\git-bash.exe
git-bash.exe

No errors occured, but shell did not change. Powershell still persists even after restarting the program multiple times after the changes mentioned above.

## Additional information
No ongoing theories thus far.

panekj:

Fixes #980 
Fixes #1155

### DIFF
--- a/lapce-proxy/src/terminal.rs
+++ b/lapce-proxy/src/terminal.rs
@@ -1,8 +1,6 @@
-#[cfg(target_os = "linux")]
-use std::process::Command;
 use std::{
     borrow::Cow,
-    collections::{HashMap, VecDeque},
+    collections::VecDeque,
     io::{self, ErrorKind, Read, Write},
     path::PathBuf,
 };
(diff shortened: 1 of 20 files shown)


lapce-proxy/src/terminal.rs
```diff
@@ -41,62 +43,20 @@ pub struct Terminal {
...
-                })
-            } else {
-                let program = parts.next().unwrap();
-                if let Ok(p) = which::which(program) {
-                    config.pty_config.shell = Some(Program::WithArgs {
-                        program: p.to_str().unwrap().to_string(),
-                        args: parts.map(|p| p.to_string()).collect::<Vec<String>>(),
-                    })
-                }
-            }
```
dzhou121, line 98: Why do you remove the flatpack related stuff?

panekj: Because it was a hack, with terminal profiles user can configure if they want to use host tools or not

PR #2475, opened 2023-06-21
reimplementing source control branches

tk103331:

reimplementing source control branches

image


dzhou121:
Would you mind implement this in the palette, because there's lots of UI interactions missing in the current implementation. It would be much easier in the palette.

dzhou121:
Actually I think it's already being worked on here https://github.com/lapce/lapce/pull/2485

tk103331:
duplicated

PR #2477, opened 2023-06-22
Fix reimpl show language in status

tk103331:

image


panekj:
please rebase on master

PR #2482, opened 2023-06-24
fix: re-implement cursor info

panekj:


Toromambo (commented):
8368dbf4fd039ec9d41d68e4a5c18bd415d531fb

PR #2485, opened 2023-06-25
fix: re-implement reference checkout, include tags

panekj:

### DIFF
--- a/lapce-app/src/palette.rs
+++ b/lapce-app/src/palette.rs
@@ -46,6 +46,7 @@ use crate::{
     keypress::{condition::Condition, KeyPressData, KeyPressFocus},
     main_split::MainSplitData,
     proxy::path_from_url,
+    source_control::SourceControlData,
     window_tab::{CommonData, Focus},
     workspace::{LapceWorkspace, LapceWorkspaceType, SshHost},
 };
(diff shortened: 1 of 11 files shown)


lapce-app/src/palette.rs
```diff
@@ -95,6 +96,7 @@ pub struct PaletteData {
     pub executed_run_configs: Rc<RefCell<HashMap<(RunDebugMode, String), Instant>>>,
     pub main_split: MainSplitData,
     pub references: RwSignal<Vec<EditorLocation>>,
+    pub source_control: RwSignal<SourceControlData>,
```
dzhou121, line 99: ```SourceControlData``` doesn't need to be a signal.

panekj: Fixed

lapce-app/src/palette/kind.rs
```diff
@@ -24,13 +25,7 @@ impl PaletteKind {
             // PaletteKind::GlobalSearch => "?",
             PaletteKind::Workspace => ">",
             PaletteKind::Command => ":",
-            PaletteKind::File
```
dzhou121, line 27: I would leave this so that you don't forget if it has a kind symbol when a new palette kind is added.

lapce-app/src/palette/kind.rs
```diff
@@ -59,21 +54,14 @@ impl PaletteKind {
 
     pub fn get_input<'a>(&self, input: &'a str) -> &'a str {
         match self {
-            PaletteKind::File
```
dzhou121, line 62: and this

Issue #2446, opened 2023-06-10
Horizontal scroll bar not automatically hidden when no text is out of bounds


OTonGitHub:
## Lapce Version
0.2.8

## System information
Windows 11

## Describe the bug
In the code editor pan, if the lines do not exceed horizontal space available, the horizontal scroll bar does not appear, of course, once the line gets too long and get out of bounds, the bar appears, and you can scroll.
The issue is that, if for example you split the long line and stay in bound, the horizontal scroll bar still does not disappear, even though it is not needed. VSC has this behavior correctly, you will know for sure there is nothing left at the end of the line if this behavior is corrected to have the bar disappear once text is back in bound. If you restart the application the bar goes away, but it should have been gone automatically. This is not really a breaking bug, nor does it feel like a feature, something to fix, hope someone attends to this, I hope I have been clear, I have attached a video, I could tackle this myself too actually, if given the chance, thank you.

https://github.com/lapce/lapce/assets/81725875/6dbe6afe-85b8-40c4-bd42-8e83b048eb35

panekj:
This issue is fixed in `master` already

OTonGitHub:
I have built from masters and the horizontal scrollbar related issue does indeed seem to be fixed.
However some other features from 0.2.8 seems to not be in master or rather bugs present in master. Will open new issue for that.

Anyway, as mentioned, issue seems to be fixed, I will now send issue for closing.

panekj:
> Will open new issue for that.

Please don't. #2417

OTonGitHub:
gotcha, I missed the pin, have not actively checked repo in a while (is there actually another way to stay up to date on what's happening to Lapce? I know less platforms the better, just asking. I am new to git etc)

Some stuff seems to be missing in the task list in the migration task list. I will mention it as a comment in the pinned issue. How to actually contribute to this repo, I found some very easy to fix (actual) bugs.

panekj:
we are most active on Discord, 2nd place would be here (GitHub Issues and Pull Requests), anywhere else it's mostly major news

PR #2496, opened 2023-07-01
Fix Horizontal scroll bar not automatically hidden when no text is out of bounds.

weykon:
- [ ] Fix horizontal scroll bar not automatically hidden when no text is out of bounds.

to commit #2446


panekj:
`lapce-ui` is not used anymore, all application logic is in `lapce-app`

panekj:
it's always `HEAD`, which is `master`

PR #2524, opened 2023-07-28
Add libxkbcommon-x11-devel as a dependency in lapce.spec

Titaniumtown:
Fixes build on Fedora and other rpm-based distros.


Titaniumtown:
I don't know of a better solution for builds based on the master branch except for maintaining a separate fork with the only change being the addition of a .spec file.

PR #2526, opened 2023-07-30
Re-implemented "About Lapce" popup.

Wcubed:
Implemented the "about lapce" popup for the floem ui. Progress for #2417.

Opening the popup greys out the rest of the editor, just like in the druid version. Clicking anywhere on the greyed-out area, or pressing the "close modal" key closes the popup.

about


dzhou121:
Excellent work. Thanks!

PR #2527, opened 2023-07-31
New reactive

dzhou121:


lapce-app/src/editor.rs
```diff
@@ -290,6 +285,7 @@ impl EditorData {
             }
         }
 
+        println!("run edit command");
```
Wcubed, line 288: Are all the print lines intentionally left in 😃 ?

dzhou121: left by accident :)

PR #2548, opened 2023-08-15
fix: update toml_edit and use toml crate for easy parsing

panekj:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -757,16 +757,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bf43edc576402991846b093a7ca18a3477e0ef9c588cde84964b5d3e43016642"
 
 [[package]]
-name = "combine"
-version = "4.6.6"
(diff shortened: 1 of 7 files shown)


dzhou121:
What’s the benefit of this?

panekj:
`toml_edit` removes `easy` feature in newer versions

panekj:
https://github.com/toml-rs/toml/blob/25fafe6766e8f7bf5353ea8a1afbf889ae38e9b2/crates/toml_edit/CHANGELOG.md?plain=1#L105

panekj:
It's also to remove duplicate crate versions but since everything is connected and depends on each other it's quite hard to do properly at once

panekj:
> additional toml crate?

We already have it in lockfile so it's not really additional

https://github.com/lapce/lapce/blob/f36c9612edd9db1504f30a5650583ba0d0318e51/Cargo.lock#L4836-L4843

dzhou121:
Got it. Thanks!

PR #2549, opened 2023-08-15
Vim motion multi-line yank and delete

MarcusGrass:

Adds multi-line yank and delete for vim-motions, `3yy`, `5dd` f.e. 

Achieved by only taking the count from the keypress state if the command was indeed executed, and only marks the command as executed when a change is committed. 

The behavior of something like `3dj` is the same as the previous behaviour on `dj`.

I made an exception to the clippy too many args on one function, I could pack that into a struct, tuple, or whatever, depending on if that's preferred.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Features/Changes
 - [#2425](https://github.com/lapce/lapce/pull/2425): Reimplement completion lens
 - [#2498](https://github.com/lapce/lapce/pull/2498): Show Lapce as an option when doing "Open With..." on Linux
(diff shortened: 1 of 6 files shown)


dzhou121:
Thanks and welcome to Lapce!

One suggestion around `count` is maybe we keep the old way to `take()` it, and then store it with motion mode. There's a bit of ambiguity on `CommandExecuted` so it might not be safe to rely on it to put count back.

MarcusGrass:
@dzhou121 Changed to store count in motion mode, then reuse the cached MotionMode if the discriminant is the same

dzhou121:
Thanks!

PR #2558, opened 2023-08-19
Blink cursor

dzhou121:


lapce-app/src/config/editor.rs
```diff
@@ -116,7 +116,7 @@ pub struct EditorConfig {
     #[field_names(
         desc = "Set the cursor blink interval (in milliseconds). Set to 0 to completely disable."
     )]
-    pub blink_interval: u64, // TODO: change to u128 when upgrading config-rs to >0.11
+    blink_interval: u64, // TODO: change to u128 when upgrading config-rs to >0.11
```
panekj, line 119: that comment can be removed since it was meant only for compatibility with druid

PR #2565, opened 2023-08-26
Bump floem, new style methods

jrmoulton:
update lapce for new style methods. Also removed references to scroll bar color and scroll bar color is only referenced in the top View Style


dzhou121:
I've already done this on my branch.

PR #2569, opened 2023-08-27
Context menu pos

zarathir:
This PR changes the calls to open the context menus from `show_context_menu` to the new `popout_menu` and `context_menu` functions from floem.

I'm opening this as a draft, as I currently only have access to a linux machine and the context menus don't work on linux (I tested the first commit on Windows and it worked, so the rest should also JustWork™). Maybe someone else has the time to test this, otherwise I'll have access to a windows machine in a week.

### DIFF
--- a/lapce-app/src/editor.rs
+++ b/lapce-app/src/editor.rs
@@ -174,7 +174,6 @@ pub struct EditorData {
     pub find_focus: RwSignal<bool>,
     pub active: RwSignal<bool>,
     pub sticky_header_height: RwSignal<f64>,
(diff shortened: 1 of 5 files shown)


Wcubed:
Can confirm that the popouts should work on windows, as I just did almost the exact same pull request :P. One thing to note though: For some reason the right click context menu in the git changes window doesn't open. At least, it didn't when I tried porting it to `context_menu`.

zarathir:
I'm closing this in favor of #2570. I'll try to have a look at `context_menu`, once I'll have some more time to dig into it.

PR #2574, opened 2023-08-28
Add new palette menu listing palette kinds

### LINKED ISSUE #2445: Show Help Information in Palette

It would be useful to have a quick way of getting the possible palette kinds (by prefix) while in the editor.  

When you enter `?` into the palette, VSCode (for example) shows the various prefixes that can be used.

We could do something like showing it on the side of the palette, like as extra buttons, when the palette is empty. I think the separate `?` menu is simpler, though.

riverbl:
The new menu is accessible by typing `?` into the palette, or by executing the `palette.palette_help` command.

It only lists palette kinds that are accessible by typing a prefix into the palette, and does not list itself.

Closes #2445.


lapce-app/src/palette/kind.rs
```diff
@@ -19,6 +24,7 @@ impl PaletteKind {
     /// The symbol/prefix that is used to signify the behavior of the palette.
     pub fn symbol(&self) -> &'static str {
         match &self {
+            PaletteKind::PaletteHelp => "?",
```
panekj, line 27: `?` is GlobalSearch prefix

riverbl: What prefix would you like to be used?

lapce-app/src/command.rs
```diff
@@ -31,7 +31,7 @@ pub struct LapceCommand {
     pub data: Option<Value>,
 }
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
```
riverbl, line 34: I've changed `CommandKind` and all its subcommand types to implement `Copy`. I'm wary of making a far reaching change like this, but couldn't see any reason why these types shouldn't be `Copy`, so decided to make the change and let the maintainers kick it back if it is incorrect.

There are some clippy warnings about cloning `Copy` types as a result of this change - I'll make another commit fixing these if the change is OK.

lapce-app/src/palette/kind.rs
panekj: It's not currently used because it's commented
image

lapce-app/src/command.rs
dzhou121: maybe it can't be `Copy` because of strum? I'm not sure. Let's get the palette one merged in first.

lapce-app/src/palette/kind.rs
dzhou121: > What prefix would you like to be used?

I discussed with @panekj and `?` is fine.

lapce-app/src/command.rs
riverbl: I just did a quick search and didn't find anything on `strum` causing problems with `Copy`, and it compiles fine with the change and seems to work - but fair enough. I'll revert the addition of `Copy`.

PR #2596, opened 2023-09-04
Update building-from-source.md

theplasticbrick:
## Lapce Version

0.2.8 

## System information

Fedora running on WSL2 as well as a clean install of Fedora running in a VM

## Describe the bug
The source fails to compile when running "cargo build --release" using the standard instructions for building from source. 

## Additional information
Based on the error logs, the compilation was failing due to two modules not being found during compilation.
Once I had installed the perl-FindBin and perl-File-Compare modules using the standard `sudo dnf install` the compilation completed without issue.
`sudo dnf install perl-File-Compare perl-FindBin`

I found this on Fedora running on WSL2 , and recreated it with a fresh install of Fedora (fully updated) with only Git (via dnf install) and rustup installed (using 
`curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
 per the Rust documentation here https://forge.rust-lang.org/infra/other-installation-methods.html)

### DIFF
--- a/docs/building-from-source.md
+++ b/docs/building-from-source.md
@@ -12,7 +12,7 @@ sudo apt install clang libxkbcommon-x11-dev pkg-config libvulkan-dev libwayland-
 ```
 #### Fedora
 ```sh
-sudo dnf install clang libxkbcommon-x11-devel libxcb-devel vulkan-loader-devel wayland-devel
+sudo dnf install clang libxkbcommon-x11-devel libxcb-devel vulkan-loader-devel wayland-devel perl-File-Compare perl-FindBin
 ```
 #### Void Linux
 ```sh


lineality:
unless this is already resolved: installing pearl first fixed build-fail for me: 
## Fedora:
follow instructions here:
https://github.com/lapce/lapce/blob/master/docs/building-from-source.md 


Follow instruction below to set path to lapce, where path to 
lapce is like: home/YOURNAME/lapse/target/release/lapce


1. Install Pearl (from official fedora source)

https://developer.fedoraproject.org/tech/languages/perl/perl-installation.html
```bash
$ sudo dnf install perl-core
mkdir -p ~/.local/bin
```


2. Create the ~/.local/bin directory
```bash
mkdir -p ~/.local/bin
```

3. Create a symbolic link:
   - Open a terminal and run the following command to create a symbolic link from the executable to the chosen directory:
     ```bash
     ln -s home/YOURNAME/lapse/target/release/lapce ~/.local/bin/lapce
     ```

4. Add the directory to PATH (if necessary):
     ```bash
     export PATH="$PATH:$HOME/.local/bin"
     ```

5. Refresh the shell:
   - Run the following command to reload the shell configuration:
     ```
     source ~/.bashrc
     ```

6. Now you can call "lapce ." or "lapce --version" from cli

PR #2604, opened 2023-09-07
feat(editor): bracket colorization

bbrabbitt:
This is a continuation of #2344 . I open a new PR to make it clearer.


dzhou121:
If I understand it correctly, it parses the whole document on every update. I'm concerned about the performance for big documents. Do you think you can change it to only parse the part on the screen?

bbrabbitt:
> If I understand it correctly, it parses the whole document on every update. I'm concerned about the performance for big documents. Do you think you can change it to only parse the part on the screen?

Yes and that's my plan for next step. I also roughly tested the current implementation on an M2 mba and in release mode I start experience lagging when the file is ~20000 lines long. Without colorization I can experience the similar lagging at ~30000 lines. I can add a switch to disable colorization beyond 20000 lines or so until I work out the partial update.

awoimbee:
> Do you think you can change it to only parse the part on the screen?

As bbrabbitt already kinda said, it's much easier said than done, as is explained in this [vscode blog post](https://code.visualstudio.com/blogs/2021/09/29/bracket-pair-colorization) (originally linked to by bbrabbitt).

> I can add a switch to disable colorization beyond 20000 lines

This is definitively required but based on the byte size of the file, if vscode can open >15MB single line json files, lapce should also do it, but faster.

PR #2622, opened 2023-09-19
Add support for highlighting enclosing brackets

riverbl:
Highlight enclosing brackets when `highlight_matching_brackets` is enabled.

Draw lines delineating the enclosing scope when `highlight_scope_lines` is enabled.

Currently only draws the highlights and scope lines for one cursor, even if there are multiple cursors.

Currently doesn't draw highlights or scope lines on the sticky header.

Relates to #2417.

### DIFF
--- a/lapce-app/src/editor/view.rs
+++ b/lapce-app/src/editor/view.rs
@@ -1,4 +1,4 @@
-use std::{collections::HashMap, rc::Rc, sync::Arc};
+use std::{cmp, collections::HashMap, rc::Rc, sync::Arc};
 
 use floem::{
     action::{set_ime_allowed, set_ime_cursor_area},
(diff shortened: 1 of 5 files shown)


lapce-app/src/editor/view.rs
```diff
@@ -843,6 +843,210 @@ impl EditorView {
...
+                .map(|line_info| line_info.y as f64 + line_height);
+            let end_line_y = screen_lines
+                .info
+                .get(&end_line)
+                .map(|line_info| line_info.y as f64 + line_height);
+
+            // We only need to draw anything if start_line is on or before the visible section and
+            // end_line is on or after the visible section.
+            let y0 = start_line_y.or_else(|| {
+                (start_line < *screen_lines.lines.first().unwrap())
```
riverbl, line 940: Is it valid to assume that `screen_lines.lines` is non-empty and sorted?

lapce-core/src/syntax/mod.rs
```diff
@@ -841,11 +835,13 @@ impl Syntax {
             if start >= self.text.len() {
                 return None;
             }
-            let c = self.text.byte_at(start) as char;
-            if matching_pair_direction(c) == Some(true) {
-                let end = self.find_matching_pair(start)?;
-                if end >= offset {
-                    return Some((start, end));
+            if start < offset {
```
riverbl, line 838: Previously the the offset immediately before an opening bracket was considered enclosed (i.e.: `|(    )`). This was inconsistent with `WordCursor::find_enclosing_pair`, which is used for language unaware bracket search, so I've changed it to not consider that offset to be enclosed.

lapce-core/src/word.rs
```diff
@@ -428,7 +428,7 @@ impl<'a> WordCursor<'a> {
             if matching_pair_direction(c) == Some(true) {
                 let opening_bracket_offset = self.inner.pos();
                 if let Some(closing_bracket_offset) = self.match_pairs() {
-                    if (opening_bracket_offset..closing_bracket_offset)
+                    if (opening_bracket_offset..=closing_bracket_offset)
```
riverbl, line 431: This fixes a bug where the offset immediately before the closing bracket (i.e.: `(   |)`) was not considered enclosed.

lapce-app/src/editor/view.rs
dzhou121: It's sorted but could be empty I think.

riverbl: I've changed it to not assume that `screen_lines.lines` is non-empty.

Issue #898, opened 2022-07-29
shortcut don't work in other languages


yzn-h:
when I switch to Arabic I cant ctrl+z i have to switch layout to undo :(

MinusGix:
Is this a different keyboard layout? If so, what operating system are you on?

yzn-h:
I'm using pop os
when i use the defulte arabic layout i cant ctrl+z but when using the defulte english layout ctrl+z work

Mart-Bogdan:
I can confirm. Same here, on Windows 10. 
Having Cyrillic keyboard layout.

Issue #879, opened 2022-07-27
Copy/Paste doesn't work when caps lock is enabled


jath03:
Most applications afaik can still use Ctrl+c, Ctrl+v, and most other shortcuts regardless of whether the caps lock is enabled.  In lapce, however, enabling caps lock basically adds a <shift> keypress to each shortcut you attempt to use.

Mart-Bogdan:
I think it is related to #898

PR #2634, opened 2023-09-25
Layout-agnostic shortcuts

victorcrimea:
Before this PR Lapce (both druid and floem variants) handled keybindings via keys' textual representation. 

For example:
`ctrl+s` worked only when the keyboard layout was English or a similar Latin layout. If it is pressed with Cyrillic layout it didn't work.

This fixes https://github.com/lapce/lapce/issues/898 

- [ not yet ] Added an entry to `CHANGELOG.md` if this change could be valuable to users


victorcrimea:
fixes also https://github.com/lapce/lapce/issues/879

dzhou121:
One issues with using scan code as the default is the Vim key bindings. Maybe we could put the scan code key in `[]` just like VScode does, so that we can use both for the key binding.

https://code.visualstudio.com/docs/getstarted/keybindings#_keyboard-layoutindependent-bindings

victorcrimea:
@dzhou121  I do not understand what is the problem with vim key bindings? Can't we use scancodes for them as well?

dzhou121:
vim uses lots of characters as key bindings, e.g. ```$```, ```*```, ```/``` etc. And because they are not scan codes in Vim, you might break things if you use scan code for it. 

What's your thoughts on ```[]``` for scan code key bindings?

victorcrimea:
Having a flexibility to choose from scancode/textvalue may be a good option. At least if millions already use it in vscode. I'm also digging through other software to review their solution. Do not want to reinvent the wheel.

Sorry, as non-modal user I do quite not understand how do you use $ as keybinding anyway. THere is no suck key on keyboard. I assume "Shift+4" should be pressed to produce, is it true? So you probably cannot assign keybinding like "shift+$" because shift is already taken for producing dollar sign itself.

dzhou121:
`$` is shift+4 for standard US layout but probably not for some other layout. It's just `$` in Lapce because it uses the textual representation.

victorcrimea:
Just checked the behaviour of the vscode: while they additionally support specifying keycodes in [] brackets. Their usual shortcuts work great in cyrillic.

victorcrimea:
If this PR does not break current behavior in both modal and non-modal modes I'd suggest discussing and adopting new/improved keybinding model in a separate issue/PR. This one is solely to fix not working keybindings in layouts other than US.
I know this brings issue to dvorak layout users, but I assume that alternative non-latin keyboard users(Greek,Russian,Georgian,Ukrainian,Bulgarian, etc) will benefit from this change in the interim, before better, more flexible solution will be found.

dzhou121:
That makes sense. Thanks for the PR.

dholth:
This isn't working for me on OSX with a Dvorak layout and revision `f72401629787cea3a0cd45f2d62dc138b464f345 Fri Oct 6 03:49:27 2023 -0400`; is extra (re-) configuration required?

victorcrimea:
@dholth Unfortunately that was a quick fix that fixes keybindings for all non-latin layouts but at the same time breaks alternative latin layouts(qwertz,azerty,dvorak,colemak). 

I'm working on a better implementation when all Latin layouts will work as expected and non-Latin layouts will work too. Discussion is [here](https://github.com/lapce/lapce/issues/2636).

In the interim, I can suggest revering this PR so you can use Lapce.

victorcrimea:
Reworking the config file is also possible by replacing letters in keybindings with qwerty variant of the dvorak keys.

PR #2668, opened 2023-10-10
feat: add thrift language id support

joyme123:


joyme123:
I want to develop [thrift lsp](https://github.com/joyme123/thrift-ls) plugin, so thrift language id is needed here.

PR #2672, opened 2023-10-11
feat: support thrift highlight

joyme123:
image


joyme123:
#272

panekj:
We're not accepting more built-in grammars anymore

PR #2683, opened 2023-10-15
Terminal font height setting

zarathir:
The terminal font height settings are different then the editor settings. This PR changes the calculation of the terminal font height so the setting now behaves the same as setting the editor font height.

This PR also adds a contrast checker for the terminal, which switches the foreground color to the terminal background color if the contrast is bad.

dark theme

light theme

### DIFF
--- a/lapce-app/src/terminal/view.rs
+++ b/lapce-app/src/terminal/view.rs
@@ -7,7 +7,10 @@ use alacritty_terminal::{
 use floem::{
     cosmic_text::{Attrs, AttrsList, FamilyOwned, TextLayout, Weight},
     id::Id,
-    peniko::kurbo::{Point, Rect, Size},
+    peniko::{
+        kurbo::{Point, Rect, Size},
+        Color,
+    },
     reactive::{create_effect, ReadSignal},
     view::{ChangeFlags, View},
     Renderer,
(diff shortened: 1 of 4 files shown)


lapce-app/src/terminal/view.rs
```diff
@@ -291,7 +294,12 @@ impl View for TerminalView {
             }
 
             if term_bg != bg {
-                let rect = Size::new(char_width, line_height)
+                let term_fg = *config.get_color(LapceColor::TERMINAL_FOREGROUND);
+                if check_contrast(&fg, &bg) && term_fg == fg {
+                    fg = term_bg;
+                };
+                let width = char_width * 1.1;
```
zarathir, line 301: I'm not sure if this is the right way to do this, but without this modifier the background only renders until the end of the cell. This looks like there is a frame around the cell.

dzhou121:
I'm not sure if we should check the contrast. I feel like it should be something that the theme author makes right. Also, I'm not sure about the overhead of this check.

zarathir:
I don't mind reverting the contrast check.
I thought it would be a nice addition, but I see what you mean with the overhead, as this check is called on (almost) every terminal paint call.

dzhou121:
If you could revert the contrast check, that would be great.

zarathir:
@dzhou121 Done :)

dzhou121:
Thanks!

PR #2686, opened 2023-10-17
add windows-sys, add error popup on panic, bring back panic hook to logfile

panekj:

### DIFF
--- a/lapce-app/src/app.rs
+++ b/lapce-app/src/app.rs
@@ -3375,6 +3375,8 @@ pub fn launch() {
     let (reload_handle, _guard) = logging();
     tracing::info!("Starting up Lapce..");
 
+    panic_hook();
+
     // if PWD is not set, then we are not being launched via a terminal
     #[cfg(any(target_os = "macos", target_os = "linux"))]
     if std::env::var("PWD").is_err() {
@@ -3421,7 +3423,16 @@ pub fn launch() {
     crate::update::cleanup();
 
     let _ = lapce_proxy::register_lapce_path();
-    let db = Arc::new(LapceDb::new().unwrap());
+    let db = match LapceDb::new() {
+        Ok(db) => Arc::new(db),
+        Err(e) => {
+            #[cfg(windows)]
+            error_modal("Error", &format!("Failed to create LapceDb: {e}"));
+            #[cfg(not(windows))]
+            error!("Failed to create LapceDb: {e}");
+            std::process::exit(1);
+        }
+    };
     let scope = Scope::new();
     provide_context(db.clone());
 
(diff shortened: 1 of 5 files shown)


lapce-app/src/command.rs
```diff
@@ -498,6 +498,10 @@ pub enum LapceWorkbenchCommand {
     #[strum(serialize = "previous_error")]
     PreviousError,
 
+    #[strum(serialize = "panic")]
+    #[strum(message = "Panic Editor")]
+    Panic,
```
dzhou121, line 503: Do we really want this command in the palette?

lapce-app/src/app.rs
```diff
@@ -3719,3 +3730,70 @@ pub fn window_menu(
...
+            }
+        }
+
+        #[cfg(windows)]
+        error_modal("Error", &info.to_string());
+    }))
+}
+
+#[cfg(windows)]
+fn error_modal(title: &str, msg: &str) -> i32 {
```
dzhou121, line 3771: I don't see the need to use a modal for panic messages.

dzhou121: The log should be enough I think?

panekj: Log doesn't give a visual clue that program failed to launch or why

lapce-app/src/command.rs
panekj: I don't think there is anything wrong with that

dzhou121: I find it odd to have a command to crash your editor.

lapce-app/src/app.rs
dzhou121: It would be more useful if mac and linux can have it so that it's more consistent between platforms.

panekj: That's my goal but I don't know how to do that on mac/linux yet
macOS does have own pop-up though so I'm also unsure if that's needed, I'll need to test how it behaves
app-quit-unexpectedly-crash-report-dialog-window-mac-610x405-3458731267
linux might be tricky because of wayland/xorg

lapce-app/src/command.rs
panekj: For basic users, that's true, but when developing the editor it sometimes might be helpful (like testing this change) I can remove it though if you want

lapce-app/src/app.rs
dzhou121: It makes sense to me now. We can add support for other platforms in other PRs.

lapce-app/src/command.rs
dzhou121: Yes please. For testing, you can just panic!() after the hook?

PR #2698, opened 2023-10-25
Added navigation keybindings

MyGodItsFull0fStars:

Added keybindings that are often used in macOS systems or the emacs editor to navigate UP or DOWN via `CTRL+P` and `CTRL+N` respectively.

I made sure to ignore these keybindings in case the `list_focus` mode is enabled, so the navigation in this mode is not overwritten by the added keybindings.

Please let me know if I need to make any modifications or improvements.

### DIFF
--- a/defaults/keymaps-common.toml
+++ b/defaults/keymaps-common.toml
@@ -218,6 +218,18 @@ command = "right"
 mode = "i"
 
 [[keymaps]]
+key = "ctrl+p"
+command = "up"
+when = "!list_focus"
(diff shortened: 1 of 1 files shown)


panekj:
If the keybinding comes from emacs, why is it added to modal mode?

dzhou121:
If you don't specify the insert mode, then it would work in all modes, which isn't ideal.

MyGodItsFull0fStars:
Thank you for accepting my PR!

PR #2723, opened 2023-10-29
Line wrap

MinusGix:

Not done yet! But almost... 
I'll probably just have to collapse the commits or something. There's a lot of in-progress stuff in there that aren't really useful intermediate information or useful intermediate builds.  

----

General Introduction:  
This uses cosmic-text's line-wrapping to allow us to wrap at some width, whether viewport or specifically written width.  
- Currently, there is no column wrapping! Cosmic-text does not seem to support this, I believe it wouldn't be hard to implement... but do that in a separate PR.  

This introduces a `Lines` structure in `visual_line.rs`.  
`Lines` is per `EditorViewData`, and stores the text layouts for that view. It provides a variety of functions for iterating over the lines in the editor by how they are wrapped.  
  
There are two new ways of talking about lines: `VLine` (Visual Line) and `RVLine` (Relative Visual Line).  
`VLine` is whatever the actual *position* (sortof index) the line (or piece of a wrapped line) is at. This can be expensive to compute, because it essentially requires looking over all the positions previously. Usually it is not needed, but it is needed for computing `ScreenLines`.  
`RVLine` is the more typical structure. It is essentially a `(line, line index)` pair. The line is the typical buffer line, and the line index is which wrapped line it is referring to.  
Think of:  
```text
1 | this is a wrapped
    |  line
2 | this is a line
```
Then the first displayed line ("this is a wrapped") has a buffer line of `0`, a VLine of `0`, and an RVLine of `(0, 0)`  
The second displayed line (" line") has a buffer line of `0`, a VLine of `1`, and an RVLine of `(0, 1)`. Since it is the second part of the overall buffer line.  
The third displayed line ("this is a line") has a buffer line of `1`, a VLine of `2`, and an RVLine of `(1, 0)`.  
  
Essentially, for line wrapping, we have to break the one-to-one translation between the VLine (line positioned in the display) and the buffer line. You can still do `vline * line_height` to get the position in the screen.  
  
`RVLine` is what is typically used, because most of the code does not care about the absolute position information. Ideally, most code should rarely deal with `VLine`s.  
  
-------

Most positioning logic now has to take care of affinity. `CursorAffinity` is what decides where to place the cursor when one offset has potentially multiple positions.  
Like how at the end of a wrapped line, the very end is the same offset as the start of the next piece.

-------

Other changes:
- Diff's `vline` is now referred to as `y_idx`, because currently it does not actually necessarily match up with a real line.
  - But I think it might be a good idea to have `Lines` keep track of this in the future, as a general arbitrary space-taking widgets features
- `ScreenLines` is now cached in a kinda wacky manner. I have comments detailing it, but I should rewrite them to be more explanatory.  
  - `ScreenLines` is more expensive to compute, since it has to work with vlines (turning the `viewport.y0` into a vline, because there may be any number of wrapped lines above it)
- This allows inlay hints and error lens to be multiline
  - Error lens is behind an option. I don't think I added one for inlay hints, but nobody does multiline inlay hints

-------

Possible issues:
- I have a variety in my notes that I plan to fix
- Our fork of cosmic-text needs the line-height to be fixed, it is positioning incorrectly for wrapped lines
- Diff behavior might have changed because I got rid of `actual_line` and `visual_line`
- I should really test on big large files! It may be that there's some big problem due to using vlines that it dislikes, but I really hope it doesn't have issues.
- Need to ensure the DAP pr stuff works
- Need to readd the IME code, I removed it temporarily, and install an IME so I can test it
- There may be some screen lines cache issues lurking... I sometimes run into hard to reproduce annoyances.

--------

TODOs (for the future):
- Diff does not currently do any wrapping
  - I can see how I'd structure this, but I don't wanna do it in this PR, so it has wrapping disabled.
  - (To do the synchronization between both sides it needs to do the positioning logic with access to both editors)
- We need to implement some sort of wrapping avoidance for really long lines
  - Ideally along with some other PR that just makes really long lines work nicely even without wrapping...
- It would be nice to have a wrapping style that is like `min(editor-width, wrap-column)`. So you can keep it to an expected column limit, while also keeping everything within view
- We may want to extend the font sizes provides to also provide line heights?
  - I think this mostly depends on how we integrate things like LSP code lens, whether that goes into `Lines` with special handling or some really weird wrapper. I'm leaning towards into `Lines`, via some generic 'special object insertion method' that works for Diff.
- Indent when wrapping. Cosmic-text doesn't support this automatically, but I think it wouldn't be too hard to add ourselves or upstream to cosmic-text?
- Indent multiline error lens, I think this would appear better
- Limit multiline error lens length, some languages/LSPs just spam them because you don't usually see them all in the editor
- Implement Copilot as a plugin
- Way more tests for various multiline phantom text and friends
- Maybe try to modify cosmic-text to have a better API for iterating over parts of lines :)))))
- Eventually get rid of the hacky code that compensates for cosmic-text currently stripping a single space at the end of a wrapped line
- Various todos littered around
- Cosmic seems to sometimes wrap things like `    Some(blah)` to the next line, like one blank line and then the `Some(blah)`, which seems undesirable
- I think `SelRegion` should probably an affinity for start/end, but propagating it through all the logic is uh.... unpleasant
- `ScreenLines` gets recomputed a bunch of times if you edit things, it should be smarter!
  - I think having batch for signals would help, but I also think this is due to multiple relevant signals getting updated when something changes that might not be easily batched...
  - There's certainly stuff I could do about being smarter about when we clear things, but hopefully there's some nice elegant way to do better.


MinusGix:
I squashed all my commits because that made rebasing significantly easier. I likely would have had to do it anyway.  
(I need to look through things that #2803 does and make sure I replicate how that fixes things, since I had to change pieces like I changed the previous impl to make it work; I may have missed/broken some piece of that.)

MinusGix:
I'm not sure of the right way to make the vim-like block selection behave.  
(in nvim):  
image
image

So by default it seems like it goes by the line-index rather than per screen/visual line. Thinking of just putting this off to the future...

MinusGix:
- Line height in cosmic-text still needs to be fixed
- I still want to go through a bunch of my code and todos and clean it up

MinusGix:
Mostly fixes #945–we are missing various pieces of wrapping (column wrapping, indentation of wrapped lines, etc) but those can be broken out as separate issues.

PR #2741, opened 2023-11-01
add dynamic title

panekj:

### DIFF
--- a/lapce-app/src/window_tab.rs
+++ b/lapce-app/src/window_tab.rs
@@ -1400,6 +1400,35 @@ impl WindowTabData {
             }
             InternalCommand::FocusEditorTab { editor_tab_id } => {
                 self.main_split.active_editor_tab.set(Some(editor_tab_id));
+                let editor_tab = self
+                    .main_split
+                    .editor_tabs
+                    .with_untracked(|tabs| tabs.get(&editor_tab_id).cloned());
+                let tab_info = editor_tab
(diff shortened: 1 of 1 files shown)


dzhou121:
It's already setting title like this
https://github.com/lapce/lapce/blob/master/lapce-app/src/app.rs#L3284

panekj:
> It's already setting title like this [`master`/lapce-app/src/app.rs#L3284](https://github.com/lapce/lapce/blob/master/lapce-app/src/app.rs?rgh-link-date=2023-11-01T11%3A11%3A56Z#L3284)

But this just works for workspace?

dzhou121:
It defaults to Lapce if no workspace is open.

panekj:
I don't mean whether it's working with just workspace, this will update the title only for workspace change

dzhou121:
oh you added editor tab to the title. sorry my bad. 

I would prefer it to use `.window_title` on the window_tab view, so that whenever the active editor tab is changed, it will be updated.

panekj:
Should it be attached on `window_tab` and not `editor_tab`?

dzhou121:
Yeah on `window_tab` as you can get the active editor tab there.

dzhou121:
Actually you can just do it here

https://github.com/lapce/lapce/blob/master/lapce-app/src/app.rs#L3284

Get the active editor tab from window_tab

Issue #2736, opened 2023-11-01
Fallback font rendering not working very well in 0.3.0


lnicola:
## Lapce Version

v0.3.0

## System information

Fedora 39

## Describe the bug

Code is rendered using a variable-width font, the terminal is a variable-width font in a fixed-size grid.

I checked the settings and the font was Cascadia Code, which I don't have. Switching to `monospace` didn't change anything. For me, `monospace` is `"Noto Sans Mono"`. Using `Noto Sans Mono` or a different font works.

## Additional information

![image](https://github.com/lapce/lapce/assets/308347/704143d4-ba59-4702-bb4e-cfaae2b2db26)

PR #2744, opened 2023-11-01
update screenshot

panekj:


dzhou121:
It looks a bit wired with variable width font in the editor.

panekj:
It seems to be this issue https://github.com/lapce/lapce/issues/2736

dzhou121:
Another request. Can you make the window size a bit smaller so that it's easier on the eyes for small screens.

panekj:
Is that small enough? I generally try to keep the original aspect ratio so the image scales itself

dzhou121:
Thanks!

PR #2764, opened 2023-11-06
Auto-surround config

### LINKED ISSUE #2751: When "Auto Close Matching Pairs" is disabled, when text is selected and a bracket or quote is pressed it shouldn't auto close or auto surround the text
## Lapce Version

Flatpak version, About dialog says "0.2.8"

## System information

openSUSE Leap 15.5, installed via flathub

## Describe the bug
When "Auto Close Matching Pairs" is disabled, when text is selected and a bracket or quote is pressed it shouldn't auto close or auto surround the text. It should just do exactly what is typed instead of trying to be smart, since that seems to be the point of that option being disabled. Therefore, this seems buggy and inconsistent right now.

## Additional information
This is what happens when I press `[` with "Auto Close Matching Pairs" disabled:

https://github.com/lapce/lapce/assets/64124388/e3502653-045d-411c-b5e1-64a250a8a585.webm

Expected result would be that `abc` disappears and a single `[` is entered instead.

ElmeriVincent:

Fixes #2751

### DIFF
--- a/defaults/settings.toml
+++ b/defaults/settings.toml
@@ -22,6 +22,7 @@ completion-show-documentation = true
 show-signature = true
 signature-label-code-block = true
 auto-closing-matching-pairs = true
+auto-surround = true
 hover-delay = 300                       # ms
 modal-mode-relative-line-numbers = true
 format-on-save = false
--- a/extra/schemas/settings.json
+++ b/extra/schemas/settings.json
@@ -221,6 +221,9 @@
                 "auto-closing-matching-pairs": {
                     "type": "boolean"
                 },
+                "auto-surround": {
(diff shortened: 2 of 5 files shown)


dzhou121:
Maybe we should add a different config for it? This is how it's called in VSCode. 

```
Auto Surround
Controls whether the editor should automatically surround selections when typing quotes or brackets.
```

ElmeriVincent:
> Maybe we should add a different config for it? This is how it's called in VSCode.
> 
> ```
> Auto Surround
> Controls whether the editor should automatically surround selections when typing quotes or brackets.
> ```

Wasn't aware of that.. Sounds good.

PR #2782, opened 2023-11-13
Add UI inspector command

Zoxc:
This adds a command to open the Floem Inspector, which can be handy to investigate performance and layout issues.


lapce-app/src/window_tab.rs
```diff
@@ -1129,6 +1129,9 @@ impl WindowTabData {
             FocusTerminal => {
                 self.common.focus.set(Focus::Panel(PanelKind::Terminal));
             }
+            OpenUIInspector => {
+                self.common.view_id.get().inspect();
```
dzhou121, line 1133: It's better to `get_untracked()`

Issue #2772, opened 2023-11-08
Terminal uses `editor.background` instead of `terminal.background` from color theme


RinLovesYou:
## Lapce Version
nightly de2d7b1

## System information
arch linux

## Describe the bug
The bottom panel's color should be determined by `terminal.background`, but it uses `editor.background`

## Additional information

![image](https://github.com/lapce/lapce/assets/29461788/64ea38f0-201b-4342-aac4-2a5abe48642f)
![image](https://github.com/lapce/lapce/assets/29461788/fa231d7b-c379-43c9-9ba2-b88be23c410a)

ElmeriVincent:
@RinLovesYou Would this be more like it?

![Screenshot from 2023-11-14 21-00-06](https://github.com/lapce/lapce/assets/77973084/5147d32c-d88a-4033-ba4c-4cc3bac8a41a)

PR #2786, opened 2023-11-14
Fix/terminal-background

evnct:
Fixes -> #2772


dzhou121:
panel items don't have backgrounds set including terminal so that when they are at the bottom, they will be the same as editor background. So if you drag the terminal panel to the right or left, it will have the panel background.

ElmeriVincent:
You mean like this (dragging it to the right)? 

Screenshot from 2023-11-14 23-36-37

dzhou121:
Yes.

terminal.background probably shouldn't be there to confuse people.

ElmeriVincent:
Checking if I understood right. What is in the image now at the bottom right side should have the editor background and what is at the top right should have panel background?

Issue #2676, opened 2023-10-14
Weird clipboard crash


MinusGix:
## Lapce Version

Almost latest master

## System information

Linux, Wayland

## Describe the bug
I was spamming typing in my editor to test my screen lines cache and it crashed with
```
Maximum number of clients reached
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: XcbConn(Connection)', lapce-app/src/doc.rs:79:43
```

## Additional information

Which is
```rust
impl SystemClipboard {
    pub fn new() -> Self {
        SystemClipboard {
            ctx: ClipboardProvider::new().unwrap(),
        }
    }
}
```
and it is constructed every time an edit is done.  
Presumably it is either getting constructed too often while other clipboard connections are open, or it isn't getting cleaned up at all.  
Possibly we should just have a general shared instance on `common` or something?

lukaslihotzki:
I can trigger this reliably by holding down `Ctrl+V`.

PR #2794, opened 2023-11-19
Switch clipboard functionality to a maintained crate

tv42:
Switch from crate clipboard to arboard

This avoids a dependency via x11-clipboard to an old version of xcb,
v0.3. Problems and annoyances with xcb v0.3 include

- safety: https://github.com/aweinstock314/rust-clipboard/issues/90
- build script depends on python
- won't build in a sandbox, as it writes to the source directory

See also https://github.com/aweinstock314/rust-clipboard/issues/91

---

initialize SystemClipboard only once

The `x11-clipboard` crate starts a new thread on every `new` call,
previously triggered on every copy to clipboard operation, and these
threads seem to stick around.

This will also help migrating away from the
`clipboard`/`x11-clipboard` crates.

---

### DIFF
--- a/lapce-app/src/doc.rs
+++ b/lapce-app/src/doc.rs
@@ -1,6 +1,7 @@
 use std::{
     cell::RefCell,
     collections::HashMap,
+    ops::DerefMut,
     path::{Path, PathBuf},
     rc::Rc,
     sync::{atomic, Arc},
@@ -437,7 +438,7 @@ impl Document {
             return Vec::new();
         }
 
-        let mut clipboard = SystemClipboard::new();
+        let mut clipboard = self.common.clipboard.lock();
         let old_cursor = cursor.mode.clone();
         let deltas = self.syntax.with_untracked(|syntax| {
             self.buffer
@@ -447,7 +448,7 @@ impl Document {
                         buffer,
                         cmd,
                         syntax,
-                        &mut clipboard,
+                        clipboard.deref_mut(),
                         modal,
                         register,
                         smart_tab,
--- a/lapce-app/src/doc.rs
+++ b/lapce-app/src/doc.rs
@@ -8,7 +8,6 @@ use std::{
     time::Duration,
 };
 
-use clipboard::{ClipboardContext, ClipboardProvider};
 use floem::{
     action::exec_after,
     cosmic_text::{Attrs, AttrsList, FamilyOwned, TextLayout},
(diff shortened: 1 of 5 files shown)


dzhou121:
Thanks for the PR!

Do we want to enable the wayland feature tag for arboard?

dzhou121:
Also, do you know how does it compare with copypasta?

tv42:
Yeah, could enable wayland. I figured that'd be more of a follow-up, to avoid accidentally breaking things (the X11 to X11 part is pretty safe).

I don't know much about copypasta. I looked at APIs and arboard seemed to have the nicest API, and more downloads.

panekj made https://github.com/lapce/lapce/pull/2660 which for some reason points to a fork of copypasta, and panekj's active on Discord, but that's too much politics for me.

panekj:
> panekj made #2660 which for some reason points to a fork of copypasta

it points to a fork since it has aligned version of other dependencies that copypasta uses

> panekj's active on Discord, but that's too much politics for me.

I'm also active here, and the politics are choice of clipboard crate.

dzhou121:
> I don't know much about copypasta. I looked at APIs and arboard seemed to have the nicest API, and more downloads.

I asked about copypasta because it's what alacritty uses and alacritty is a winit application used by lots of users.

tv42:
Well if someone can get #2660 merged I'll rebase this to have the thread leak bugfix.

lukaslihotzki:
arboard uses wl-clipboard-rs for the Wayland clipboard support. wl-clipboard-rs "is intended to be used by terminal applications, clipboard managers and other utilities which don't spawn Wayland surfaces (windows)." Therefore, using wl-clipboard-rs or arboard is a bad fit for Lapce. The clipboard will break when Lapce is run in a container like Flatpak.

Use copypasta like alacritty does is the way to go.

beckend:
Tried copypasta on wayland, it doesn't work. arboard works correctly.

lapce-app/src/doc.rs
```diff
@@ -72,18 +73,32 @@ impl Default for SystemClipboard {
...
+                error
+            })
+            .ok()
     }
 
     fn put_string(&mut self, s: impl AsRef<str>) {
-        let _ = self.ctx.set_contents(s.as_ref().to_string());
+        let _ = self
+            .inner
+            .set_text(s.as_ref().to_string())
```
beckend, line 97: .to_string() not be needed right

lukaslihotzki:
copypasta does not work with short-lived `ClipboardContext`, which is bad design anyway (see also #2676). If you create the `ClipboardContext` once on program start and also do anything else like Alacritty does, everything works fine (like Alacritty). I'll publish my PR now.

panekj:
This has been now implemented as part of floem

PR #2798, opened 2023-11-19
Fix non_canonical_partial_ord_impl clippy lints

riverbl:
* Reimplement `PartialOrd` in terms of `Ord` for some types to fix `non_canonical_partial_ord_impl` (previously called `incorrect_partial_ord_impl_on_ord_type`) clippy lints.

* Fix potential incorrect order or panic when comparing `FileNodeViewData` objects containing invalid unicode.

Because `incorrect_partial_ord_impl_on_ord_type` has been renamed, clippy now produces warnings for the two instances of `#[allow(clippy::incorrect_partial_ord_impl_on_ord_type)]`. I've changed the two `PartialOrd` implementations to make clippy happy and removed the `#[allow]`s.

The new `PartialOrd` implementation for `CharBuffer` is equivalent to the old one.

The new `PartialOrd` implementation for `FileNodeItem` fixes issues with paths containing invalid unicode - the old implementation would return `None` when comparing `foo` to `b�r`, which could result in file `foo` being displayed before file `b�r` in the file explorer. If the `Ord` implementation for `FileNodeItem` was used anywhere then it would panic when attempting to perform such a comparison, but it doesn't look like it was used, so this potential panic was purely theoretical.

My editor automatically removed some trailing whitespace in a multi-line string in `char_buffer.rs`. The relevant test still passes, but I'm not sure if this trailing whitespace was present intentionally - maybe @JustForFun88 could clarify?


riverbl:
It looks like the CI failure for this PR is due to the issue fixed in #2793, and the CI failure for #2793 is due to the renaming of `incorrect_partial_ord_impl_on_ord_type`, fixed by this PR.

lapce-rpc/src/file.rs
```diff
@@ -65,25 +65,26 @@ pub struct FileNodeItem {
...
 impl Ord for FileNodeItem {
     fn cmp(&self, other: &Self) -> Ordering {
-        self.partial_cmp(other).unwrap()
+        match (self.is_dir, other.is_dir) {
+            (true, false) => Ordering::Less,
+            (false, true) => Ordering::Greater,
+            _ => {
+                let [self_file_name, other_file_name] = [&self.path, &other.path]
+                    .map(|path| {
+                        path.file_name().unwrap_or_default().to_string_lossy()
```
panekj, line 82: This will panic

PR #2803, opened 2023-11-20
Reworked cursor rendering, fixing several bugs

riverbl:
* Block cursors were rendered with a fixed width, rather than the width of the current character, when after an inlay hint and using a variable width font.

* In visual normal mode, the selection highlight didn't extend past the end of the current line when the cursor was past the end of the line.

* In visual normal mode, if there was an inlay hint after the end of the current line and the cursor was past the end of the line, the cursor was rendered after the inlay hint.

* In visual normal and block modes, when selecting backwards to the first character after an inlay hint, the selection highlight extended leftward of the cursor to include the inlay hint.

* When there were multiple cursors on the same line, only one was rendered.

* The selection highlight for visual block mode excluded the rightmost column when selecting right and up or left and down.

* When in visual block mode, when the cursor was to the left of the anchor column of the selection and past the end of the line, the cursor wasn't rendered.

### DIFF
--- a/lapce-app/src/editor/view.rs
+++ b/lapce-app/src/editor/view.rs
@@ -24,7 +24,6 @@ use lapce_core::{
     buffer::{diff::DiffLines, rope_text::RopeText},
     cursor::{ColPosition, CursorMode},
     mode::{Mode, VisualMode},
-    selection::Selection,
 };
 use lapce_rpc::dap_types::{DapId, SourceBreakpoint};
 use lapce_xi_rope::find::CaseMatching;
@@ -32,7 +31,7 @@ use lapce_xi_rope::find::CaseMatching;
 use super::{
     gutter::editor_gutter_view,
     view_data::{EditorViewData, LineExtraStyle},
-    EditorData,
+    EditorData, CHAR_WIDTH, FONT_SIZE,
 };
 use crate::{
     app::clickable_icon,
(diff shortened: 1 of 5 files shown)


lapce-app/src/editor/view.rs
```diff
@@ -1268,241 +1424,60 @@ fn get_sticky_header_info(
...
-                            1,
-                        );
-                        let (_, end_col) = view.offset_to_line_col(end_offset);
-                        end_col.max(start_col).min(max_col)
-                    }
-                };
-                (right, false)
-            }
+        let width = if after_last_char {
+            CHAR_WIDTH
```
riverbl, line 1466: Previously, the default block cursor width was `7.0` and the distance the selection highlight would extend beyond the end of the line when the next line was selected was `7.5`. I've defined `CHAR_WIDTH = 7.5` and use it for both cases.

lapce-app/src/editor/view.rs
```diff
@@ -1268,241 +1424,60 @@ fn get_sticky_header_info(
...
 
-        let (right_col, line_end) = match mode {
-            VisualMode::Normal => {
-                if line == end_line {
-                    let max_col = view.line_end_col(line, true);
+                        (preedit_line == line).then_some(start)
+                    })
+                })
+            })
+            .unwrap_or(0);
```
riverbl, line 1460: I'm not sure what a preedit is or what this logic is for.

lapce-app/src/editor/view.rs
```diff
@@ -399,65 +509,113 @@ impl EditorView {
             },
         );
```
riverbl, line 510: I wasn't able to get debugging working in Lapce, so haven't tested that the line with the active breakpoint is rendered correctly.

riverbl:
Previously, the lines containing the cursors were highlighted in insert and normal modes, but not in any of the visual modes. This PR changes it so that the lines containing the cursors are highlighted in all modes, as this slightly simplifies the code. If highlighting the line containing the cursor in visual mode is undesirable, this could be changed.

lapce-app/src/editor/view.rs
dzhou121: preedit is for IME

dzhou121:
> Previously, the lines containing the cursors were highlighted in insert and normal modes, but not in any of the visual modes. This PR changes it so that the lines containing the cursors are highlighted in all modes, as this slightly simplifies the code. If highlighting the line containing the cursor in visual mode is undesirable, this could be changed.

current line isn't highlighted in visual mode was to keep the same behaviour with Vim.

riverbl:
> current line isn't highlighted in visual mode was to keep the same behaviour with Vim.

I've changed it to not highlight the current line in visual mode, and fixed an issue with the previous version of the PR that would have resulted in the line containing the active breakpoint not being highlighted unless there was a cursor on it.

lapce-app/src/editor/view.rs
riverbl: I've done some testing entering emoji with IBus and cursor placement seems to work as well as it did prior to this PR.

IME doesn't work well with multiple cursors - only one cursor displays the preedit text, and with multiple cursors on the same line, the cursors other than the one displaying the preedit text render in the wrong place. These issues were both present prior to this PR though, so I think they could be fixed in a separate PR.

MinusGix:
For removing the whitespace at the end of comments, I think we should keep those in the future. (I feel like I commented on this before saying whatever/its fine to remove them, but might have not; anyway after recently checking what was up with RA not rendering them I'm readjusting my opinion)  
The double spaces are the markdown syntax for making a linebreak, which allows seeing them as newlines within rustdoc. Rust analyzer at the moment trims this whitespace when doing hover documentation, which I believe it did not do previously. See https://github.com/rust-lang/rust-analyzer/issues/15877

(So in future prs please don't cut off the double spaces. Rustfmt doesn't do so by default and Rustfmt would catch extra whitespaces in code)
(also, thanks for the pull requests they're great)

riverbl:
@MinusGix thanks for the heads up - I wasn't aware that trailing whitespace had semantic meaning in Markdown.

I personally think that a trailing backslash is a cleaner way to indicate a hard line break. Lapce, VS Code and `cargo doc` all support this.

Still, I'll make sure not to strip trailing spaces from doc comments in the future.

PR #2804, opened 2023-11-20
Fix invalid unicode filenames not being displayed

riverbl:
Fix filenames containing invalid unicode not being displayed in the breadcrumbs, open editors list or tab titles.

Previously the tab and open editors list would have blank titles if the filename was not invalid unicode, and any directories or filenames containing invalid unicode were omitted from the breadcrumbs.

### DIFF
--- a/lapce-app/src/editor/view.rs
+++ b/lapce-app/src/editor/view.rs
@@ -1931,14 +1931,16 @@ fn editor_breadcrumbs(
                                     .to_path_buf();
                             }
                             path.ancestors()
-                                .collect::<Vec<_>>()
-                                .iter()
-                                .rev()
                                 .filter_map(|path| {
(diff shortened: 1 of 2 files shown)


MinusGix:
For removing the whitespace at the end of comments, I think we should keep those in the future. (I feel like I commented on this before saying whatever/its fine to remove them, but might have not; anyway after recently checking what was up with RA not rendering them I'm readjusting my opinion)  
The double spaces are the markdown syntax for making a linebreak, which allows seeing them as newlines within rustdoc. Rust analyzer at the moment trims this whitespace when doing hover documentation, which I believe it did not do previously. See https://github.com/rust-lang/rust-analyzer/issues/15877

(So in future prs please don't cut off the double spaces. Rustfmt doesn't do so by default and it will catch extra whitespaces in code)

Issue #2787, opened 2023-11-15
Lapce does not cooperate with system clipboard and appears to use its own internal one


alexpyattaev:
## Lapce Version

Lapce v0.3.0 
Nightly 8dc41a0 

## System information

Arch x86_64, KDE wayland 

## Describe the bug
When I copy text into clipboard (via middle mouse button or keyboard hotkey) in any other application, the text can then be pasted as usual (with middle mouse or keyboard hotkey). When I copy text in Lapce (via hotkey) it can then be pasted in lapce (via hotkey). However, there appears to be no connection between lapce clipboard and system clipboard, i.e. text copied from another app can not be pasted into lapce and vice versa.  

## Additional information
I have seen similar issues on other KDE setups. The issue does not seem to occur on Ubuntu desktop with same version of Lapce. 
If necessary I can try to investigate this further but I have no idea what to look for - there are no errors in any logs that I can see.

dvaerum:
I am having the same experience, I am on version v0.3.0, just to say that you are not alone :sweat_smile:

MinusGix:
I'm seeing this on Sway as well, so it isn't strictly KDE probably.

lukaslihotzki:
This is related to #2768 and is caused by Lapce only supporting the X11 clipboard, not the Wayland clipboard. So, Lapce runs its window on Wayland, but opens an extra connection to X11 for the clipboard.

I propose that Lapce should do what Alacritty does, because the Alacritty clipboard works well on Wayland and it also uses winit and copypasta. This would be done by copying the [clipboard.rs file](https://github.com/alacritty/alacritty/blob/master/alacritty/src/clipboard.rs) from Alacritty to Lapce and use that instead of `ClipboardProvider::new()`. Also, we would need the `RawDisplayHandle` from winit (or floem) to pass it to clipboard.rs.

alexpyattaev:
I would have never guessed it is even possible to use X11 clipboard from wayland app... I'd be happy to test any fix if that would help but I'm certainly not qualified to implement this.

lukaslihotzki:
@alexpyattaev, then please test the fix in #2815.

alexpyattaev:
Thank you for addressing this!

Issue #2768, opened 2023-11-07
Crash when pressing enter or backspace in editor


lukaslihotzki:
## Lapce Version

0.3.0

## System information

Arch Linux, Wayland, sway, XWayland disabled (start sway with `xwayland disable` in the config).

## Describe the bug
When XWayland is disabled, pressing Enter or Backspace in the Editor crashes lapce:
```
thread 'main' panicked at lapce-app/src/doc.rs:75:43:
called `Result::unwrap()` on an `Err` value: XcbConn(ClosedParseErr)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Panic is at https://github.com/lapce/lapce/blob/master/lapce-app/src/doc.rs#L75.

## Additional information

This crash is caused by an Xcb connection fail because the Clipboard is X11-only. This issue can be fixed in multiple ways:
- Add native Wayland support for the clipboard.
- Use the clipboard API only if the user does a clipboard-related action.
- Handle the error gracefully, so just copy and paste does not work if the clipboard API errors.

Also, when this code needs a redesign, the ClipboardProvider should be kept after creation. This may fix #2676.

panekj:
Wayland is not supported

Kulasko:
Can confirm this error using Arch/GNOME/X11.

PR #2815, opened 2023-11-25
bump floem and use floem::Clipboard

lukaslihotzki:
After #2794 and #2660, I also made a clipboard PR. It supports Wayland and uses the clipboard protocol for GUI applications there, so the clipboard should reliably work in containers like Flatpak in the future without extra permissions. Fixes https://github.com/lapce/lapce/issues/2787, https://github.com/lapce/lapce/issues/2768, https://github.com/lapce/lapce/issues/2676.

It requires the corresponding PR in floem (https://github.com/lapce/floem/pull/203).

### DIFF
--- a/lapce-app/Cargo.toml
+++ b/lapce-app/Cargo.toml
@@ -51,8 +51,7 @@ sled = "0.34.7"
 bytemuck = "1.14.0"
 tokio = { version = "1.21", features = ["full"] }
 futures = "0.3.26"
-clipboard = "0.5.0"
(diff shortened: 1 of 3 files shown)


lapce-app/Cargo.toml
```diff
@@ -52,8 +52,7 @@ bytemuck = "1.14.0"
 tokio = { version = "1.21", features = ["full"] }
 futures = "0.3.26"
 clipboard = "0.5.0"
```
panekj, line 54: clipboard is not removed

alexpyattaev:
Hey, this appears to work correctly. Thank you! It did crash on the very first startup though. But I am quite sure this is unrelated.

PR #2820, opened 2023-11-28
fix: save without formatting now calls save.

D-T-666:
Along with a few other `FocusCommand`s (`ToggleCaseSensitive`, `GlobalSearchRefresh`, `SearchInView`, and 23 more), `SaveWithoutFormatting` was not being handled. This pr adds handling for `SaveWithoutFormatting` and sets the catch-all match arm to be `_ => unimplemented!()`

### DIFF
--- a/lapce-app/src/editor.rs
+++ b/lapce-app/src/editor.rs
@@ -756,6 +756,9 @@ impl EditorData {
             FocusCommand::Save => {
                 self.save(true, || {});
             }
+            FocusCommand::SaveWithoutFormatting => {
+                self.save(false, || {});
+            }
             FocusCommand::InlineFindLeft => {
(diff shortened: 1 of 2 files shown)


lapce-app/src/editor.rs
```diff
@@ -784,7 +787,7 @@ impl EditorData {
                     self.common.find.replace_focus.set(true);
                 }
             }
-            _ => {}
+            _ => unimplemented!(),
```
panekj, line 790: this will crash editor

panekj (changes requested):

lapce-app/src/editor.rs
D-T-666: Oh okay. Fixed it!

PR #2836, opened 2023-12-07
Doc backend

MinusGix:

This PR is part of extracting the editor code into a crate for users of Floem.
Goals:
- Make `Document`/`Editor`/etc. less dependent on Lapce-specific behaviors
  - Allowing using whatever 'backend' of saving/loading files, such as purely locally
  - Syntax highlighting
  - Phantom text
 - Allow more testing & bench marking
   - Currently many pieces of the code are annoying to test
   - Ex: Movement would be a great place to test, but currently hard due to how closely it is linked to various other parts of the code (like `CommonData` + `EditorViewData`)
   - Ex: Editing the file is a great place to benchmark, with definite areas where it is slower than needed
- Ideally, make the code simpler!

The extraction will be done over multiple PRs, to avoid ending up with one massive PR and allowing discussion of changes since there are open design decisions.  
  
This PR specifically gives the `Document` a backend. Types that implement the `Backend` trait need to handle various actions like 'what should happen when we save the file', 'what is the phantom text for this line; while also handling "event"-like functions such as whatever actions need to be done when the content is initialized, updated, or a delta is applied..  
  
Specifically, I broke out various pieces of `Document` logic out from `Document` and onto `DocBackend`. `DocBackend` serves as the Lapce-specific behavior, which will stay within Lapce.  
Pieces moved to the backend:  
- LSP specific features (diagnostics, inlay hints, code actions, completion lens)
- Git history
- Syntax highlighting
  - Treesitter + LSP semantic highlighting are done from Lapce
  - This would allow library users to apply whatever form of highlighting they want, whether they have a TextMate parser or a quite absurd regex

This PR does not fully make `Document` independent, but does make it quite a bit closer to that.

Notes:
- I don't like `pre_update_init_content` but syntax highlighting needs to happen before the call to `on_update` I believe
- I don't like having to pass a custom `prev_unmatched` function to `Editor` instead of `Syntax`.
- I don't entirely like changing `LapceLanguage::comment_token` to return a `&'static str`. However, we're currently returning `&'static str`s elsewhere. If I did not do that, I'd either need to clone or use a callback to properly pass the `Syntax`'s comment token around
- I need to fix `Backend::on_update` taking treesitter edits since I don't want to force treesitter on users of `floem-editor`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1673,7 +1673,7 @@ checksum = "98de4bbd547a563b716d8dfa9aad1cb19bfab00f4fa09a6a4ed21dbcf44ce9c4"
 [[package]]
 name = "floem"
 version = "0.1.0"
-source = "git+https://github.com/lapce/floem?rev=0694f9672a10a49290af223cac3dc2ab53b1b1dd#0694f9672a10a49290af223cac3dc2ab53b1b1dd"
+source = "git+https://github.com/lapce/floem?rev=9dba15607de6d850e47a94f27cd2909967d9e732#9dba15607de6d850e47a94f27cd2909967d9e732"
 dependencies = [
  "bitflags 2.4.0",
  "copypasta",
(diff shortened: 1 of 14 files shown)


panekj:
> This PR needs [lapce/floem#216](https://github.com/lapce/floem/pull/216)

You can bump floem dep?

PR #2862, opened 2023-12-25
working snap

kz6fittycent:
panekj has published a snap but it's not functioning. This snap will work with a GUI, etc. It should still probably remain in the edge channel, but it's working.


panekj:
The decision was that we don't support snap

kz6fittycent:
I can provide that support, since I've got it working. If you want to make me a collaborator on your snapcraft account, I can get that going for you.

panekj:
> I can provide that support, since I've got it working.

Are you going to provide ongoing support for building/testing the snap as well as help snap users?

kz6fittycent:
Sure can. I've been doing this for a while.

.github/workflows/test-snap-can-build.yml
```diff
@@ -0,0 +1,25 @@
...
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+    
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [16.x]
```
panekj, line 14: Is this needed?

README.md
```diff
@@ -41,6 +41,11 @@ Lapce (IPA: /læps/) is written in pure Rust with a UI in [Floem](https://github
 You can find pre-built releases for Windows, Linux and macOS [here](https://github.com/lapce/lapce/releases), or [installing with a package manager](docs/installing-with-package-manager.md).
 If you'd like to compile from source, you can find the [guide](docs/building-from-source.md).
 
+You can also install the snap:
+```
+sudo snap install --edge lapce
+```
+
```
panekj, line 48: This should be in package manager doc

snap/snapcraft.yaml
```diff
@@ -0,0 +1,78 @@
...
+parts:
+  lapce:
+    source: https://github.com/lapce/lapce
+    source-type: git
+    plugin: rust
+    after: [cargo]
+
+    override-pull: |
+      snapcraftctl pull
+      snapcraftctl set-version "$(git describe --tags | sed 's/^v//' | cut -d "-" -f3)"
```
panekj, line 43: Lapce requires `RELEASE_TAG_NAME` to be set to correct version which can be `nightly` or git tag

snap/snapcraft.yaml
```diff
@@ -0,0 +1,78 @@
+name: lapce
+adopt-info: lapce
+summary: Lightning-fast And Powerful Code Editor
+description: |
+    Lapce (IPA: /læps/) is written in pure Rust with a UI in Floem. It is designed with Rope Science
+    from the Xi-Editor which makes for lightning-fast computation, and leverages Wgpu for rendering. 
+    More information about the features of Lapce can be found on the main website and user
+    documentation can be found on GitBook.
+
+base: core20
```
panekj, line 10: Would the newer `core22` base work? I'd rather keep all components on latest versions

snap/snapcraft.yaml
```diff
@@ -0,0 +1,78 @@
...
+    
+    override-pull: |
+      snapcraftctl pull
+      snapcraftctl set-version "$(git describe --tags | sed 's/^v//' | cut -d "-" -f2)"
+
+    build-packages:
+      - cargo
+      - rustc
+      - pkg-config
+      - librust-openssl-sys-dev
```
panekj, line 78: Do we have to build rust in snap? I'd rather grab rust from upstream like here: https://github.com/lapce/snapcraft-lapce/blob/c1555d285487e57b720fd249f9b64144610aa8f0/snap/snapcraft.yaml#L48

snap/gui/lapce.desktop
```diff
@@ -0,0 +1,12 @@
...
+Name=Lapce
+GenericName=Lapce
+Comment=Lightning-fast And Powerful Code Editor
+Exec=lapce
+Icon=${SNAP}/meta/gui/lapce.png
+Terminal=false
+Type=Application
+StartupNotify=false
+Categories=Application;
+Keywords=lapce;
```
panekj, line 12: Do we need separate desktop file for snap? I'd rather re-use what we already have in `extra/`

snap/gui/lapce.png
panekj, line 1: can't we use already present logo in repo?

.github/workflows/test-snap-can-build.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: snapcore/action-build@v1
+        id: build
+
+      - uses: diddlesnaps/snapcraft-review-action@v1
+        with:
+          snap: ${{ steps.build.outputs.snap }}
+          isClassic: 'false'
```
panekj, line 25: Maybe we can move that as a separate job in https://github.com/lapce/lapce/tree/master/.github/workflows/ci/yml ?

.github/workflows/test-snap-can-build.yml
kz6fittycent: Yes, otherwise you get a deprecation warning from GH.

snap/snapcraft.yaml
kz6fittycent: sure that's easy enough to change.

.github/workflows/test-snap-can-build.yml
kz6fittycent: We don't have to include these with the PR. They were intended to help you with managing the snap. These all can be omitted - including the other items above.

snap/gui/lapce.png
kz6fittycent: Well, it will need to be a .png and be 256x256.

.github/workflows/test-snap-can-build.yml
kz6fittycent: Also, I think it's fine to omit this from the PR if I'm managing the snap.

snap/snapcraft.yaml
kz6fittycent: If you look in this section of `parts`, I am building it from upstream:
```
  cargo:
    source: https://github.com/rust-lang/cargo
    source-type: git
    plugin: rust
 ```

panekj: I know, that's my point - can we not build it at all?

kz6fittycent:
On taking some inventory, I think maybe I'll take a pass on maintaining the snap. Feel free to use the snapcraft.yaml I've provided. I've got a number of other priorities to manage.

Issue #2822, opened 2023-11-28
No "Save As" and "Rename File" options


Unisa-Merni-392:
## Is your feature request related to a problem? Please describe.
For the life of me, I can't find a way to change the name of newly created file. Ut gets the name "Untitled-1" and there is no obvious way to change that.

On top of that, it never seems to get actually saved, even after "Save". So I can't rename it manually.

On top of that, neither existing "Save" command or CTRL+S keybinding seem to work.
I get the dialog window Save/Don't Save/Cancel  when I try to close the editor tab, but even if I click "Save", tab doesn't close, it retains its status as changed/not saved and there is no file on the disk. 

## Describe the solution you'd like
 - New command: "Save As"
 - New option somewhere to rename the file in project.
Perhaps by right-clicking on its open tab and/or its icon in left window/file manager etc

PR #2863, opened 2023-12-26
Implement path renaming from the file explorer

riverbl:
Implement functionality for renaming files and directories, accessible by right clicking the item in the file explorer and selecting "Rename...".

There was already some functionality in `lapce-proxy` for renaming paths, but there was no UI for using it. This PR adds a UI, and also makes some changes to the existing functionality:
* When renaming a file, any buffers affected by the rename are updated to use the new path.
* Non existing ancestor directories of the new path are created on rename.
* A new request, `TestCreateAtPath`, is added to allow querying whether a rename will succeed without performing it (not all failure reasons are covered - for example, it is not checked whether the user has permission to create a file in the target directory).

Note that this functionality can be used to move items as well as rename them - the user can rename `foo` to `../bar/foo` for example.

Relates to #2822.

### DIFF
--- a/lapce-app/src/command.rs
+++ b/lapce-app/src/command.rs
@@ -524,10 +524,13 @@ pub enum InternalCommand {
     OpenFileChanges {
         path: PathBuf,
     },
-    StartRenameFile {
+    StartRenamePath {
         path: PathBuf,
     },
-    FinishRenameFile {
+    TestRenamePath {
+        new_path: PathBuf,
+    },
+    FinishRenamePath {
         current_path: PathBuf,
         new_path: PathBuf,
     },
--- a/lapce-app/src/command.rs
+++ b/lapce-app/src/command.rs
@@ -524,6 +524,13 @@ pub enum InternalCommand {
     OpenFileChanges {
         path: PathBuf,
     },
+    StartRenameFile {
+        path: PathBuf,
+    },
+    FinishRenameFile {
+        current_path: PathBuf,
+        new_path: PathBuf,
+    },
     GoToLocation {
         location: EditorLocation,
     },
--- a/lapce-app/src/file_explorer/view.rs
+++ b/lapce-app/src/file_explorer/view.rs
@@ -5,7 +5,7 @@ use floem::{
     event::{Event, EventListener},
     peniko::Color,
     reactive::{create_rw_signal, RwSignal},
-    style::{AlignItems, CursorStyle, Style},
+    style::{AlignItems, CursorStyle, Position, Style},
     view::View,
     views::{
         container, container_box, label, list, scroll, stack, svg, virtual_list,
@@ -14,7 +14,7 @@ use floem::{
     EventPropagation,
 };
 use lapce_core::selection::Selection;
-use lapce_rpc::file::FileNodeViewData;
+use lapce_rpc::file::{FileNodeViewData, IsRenaming, RenameState};
 use lapce_xi_rope::Rope;
 
 use super::{data::FileExplorerData, node::FileNodeVirtualList};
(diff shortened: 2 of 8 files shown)


lapce-app/src/file_explorer/view.rs
```diff
@@ -1,30 +1,67 @@
...
     command::InternalCommand,
     config::{color::LapceColor, icon::LapceIcons},
     editor_tab::{EditorTabChild, EditorTabData},
-    panel::{position::PanelPosition, view::panel_header},
+    panel::{kind::PanelKind, position::PanelPosition, view::panel_header},
     plugin::PluginData,
-    window_tab::WindowTabData,
+    text_input::text_input,
+    window_tab::{Focus, WindowTabData},
 };
```
riverbl, line 31: I don't like having this function here. The problem is that `ERROR_LENS_ERROR_BACKGROUND` is transparent, so other filenames would be visible through the error box if it were used for the error box background. Just setting the alpha channel to 255 would result in the colour looking wrong.

Perhaps a new colour should be defined for the error box background, or maybe an existing dependency has a blend function that I didn't spot?

lapce-app/src/file_explorer/view.rs
```diff
@@ -1,30 +1,67 @@
...
 
+/// Blends `foreground` with `background`.
+///
+/// Uses the alpha channel from `foreground` - if `foreground` is opaque, `foreground` will be
+/// returned unchanged.
+///
+/// The result is always opaque regardless of the transparency of the inputs.
+fn blend_colors(background: Color, foreground: Color) -> Color {
```
riverbl, line 38: I don't like having this function here. The problem is that `ERROR_LENS_ERROR_BACKGROUND` is transparent, so other filenames would be visible through the error box if it were used for the error box background. Just setting the alpha channel to 255 would result in the colour looking wrong.

Perhaps a new colour should be defined for the error box background, or maybe an existing dependency has a blend function that I didn't spot?

lapce-app/src/command.rs
```diff
@@ -524,6 +524,16 @@ pub enum InternalCommand {
...
+    StartRenamePath {
+        path: PathBuf,
+    },
+    TestRenamePath {
+        new_path: PathBuf,
+    },
+    FinishRenamePath {
+        current_path: PathBuf,
+        new_path: PathBuf,
+    },
```
panekj, line 536: Why do we need those? It should be just `Rename` command

riverbl: `StartRenamePath` opens the text box in the file explorer to allow the user to enter a new path, but doesn't perform a rename.

`TestRenamePath` queries `lapce-proxy` to ask if the rename will succeed, but doesn't perform a rename. It's used to provide the user feedback on each keypress if, for example, they have entered a path that already exists.

`FinishRenamePath` attempts to perform the rename.

panekj: We don't need to ask proxy for validation, we already have tree in lapce-app

riverbl: The user may have entered a path like `foo/.././bar`, or a path containing symbolic links. `lapce-proxy` can check whether a path is valid for the rename by querying the actual filesystem, which is more robust than trying to handle edge cases without touching the filesystem.

riverbl: I've thought about this a bit more, and think that paths containing `.`, `..`, `//` and symbolic links could all be handled using the tree in `lapce-app`, and in any case if some edge case isn't handled then that may be okay, as `TestRenamePath` is only supposed to provide a best effort guess of whether a rename will succeed, not a guarantee.

I don't think that renames to paths outside the workspace (for example, renaming `foo` located at workspace root to `../foo`) could be validated without querying `lapce-proxy` though, as the tree only knows about paths in the workspace.

panekj: Moving files outside workspace should be forbidden operation.

Issue #2731, opened 2023-10-31
Keyboard shortcuts wrong keyboard layout detection


RichardFevrier:
## Lapce Version

nightly-b4a6ec5

## System information

Fedora Silverblue 38 Hyprland

## Describe the bug

Using a non QWERTY keyboard resulting in wrong keys detection in the Keyboard Shortcuts section.

E.g. 
Remapping `Ctrl+z` is detected as `Ctrl+w` for undo.

In the rest of the editor everything seems working correctly even pressing `Ctrl+z` really performs undo action (after remapping, read my paragraph below).

Why it is a really problem?
Because by default `Ctrl+z` is mapped to closing the file like a `Ctrl+w` would do... So non QWERTY keyboard users have to remap all non equivalent default shorcuts.

zadjadr:
This might be a hyprland issue, as far as I know: wayland will set the keyboard settings via [xorg](https://wiki.archlinux.org/title/Xorg/Keyboard_configuration).

In my case, hyprland sets the layout correctly (Z is where it physically is), but reports the "us" layout to xorg:

```
❯ setxkbmap -query
WARNING: Running setxkbmap against an Xwayland server
rules:      evdev
model:      pc105
layout:     us
```

I have no workaround, yet.

RichardFevrier:
Read carefully your output `WARNING: Running setxkbmap against an Xwayland server`
XWayland, not wayland.
Nothing related to hyprland, it was working properly in version 0.2.8 this is a regression.

Mforcen:
Keybindings do not work properly also in non-us keyboard layouts. For example I have the es-ES layout and I have to use the keys in the us layout for commands to be launched.

joshkamm:
I'm noticing a similar experience using DVORAK layout on a laptop's QWERTY hardware keyboard on Windows 10, where in version 0.3.0 the Ctrl key bindings are based on their hardware QWERTY positions while in version 0.2.8 they were based on their DVORAK positions.

P.S. Thanks to everyone who volunteers their time to this project. I recently came across it and it looks cool!

LuckyTurtleDev:
> I'm noticing a similar experience using DVORAK layout on a laptop's QWERTY hardware keyboard on Windows 10.

Same with koy on QWERTZ at arch linux.

An overview of some shortcuts:
strg + h save all
strg + ä copy
strg + ü insret
strg + v undo
strg + x redo
strg + a save

This make lapce is unusable for me.

[...]

PR #2874, opened 2023-12-31
Another approach to keyboard input

RobertPietraru:
# The problem
I am using ```Ubuntu``` and I have remapped my ```Escape``` and ```CapsLock``` keys, but the editor doesn't see the remap.

## The cause
This happens because we are using ```floem::keyboard::PhysicalKey```, which has no way of knowing about my remap.

## Additional problems (Not mine, but possible)
A person using Dvorak might expect the vim keybindings to change, the same way it happens in the actual editors (Vim, NeoVim), which is not the case.

## The solution
Replace ```floem::keyboard::PhysicalKey``` with ```floem::keyboard::Key```

Here is a summary of the changes I've made:
- in the toml files I replaced "shift+letter" with "CAPITAL_LETTER"
- I modified the ```keyboard_picker_view``` so that it replaces "Shift+letter" with "CAPITAL_LETTER" (removing the shift)
 ```Shift+0``` is not always ```)``` depending on the layout, so it makes a more sense not to represent it as a combination, but as the actual character.
- Also, it makes sure that commands like "Ctrl+shift+tab" are not affected by this 
- I made the character representation (Modifier1+Modifier2+key) case sensitive only for character keys

I believe this is a better system, but there might be some things I am missing.

## Related issues
#2731 Keyboard shortcuts wrong keyboard layout detection

! This is my first time working on a Rust project (and making an open source code contribution), so if there are any problems with it, please tell me!


dzhou121:
Thanks for this PR!

But this approach will be a step back to how Lapce was dealing with keyboard shortcuts. So this PR changed the way from textural representable of a key to scan code. https://github.com/lapce/lapce/pull/2634

Ideally we'd like to be able to detect the current keyboard layout set and adapt to it accordingly. However we're waiting for the support of keyboard layout from upstream. https://github.com/rust-windowing/winit/issues/2678

Feel free to add your thoughts in this discussion https://github.com/lapce/lapce/issues/2636

RobertPietraru:
In that case should I close the pull request?

dzhou121:
> In that case should I close the pull request?

I'm afraid so. Don't feel discouraged though. Thanks for the effect anyways I know Lapce is unusable for non standard keyboard layout users at the moment.

PR #2877, opened 2024-01-02
fix: wsl

panekj:


Egg12138:
"connect WSL" still failed to do anything in v0.3.1

panekj:
> "connect WSL" still failed to do anything in v0.3.1

because it's not fixed in v0.3.1

PR #2911, opened 2024-01-06
Standardizes command messages

thiagola92:
Added `#[strum(message = "MESSAGE")]` to all shortcuts commands and changed all to start with capital letter following by lower case letters

### DIFF
--- a/lapce-core/src/command.rs
+++ b/lapce-core/src/command.rs
@@ -14,39 +14,56 @@ use crate::movement::{LinePosition, Movement};
     IntoStaticStr,
 )]
 pub enum EditCommand {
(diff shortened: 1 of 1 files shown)


panekj:
We use capital letters everywhere, so I would like it to stay that way

thiagola92:
> We use capital letters everywhere, so I would like it to stay that way

Sorry, I found two styles in the file so I made a guess

panekj:
There are conflicts to resolve in `lapce-core/src/command.rs`

MinusGix:
That is due to the move of the editor into floem, with the file now located at: https://github.com/lapce/floem/blob/main/editor-core/src/command.rs

PR #2915, opened 2024-01-10
Fixed the issue where the plugin folder couldn't be deleted when unin…

jm-observer:
Fixed the issue where the plugin folder couldn't be deleted when uninstalling a plugin, due to the WASI instance not being dropped.

### DIFF
--- a/lapce-proxy/src/plugin/mod.rs
+++ b/lapce-proxy/src/plugin/mod.rs
@@ -4,6 +4,7 @@ pub mod lsp;
 pub mod psp;
 pub mod wasi;
 
+use std::time::Duration;
 use std::{
     borrow::Cow,
     collections::HashMap,
@@ -13,7 +14,6 @@ use std::{
         atomic::{AtomicBool, AtomicU64, AtomicUsize, Ordering},
         Arc,
     },
-    thread,
 };
 
 use anyhow::{anyhow, Result};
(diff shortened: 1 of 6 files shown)


lapce-proxy/src/plugin/mod.rs
```diff
@@ -1375,6 +1375,8 @@ pub fn remove_volt(
                 .volt_removing(volt.clone(), "Plugin Directory not set".to_string());
             anyhow::anyhow!("don't have dir")
         })?;
+        // Waiting for the instance of WASI to be dropped.
+        std::thread::sleep(std::time::Duration::from_secs(1));
```
dzhou121, line 1379: It would be much better to wait for the wasi instance to be dropped in `stop_volt()` I think.

My idea is in `PluginServerRpcHandler::shutdown()`, you'd wait for a new event like `PluginServerRpc::Exited` which was sent from the wasi msg sending loop break. (wasi.rs line 464)

jm-observer: Okay, let me optimize it tomorrow.

jm-observer: I've reorganized the code and have a few questions. 1. Can `PluginServerRpcHandler::mainloop` not delete the plugin folder because it can't access class `PluginCatalogRpcHandler`? 2. After calls `PluginServerRpcHandler::shutdown()`, `PluginServerRpcHandler::mainloop` can no longer accept new events, nor can it send notifications and requests to `PluginCatalogRpcHandler`? in addition, I think the most clear and straightforward way is for `ProxyRpcHandler` to send a close request directly, and then delete the folder in the callback. What do you think?

dzhou121: You are right. Deleting in a callback looks to be the best solution.

PR #2921, opened 2024-01-15
Fix unicode path handling on windows LSPs

### LINKED ISSUE #2920: Unicode path handling not working on Windows
## Lapce Version

nightly-9bddebd

## System information

Windows 11

## Describe the bug
Error paths reported through the LSP have escaped HTML characters if the original path has non-ascii characters. This causes incorrect error navigation.

## Additional information

Path reported in error should be C:/Users/þistle/bincode/src

image

My best guess is that the LSP requires HTML escaping of paths and lapce needs to unescape them.

thistle-witch:
Closes #2920 

Tested with Rust analyzer. `to_file_path()` properly converts `file:///c:/...` style URIs into properly drive rooted windows paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Bug Fixes
 - [#2779](https://github.com/lapce/lapce/pull/2779): Fix files detection on fresh git/VCS repository
+- [#2920](https://github.com/lapce/lapce/pull/2920): Fix unicode path handling on windows LSPs
 
 ## 0.3.1
 
--- a/lapce-app/src/proxy.rs
+++ b/lapce-app/src/proxy.rs
@@ -143,24 +143,6 @@ impl CoreHandler for Proxy {
     }
 }
 
-// Rust-analyzer returns paths in the form of "file:///<drive>:/...", which gets parsed into URL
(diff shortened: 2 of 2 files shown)


panekj (changes requested):
This breaks lapce

thistle-witch:
I'm not sure what you mean by "this breaks lapce"? I tested against the relevant URLs. The handling seemed to be specific to windows file URIs which I tested. No other code was affected.

panekj:
Using `to_file_path()` makes lapce stack overflow

thistle-witch:
I'm curious how you got that result. I ran a build of this myself.

thistle-witch:
If you have a test project or example URI that causes the stack overflow I can try to find out what is going on.

panekj:
The test code is what `cargo init` gives, typing and deleting `println!` causes overflow but it doesn't happen in release mode

thistle-witch:
It would be helpful if you could provide the exact URI that is causing this issue for you. I have just run a debug build with a `cargo init` project and do not get the same results at all. To confirm, which version of windows are you using? Its entirely possible this stack overflow is caused by stack size differences.

panekj:
Fixed via f2355aa0b8839aacc5fa684db2c8b75b5258e251

PR #2949, opened 2024-02-13
feat: focus search input when opening settings

ArFnds:

### DIFF
--- a/lapce-app/src/settings.rs
+++ b/lapce-app/src/settings.rs
@@ -471,7 +471,7 @@ pub fn settings_view(
         }),
         stack((
             container({
-                text_input(search_editor, || false)
+                text_input(search_editor, || true)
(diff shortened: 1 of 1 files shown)


panekj:
What's the reasoning for this change?

ArFnds:
It's mainly UX:
- when a user open the settings using `Ctrl + ,`, he can type to search the setting without using the mouse to select the field
- when a user come on the settings page using the mouse, he can directly type

lapce-app/src/settings.rs
```diff
@@ -471,7 +471,7 @@ pub fn settings_view(
         }),
         stack((
             container({
-                text_input(search_editor, || false)
+                text_input(search_editor, || true)
```
dzhou121, line 474: This doesn't need to be true. 

The `request_focus` down will make `keyboard_focus` to be `true`

PR #2955, opened 2024-02-15
Feat: add `files_exclude` option (Fixes #811)

### LINKED ISSUE #811: [feature] blacklist for file explorer tab

I-Info:

This PR fixes #811 by providing a `files_exclude` string option which using blob patterns to blacklist files and directories.

Currently using default value of `"**/{.git,.svn,.hg,CVS,.DS_Store,Thumbs.db}"` (Referenced from VSCode)

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [#1277](https://github.com/lapce/lapce/pull/1277): Error message prompted on missing git user.email and/or user.name
 - [#2910](https://github.com/lapce/lapce/pull/2910): Files can be compared in the diff editor
 - [#2918](https://github.com/lapce/lapce/pull/2918): Allow searching the shortcuts overview by shortcut (e.g. "Ctrl+P") or when condition (e.g. "list_focus")
+- [#2955](https://github.com/lapce/lapce/pull/2955): Using glob patterns to hide files or directories in explorer
 
 ### Bug Fixes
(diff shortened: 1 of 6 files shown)


imgurbot12:
@I-Info  @dzhou121 thank you for this PR, this is a step in the right direction, but I don't believe this solves https://github.com/lapce/lapce/issues/811.

The issue reference specifically is looking to include anything already in a `.gitignore` file at the root of the project folder. This adds another location to ignore specified items but does not provide an option to ignore items already specified in the .gitignore

Is there another PR that already addressed this that I'm maybe missing?

panekj:
> thank you for this PR, this is a step in the right direction, but I don't believe this solves #811.
> 
> The issue referenced specifically is looking to include anything already in a `.gitignore` file at the root of the project folder. This adds another location to ignore specified items but does not provide an option to ignore items already specified in the .gitignore

I believe it does solve the issue, as it wasn't about `.gitignore` at all. It was only listed as a "nice to have" option

imgurbot12:
@panekj I have opened a new issue as requested to hopefully reinvigorate the now closed PR working on `.gitignore` support.

PR #2999, opened 2024-02-21
chore(deps): bump floem to `341afd`

I-Info:


lapce-app/src/main_split.rs
```diff
@@ -1813,10 +1813,10 @@ impl MainSplitData {
                                     let main_split = main_split.clone();
                                     let child = child.clone();
                                     let local_main_split = main_split.clone();
-                                    if let Some(file) = file {
+                                    if let Some(mut file) = file {
                                         main_split.save_as(
                                             doc.clone(),
-                                            file.path,
+                                            file.path.pop().unwrap(),
```
panekj, line 1819: Unhandled panic

I-Info: Emm, I think it is logically safe to unwarp() results at those places.

```rust
pub fn open_file(
    options: FileDialogOptions,
    file_info_action: impl Fn(Option<FileInfo>) + 'static,
) {
    let send = create_ext_action(
        Scope::new(),
        move |(path, paths): (Option<PathBuf>, Option<Vec<PathBuf>>)| {
            if paths.is_some() {
                file_info_action(paths.map(|paths| FileInfo {
                    path: paths,
                    format: None,
                }))
            } else {
                file_info_action(path.map(|path| FileInfo {
                    path: vec![path],
                    format: None,
                }))
            }
        },
    );
    .....
```
```rust
pub fn save_as(options: FileDialogOptions, file_info_action: impl Fn(Option<FileInfo>) + 'static) {
    let send = create_ext_action(Scope::new(), move |path: Option<PathBuf>| {
        file_info_action(path.map(|path| FileInfo {
            path: vec![path],
            format: None,
        }))
    });
```

https://github.com/lapce/floem/blob/341afd255f98ca431e4f615864daa4dbcc057501/src/action.rs#L154-L160
https://github.com/lapce/floem/blob/341afd255f98ca431e4f615864daa4dbcc057501/src/action.rs#L108-L127

Certainly, unexpected panic may happen when the implementation changed in the floem upstream.
Should I use `expect()` to replace?

dzhou121: `expect()` still panics. I would probably have `if let Some(path) = file.path.first()` for it.

Issue #3003, opened 2024-02-22
Rename in File Explorer duplicates keystrokes 


ethereumdegen:
## Lapce Version

The Lapce version you are using, which can be found in "About Lapce" at the top right settings icon. 
nightly-cea8a10 

## System information

the operating system used, including its version, e.g. Windows 10, Ubuntu 18.04
Linux mint 

## Describe the bug
Rename in File Explorer duplicates keystrokes

Issue #2953, opened 2024-02-14
Right click menu doesn't open in 0.3.1


Jakraes:
## Lapce Version

0.3.1

## System information

Windows 10

## Describe the bug

Right clicking on the file explorer doesn't make the right click menu pop up (it's working like it should on 0.2.8), which means the only way to create/delete/rename files is with the command palette

## Additional information

Nothing to add

dancespiele:
the same issue in Linux

PR #3026, opened 2024-02-28
Remaining file explorer context menu entries

### LINKED ISSUE #2873: No menu in the file explorer.
## Lapce Version
0.3.1

## System information

Windows 10 

## Describe the bug

There is no menu opened with a way to make new files, if you right click in the File Explorer

## Additional information

I did not even know there was a menu, until I searched if someone requested such a menu as a feature. And saw a different bug about the menu. 

This makes this editor unusable for me.

MinusGix:
This PR is primarily for adding back the file explorer context menu items.  
Fixes #2873  #2953   
Commits:  
- Add some comments, use `let Some(x) = ... else { return; }` to avoid some over-indented code
- Remove unused `id` field.  
The `id` field was not used anywhere.  
While it could have been useful in some cases (ex: logic that wants to simply know when the file explorer was updated), I'm not sure it would have been that useful as it is changed before the updating of the explorer. It would be simple enough to add back in later, with whatever form is useful when actually needed.  

-  Initial adding of missing context menu commands
- Main commit, actually implement new file / new directory / duplicate
  - Also fixes a bug we apparently had where typing in the input box would cause multiple characters to appear
- Simplify the state machine logic to share more between the state
- Simplify the creation of children
  - Most editors have so if you create a new file the input box shows *after* all the folders. And if you are creating a new directory then it should show *before* all the folders. I wonder if there's a way to make that logic less ugly.
- Make so you can right-click anywhere in file explorer, which is nicer than requiring a file/dir to be right-clicked

Adds:
- New file
- New directory
- Reveal in file explorer
- Duplicate
- Move directory to trash
- Copy path
- Copy relative path
- Refresh

Current Issues:
- I'm getting a problem where the context menu does not show all the entries. I think there's a max size for the context menu that causes problems... I get this on normal floem as well so it is a problem there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - [#2910](https://github.com/lapce/lapce/pull/2910): Files can be compared in the diff editor
 - [#2918](https://github.com/lapce/lapce/pull/2918): Allow searching the shortcuts overview by shortcut (e.g. "Ctrl+P") or when condition (e.g. "list_focus")
 - [#2955](https://github.com/lapce/lapce/pull/2955): Using glob patterns to hide files or directories in explorer
+- [#3026](https://github.com/lapce/lapce/pull/3026): Add missing file explorer context menu entries
 
 ### Bug Fixes
 - [#2779](https://github.com/lapce/lapce/pull/2779): Fix files detection on fresh git/VCS repository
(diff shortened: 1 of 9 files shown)


MinusGix:
Also fixes #3003

PR #3058, opened 2024-03-10
Updated platform-specific deps to accommodate linux RFD issues

njrogie:
On linux, using Floem's `rfd/tokio` feature would cause major problems w/ RFD (referenced https://github.com/lapce/floem/issues/365).  

This patch causes lapce to build with `rfd-async-std` instead of `tokio` on Linux, which enables it to access the xdg endpoints provided by `rfd`. All other platforms should be unchanged.

Tested local builds on Windows 11 and Ubuntu 22.04.4 LTS (Lubuntu/LXQt)

### DIFF
--- a/lapce-app/Cargo.toml
+++ b/lapce-app/Cargo.toml
@@ -43,7 +43,6 @@ tracing.workspace = true
 tracing-subscriber.workspace = true
 tracing-appender.workspace = true
 url.workspace = true
-
 pulldown-cmark = "0.9.1"
 Inflector = "0.11.4"
 open = "5.0.1"
@@ -53,8 +52,6 @@ sled = "0.34.7"
 bytemuck = "1.14.3"
 tokio = { version = "1.36", features = ["full"] }
 futures = "0.3.26"
-floem = { git = "https://github.com/lapce/floem", rev = "764857c98dfdc2beb10966636da4cf79b677a089", features = ["serde"] }
-# floem = { path = "../../workspaces/floem", features = ["serde"] }
 config = { version = "0.13.2", default-features = false, features = ["toml"] }
 structdesc = { git = "https://github.com/lapce/structdesc" }
 base64 = "0.21.7"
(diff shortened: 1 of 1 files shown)


panekj:
The `tokio` dep is used because we use `tokio` runtime, I specifically made that change to remove `async-std` from Lapce

njrogie:
My terminology was incorrect. I wasn't removing `tokio` from the Lapce dependencies, i'm using `async-std` instead of `tokio` in the RFD dependency because it won't properly spawn xdg windows when using `tokio` on my Ubuntu build. The error it gives me when i try to use the File > Open Folder button is `thread '<unnamed>' panicked at /home/<username>/.cargo/registry/src/index.crates.io-6f17d22bba15001f/zbus-4.1.2/src/abstractions/executor.rs:189:27:
there is no reactor running, must be called from the context of a Tokio 1.x runtime`

njrogie:
If we want to find a way to get it to work with the `tokio` feature of RFD on ubuntu, i can take a look at that

panekj:
> My terminology was incorrect. I wasn't removing `tokio` from the Lapce dependencies,

You are removing `tokio` from `rfd` and adding `async-std`, this puts 2 different async runtimes in Lapce for no reason

njrogie:
There's lots of usages of `tokio` in the Cargo files but no usages of `tokio` in the codebase (from a quick `grep -r tokio`). Removing the `tokio` dep from the `lapce-app/Cargo.toml` has no effect on compilation either. Unless i'm misunderstanding `tokio` (entirely possible, i'm newer to this ) I don't see the multi runtime scenario you're describing.

As I understand it there appears to already be action taken in floem (https://github.com/lapce/floem/commit/40ab0df2ba86651a811607d6ffc9d5ccce9befe1) which should solve the issue this PR was attempting to work around. once that version of floem is pulled in, I suppose this PR is defunct

panekj:
> I don't see the multi runtime scenario you're describing.

that's not how that how works, the runtime is used no matter if we use it in our codebase or not, since it is a dependency

PR #3088, opened 2024-03-20
source-build fix for Fedora and add-path instructions for Ubuntu and Fedora

lineality:

Adding step to install pearl for fedora, and adding instructions for add-path etc. so Lapce can be used and called from cli normally.

https://github.com/lineality/fork_lapce/blob/geo_build_fix/docs/building-from-source.md

### DIFF
--- a/docs/building-from-source.md
+++ b/docs/building-from-source.md
@@ -18,7 +18,7 @@ home/YOURNAME/lapse/target/release/lapce
 ```
 
 2.3  Create a symbolic link:
-   - Open a terminal and run the following command to create a symbolic link from the executable to the chosen directory:
+- Open a terminal and run the following command to create a symbolic link from the executable to the chosen directory:
      ```bash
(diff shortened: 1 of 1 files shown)


docs/building-from-source.md
```diff
@@ -7,13 +7,79 @@ It is easy to build Lapce from source on a GNU/Linux distribution. Cargo handles
...
+     ```
+
+2.5  Refresh the shell:
+- Run the following command to reload the shell configuration:
+     ```
+     source ~/.bashrc
+     ```
+
+- Now you can call "lapce ." or "lapce --version" from cli
+
```
panekj, line 38: Please remove this, it's unnecessary since cargo already manages the binary

docs/building-from-source.md
```diff
@@ -7,13 +7,79 @@ It is easy to build Lapce from source on a GNU/Linux distribution. Cargo handles
...
```
panekj, line 82: As above

Issue #3044, opened 2024-03-07
Visually integrate file status into the file tree view


KillingSpark:
## Describe the solution you'd like

This is a feature of vscode that I think all editors with a builtin file tree viewer should have: Visually integrate the status of the files into the file tree. Examples include:

* the git status of the file (changed, untracked, staged, ...)
* the LSP status of the file (errors, warnings)

This should propagate upwards through the directories

## Describe alternatives you've considered

A common alternative (also currently implemented in lapce, but many other editors have the same solution) is having separate views for this. 

* the file tree (to open editors) 
* the errors/warnings view that displays the files as a list with the errors contained in them (and to open editors)
* the git view that displays the changed files as a list (open diffs and commiting)

Obviously the separated views have their usecases too, I don't think the visual integration of the latter two into the file tree would be able to replace them.

## Additional context

The LSP integration would be especially useful to me when I am refactoring code or changing core concepts in a library to see how many locations would be affected by a change I am considering.

The git integration would be useful to me when I am cleaning up code with many small changes and need to decide when a changeset is getting too large so I should do a commit.

Having this integrated into one view would make it easier to do these things compared to more views necessitating switching the view (git view) or using up more screen realestate (LSP view).

I would be willing to do the legwork for this feature but before I get familiar with the codebase and start work I wanted to know if this is both wanted by the project and feasible.

PR #3096, opened 2024-03-24
Add text color if file is changed according to source control

MinusGix:
image

(Re)implements  color highlighting if the file has been changed/renamed/added according to source control.

Implements part of #3044


dzhou121:
There's a conflict on changelog now.

PR #3130, opened 2024-03-30
fix settings page navigation issue

### LINKED ISSUE #2759: [Bug Report] Settings page navigation exceptions
## Lapce Version

0.3.0 

## System information

Windows 11

## Describe the bug
Sometimes, when we click on the navigation on the left side of the settings page, the settings do not jump to the specified position correctly. Especially the ones that are lower down. This is episodic, and the pattern of its appearance is not yet known.
## Additional information

For example, click the tab of "Terminal Settings", it wont jump to the settings about terminal. 
image

Long0x0:
Fixes #2759.

### DIFF
--- a/lapce-app/src/settings.rs
+++ b/lapce-app/src/settings.rs
@@ -487,12 +487,8 @@ pub fn settings_view(
             .style(|s| s.padding_horiz(50.0).padding_vert(20.0)),
             container({
                 scroll({
-                    virtual_stack(
-                        VirtualDirection::Vertical,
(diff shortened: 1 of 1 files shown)


panekj:
This still doesn't jump correctly to clicked elements

Long0x0:
hmm, have checked on windows and ubuntu, seems to be no problem.
which platform do you use?

panekj:
Windows

panekj:
Probably cargo being dumb again, it works after clean build

PR #3146, opened 2024-04-06
fix: update queries, use helix queries, add attribution

panekj:


dzhou121:
It didn't work for me when I tried to use helix queries though.

panekj:
Which queries? Most of them are already sourced from Helix repo.

dzhou121:
I mostly used the ones in the original tree-sitter-*

panekj:
I get same results for current and helix queries (since helix also derives queries from tree-sitter-* when possible).

Issue #3179, opened 2024-04-19
Editor: Not Supported


ydong08:
## Lapce Version
Lapce nightly-6fa0f4c
Windows 10 Pro 22H2
Remote Server: Debian 4.19  x86_64 GNU/Linux

## Describe the bug
Lapce under SSH remote development mode, sometimes, i encounter the following problem `Not Supported` when opening some source code files，however，other files are opened successfully.
In my case, `test_file_ns.c` are opened with `Not Supported`:
![not_support](https://github.com/lapce/lapce/assets/11952321/aa7e31f3-4e08-41b5-9be5-de7902cd1f0b)

Another file `test_file_cr.c` is all right:
![correct](https://github.com/lapce/lapce/assets/11952321/e2e3ab69-0f88-497d-bcbc-12fdad0833f4)

I check the difference between files on Remote Server as follow steps:
```
$ file test_file_cr.c
test_file_cr.c: C source, ASCII text
$ vim test_file_cr.c
:set fileencoding?
fileencoding=utf-8 
:set bomb?
nobomb

$ file test_file_ns.c
test_file_ns.c: C source, ISO-8859 text, with CRLF line terminators
$ vim test_file_ns.c
:set fileencoding?
fileencoding=euc-cn 
:set bomb?
nobomb
$ 
```

Please let me know if you need any more information.

panekj:
It's because the file contains characters that are not Unicode, from what I see it's ISO-8859-1 encoding

panekj:
Do you know which encoding it is exactly? I'm getting mixed results..
![image](https://github.com/lapce/lapce/assets/50457605/3ec34de1-12c4-4f88-b877-cfb840fc3b5c)

ydong08:
@panekj 
I get the encoding is `euc-cn` in VIM.

Thanks for implementing the encoding feature!

I merge feat/encoding branch into master locally, and then compile a new lapce. 
When i run the new lapce and start a ssh connection, remote server also starts a download process `wget https://github.com/lapce/lapce/releases/download/nightly/lapce-proxy-linux-x86_64.gz`. 
And then I replace the binary `~/.local/share/lapce-debug/proxy/lapce` with a new lapce-proxy compiled on remote server with encoding feature, the remote server is still downloading from github.

What should i do to use the new lapce-proxy compiled by myself.

panekj:
It's not possible to replace debug/nightly proxy.

ydong08:
[...]

PR #3181, opened 2024-04-19
feat: implement reading files in different file encodings

panekj:


dzhou121:
Do we need to convert it back to original encoding when saving the file?

panekj:
Yes, I just haven't implemented it yet

panekj:
Fixes #3179

panekj:
Also, to implement palette for selecting encodings so user can choose custom decoding/encoding for the file.

PR #3295, opened 2024-06-01
cargo-deny: init

cjbayliss:
* this add support for checking licenses with cargo deny check
* cargo deny can also check for security issues and unmaintained/yanked crates

see also: https://github.com/lapce/structdesc/pull/2

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,15 @@
+[licenses]
+# use SPDX short identifiers
+allow = [
+    "Apache-2.0",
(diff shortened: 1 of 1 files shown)


panekj:
I'd rather not add whatever new cargo plugin there is

cjbayliss:
Fair enough. :)

panekj:
actually this might be useful for denying features/crates, just need to figure out how to use it later and add CI job

PR #3309, opened 2024-06-08
disable selection on editor tabs

panekj:

### DIFF
--- a/lapce-app/src/app.rs
+++ b/lapce-app/src/app.rs
@@ -663,6 +663,7 @@ fn editor_tab_header(
                             .unwrap_or(true),
                         |s| s.font_style(FontStyle::Italic),
                     )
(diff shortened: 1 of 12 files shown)


lapce-app/src/app.rs
```diff
@@ -660,6 +660,9 @@ fn editor_tab_header(
                     .apply_if(tab_close_button_style == TabCloseButton::Off, |s| {
                         s.margin_right(15.0)
                     })
+                })
+                .style(move |s| {
+                    s.apply(LabelCustomStyle::new().selectable(false).style())
```
dzhou121, line 665: It's shorter to do 

```rust
.style(|s| s.set(Selectable, false))
```

panekj: replaced with `.selectable(false)`

PR #3310, opened 2024-06-08
Add debug names, trim extra containers

jrmoulton:

I also trimmed a few extra containers that were there from when floem couldn't stack styles. I checked that both sets of styles weren't conflicting and removed the extra container. There are probably quite a few more of these that can be removed.

Screenshot 2024-06-08 at 3 00 55 AM


panekj:
Can you pull latest commit, accidentally broke CI for PRs

PR #3312, opened 2024-06-09
Remove containers

### LINKED ISSUE #2883: Allow window tabs to have their close button removed just like editor tabs

Tab close buttons aren't useful when you only use keyboard to close tabs, but they take space and can inadvertently close tabs on mouse dragging

Add a `win-tab-close-button = "Off"` setting similar to `tab-close-button = "Off"` to remove the close button on window tabs

Don't know of any

## Additional context

N/a

jrmoulton:

This:
- Removes more of the extra containers
- Adds more debug names
- Removes drawing the scroll bar background in the editor view so that the editor view is consistent with the other tab views. 

I've also made an opinionated change to the design of the border on editor tabs. They now fill the full height of the tab bar. This looks more clear and less busy to me, but that's just my opinion and if people actually prefer the old way I can make it configurable. 
image

Also fixes #2883 which had been previously implemented but was not responding the config changes.

### DIFF
--- a/lapce-app/src/app.rs
+++ b/lapce-app/src/app.rs
@@ -696,6 +696,8 @@ fn editor_tab_header(
 
             let tab_style = move |s: Style| {
                 s.items_center()
+                    .justify_center()
+                    .height_full()
                     .border_left(if i.get() == 0 { 1.0 } else { 0.0 })
                     .border_right(1.0)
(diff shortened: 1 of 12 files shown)


dzhou121:
Could you please make the editor tab border configurable?

PR #3313, opened 2024-06-09
Make editor tab separator height configurable

jrmoulton:


jrmoulton:
Yeah it's ready. The close was accidental but this is good to go

Issue #2884, opened 2024-01-03
UI customization: allow changing window tab height


eugenesvk:
## Is your feature request related to a problem? Please describe.

The window tab height takes too much vertical space

## Describe the solution you'd like

I'd like it to serve only informational purpose, to just fit the height of the text without any extra padding, so would be great to have `win-header-height` config

## Describe alternatives you've considered

I've tried `header-height = 55`, but that doesn't seem to have any effect

## Additional context

N/a

PR #3322, opened 2024-06-11
Improve editor tabs

jrmoulton:

- If the editor tab close button is set to off, the file icon will now show the warn color to represent the pristine state
- Editor tab content has a tooltip that will show the full file path
- tabs now have cursor Pointer and apply HOVER_BACKGROUND on hover

Also fixed:
- Editor tab height was not connected to the config


jrmoulton:
Fixes #2884

PR #3323, opened 2024-06-12
Make showing tab forward/back configurable

jrmoulton:


panekj:
Is there a reason for that setting?

jrmoulton:
Just that I'd like to be able to disable it. If it's not general enough to include I'll just keep it in my fork.

jrmoulton:
There are a few settings as well for disabling parts of the UI to simplify things that I want to add. But same thing for those, if they aren't general enough to include I can just maintain them in a fork

dzhou121:
I think let's "clutter" the config before we figure out a better way to do it.

panekj:
As I mention on Discord, I'd prefer for it to make it an `enum` option of Left, Right, Off for all buttons together (Forward/Back/New/Close)

jrmoulton:
> As I mention on Discord, I'd prefer for it to make it an enum option of Left, Right, Off for all buttons together (Forward/Back/New/Close)

Makes sense to me

jrmoulton:
I tried changing these configs to part of an enum but that also included adding the dropdown configuration which adds more clutter and it wasn't better

PR #3328, opened 2024-06-13
run workbench commands triggered in the editor

jrmoulton:

### DIFF
--- a/lapce-app/src/editor.rs
+++ b/lapce-app/src/editor.rs
@@ -2887,7 +2887,10 @@ impl KeyPressFocus for EditorData {
         }
 
         match &command.kind {
-            crate::command::CommandKind::Workbench(_) => CommandExecuted::No,
+            crate::command::CommandKind::Workbench(cmd) => {
+                self.common.workbench_command.send(cmd.clone());
+                CommandExecuted::Yes
+            }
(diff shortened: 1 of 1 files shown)


dzhou121:
workbench command should be bubbled up if not executed.

Which command did it not work in editor?

jrmoulton:
Specifically I was using, `palette`, `NextEditorTab`, `PreviousEditorTab`. 

From discord
> Having an issue with some code actions in response to keypresses in the editor. If I set `g n` to be next editor tab when in normal mode, when I press that combo nothing happens. Using that same keyboard combo I can trigger other actions and using the command palette I can move to the next editor tab. I can also use a keyboard shortcut like `Cmd+n` to go to the next tab. I'm not sure where the disconnect is

dzhou121:
should be fixed by f258bb782c55d0d06462192319f35c892060cc23

The falling back keypress handling logic was wrong.

jrmoulton:
Thanks! There is still a similar issue in there terminal pane. 

https://github.com/lapce/lapce/blob/f258bb782c55d0d06462192319f35c892060cc23/lapce-app/src/terminal/panel.rs#L149C13-L159C14

There are keyboard shortcuts (for me Ctrl-z) that don't work in the terminal panel because the terminal receives the character and then marks it as handled even through there are matching key matches. 

Would it make sense to only mark that as handled if the key match is none?

Issue #3275, opened 2024-05-19
terminal profiles broken again


maddes8cht:
## Lapce Version

nightly -11f36a7

## System information

Windows 10

## Describe the bug
With the current nightly, the formerly closed issue #3244 is back again and worse than before:
- Not only does the before working `settings.toml` i posted in #3244 stop working again (again it always opens powershell)
- in the settings, section terminal, in the `Terminal: default profile` there is only `default`available, the other 
configured terminals have disappeared.
- when i change any settings in the graphical interface, the settings.toml gets shot in the foot (in different ways, see aditional information)

## Additional information

The last item in my list is very inconsistent and not very reproducible, thus i don't have screenshots or .toml file content for this.
I experienced once that the automatically written settings.toml was garbaged in the way that almost all closing bracket like `]` , `}` and even closing `"` had vanished, causing lapce to be stalled when loading. 
I also experienced when changing the `core: Color Theme` setting that i got a `settings.toml`with only the setting for the `core: Color Theme`, all other settings had been deleted. This seems more reproducible.

panekj:
Can't reproduce

RivenSkaye:
I can reproduce this - on latest stable even. Though slightly different from the original issue because _it's only the first tab_. If I hit the `+` I _do_ get an msys profile terminal.
![image](https://github.com/lapce/lapce/assets/14993038/b721007a-5f05-4e26-8140-b77be671c0f8)
![image](https://github.com/lapce/lapce/assets/14993038/3090ec27-086f-4415-af0b-b21bac45b8d2)
![image](https://github.com/lapce/lapce/assets/14993038/9f6dc650-37ae-424a-bfa3-82efa5ced6b7)

term config:
```toml
[terminal]
font-family = "'Hasklug NF', 'Hasklug Nerd Font'"
font-size   = 14

[terminal.default-profile]
windows = "msys"

[terminal.profiles.msys]
args    = ["--login", "-i"]
command = "C:/msys64/usr/bin/bash.exe"
env     = {MSYSTEM = "MINGW64", CHERE_INVOKING = "1", MSYS2_PATH_TYPE = "inherit", MSYS2_NOSTART = "yes", MSYSCON = "defterm"}
```

@maddes8cht you might wanna test for this as well. Though granted, it kinda sucks to have to open two shells to get the one you want.
[...]

PR #3337, opened 2024-06-21
Proposal for WindowTabData tests.

Wcubed:
This is an attempt to start with unit testing the `WindowTabData` struct and it's commands.
With the idea that when a bug is fixed, such as https://github.com/lapce/lapce/issues/3275, we can add a test for it so it doesn't come back again.

This is a draft, because there are still some problems that I am not sure what the best approach will be for, or whether this is even the correct testing direction in the first place. For example, the WindowTabData requires the `LapceDb` to exist. But that reads from real configuration files, which makes the test badly reproducable. Same for `LapceConfig` which is loaded in `WindowTabData::new`.

### DIFF
--- a/lapce-app/src/window_tab.rs
+++ b/lapce-app/src/window_tab.rs
@@ -2573,3 +2573,65 @@ fn open_uri(path: &Path) {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use floem::reactive::provide_context;
+
+    use super::*;
+
+    #[test]
+    fn default_terminal_follows_terminal_profile() {
+        let cx = Scope::new();
+
+        let (release_read_signal, _) = cx.create_signal(Arc::new(None));
+        let window_common_data = WindowCommonData {
+            window_command: Listener::new_empty(cx),
+            window_scale: cx.create_rw_signal(1.0),
+            size: cx.create_rw_signal(Size::ZERO),
+            num_window_tabs: cx.create_memo(|_| 1),
+            window_maximized: cx.create_rw_signal(false),
+            window_tab_header_height: cx.create_rw_signal(0.0),
+            latest_release: release_read_signal,
+            ime_allowed: cx.create_rw_signal(false),
+            cursor_blink_timer: cx.create_rw_signal(TimerToken::INVALID),
+            hide_cursor: cx.create_rw_signal(false),
+            app_view_id: cx.create_rw_signal(ViewId::new()),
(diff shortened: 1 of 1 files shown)


panekj:
> so it doesn't come back again.

The issue never came back, it wasn't implemented in the first place.

lapce-app/src/window_tab.rs
```diff
@@ -2573,3 +2573,65 @@ fn open_uri(path: &Path) {
...
+
+        let terminal_config = tab_data
+            .common
+            .config
+            .get_untracked()
+            .terminal
+            .get_default_profile()
+            .expect("Config should be available");
+
+        assert_eq!(terminal.title.get_untracked(), terminal_config.name);
```
panekj, line 2634: This only tests that the terminal tab title is equal to the terminal profile name, which is incorrect assumption as it cannot be upheld.

panekj:
> whether this is even the correct testing direction in the first place.

I'd much prefer people focused on https://github.com/lapce/floem/issues/8 rather than unit testing Lapce, as that issue would realistically allow for automation testing in CI.

lapce-app/src/window_tab.rs
Wcubed: Good point. I thought that would work to at least check if the profile got used.
I'd like to check if it attempted to open the corect terminal.
Now that I say that though, that is dependant on the os it is running on, which would make the tests not reproducable.
Would it be useful for me to look into how to detach the application's "innards" more from the file system / system specific things? I think it is valuable for the application core to be testable, but at the moment I get the feeling things are coupled too closely for that.

panekj: > I think it is valuable for the application core to be testable

I would say it is not...

> but at the moment I get the feeling things are coupled too closely for that.

...and You are correct, we reference things all over the codebase so testing "core" is not testing anything really.

Wcubed:
> > whether this is even the correct testing direction in the first place.
> 
> I'd much prefer people focused on [lapce/floem#8](https://github.com/lapce/floem/issues/8) rather than unit testing Lapce, as that issue would realistically allow for automation testing in CI.

That sounds like a better idea indeed 😄 . I'll close this issue.

panekj:
I want to see some testing like that in Lapce in future, but from the current code layout it's terrible to implement which is why I'm hoping for accessibility/automation stuff in Floem.

panekj:
keeping it open for future

PR #3342, opened 2024-06-27
Terminal file jump: change the judgment point to Ctrl + left mouse click.

jm-observer:
1. Terminal file jump: change the judgment point to Ctrl + left mouse click.
2. Optimize some code according to clippy.

### DIFF
--- a/lapce-app/src/terminal/raw.rs
+++ b/lapce-app/src/terminal/raw.rs
@@ -54,7 +54,10 @@ impl RawTerminal {
         proxy: ProxyRpcHandler,
         term_notification_tx: Sender<TermNotification>,
     ) -> Self {
-        let config = alacritty_terminal::term::Config::default();
+        let config = alacritty_terminal::term::Config {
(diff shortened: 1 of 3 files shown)


lapce-app/src/terminal/raw.rs
```diff
@@ -54,7 +54,10 @@ impl RawTerminal {
         proxy: ProxyRpcHandler,
         term_notification_tx: Sender<TermNotification>,
     ) -> Self {
-        let config = alacritty_terminal::term::Config::default();
+        let config = alacritty_terminal::term::Config {
+            semantic_escape_chars: ",│`|\"' ()[]{}<>\t".to_string(),
```
dzhou121, line 58: why not leave it to default?

jm-observer: The default configuration includes the character ":", so when clicking, the line number and character number after ":" cannot be obtained. For example, with "src/main.rs:11:9", the terminal can only get "src/main.rs" and cannot get the subsequent "11:9".

PR #3354, opened 2024-07-02
1. terminal: added support for Alt+ character（some shortcut keys） input

jm-observer:


panekj (changes requested):
Why does it need to allocate?

jm-observer:
I have a shortcut key alt+/ in the terminal for command completion. However, lapce's terminal doesn't handle alt+/. I believe some people have similar shortcuts, so I added this feature.

panekj:
Ok but it doesn't need to allocate a String for that

jm-observer:
Here, only String can be used.

https://github.com/jm-observer/lapce/commit/86baa4104aae23090a793772c9917e7ee7076f00#:~:text=let%20mut%20rs,(rs)

jm-observer:
Here, only String can be used.
https://github.com/jm-observer/lapce/commit/86baa4104aae23090a793772c9917e7ee7076f00#:~:text=else%20if%20key,(rs)

panekj:
Why close it?

jm-observer:
Maybe there's a better way to do it

PR #3355, opened 2024-07-03
Avoid receiving the terminal disconnection event before the terminal …

jm-observer:
Avoid receiving the terminal disconnection event before the terminal has finished reading, which leads to incorrect display.

### DIFF
--- a/lapce-proxy/src/terminal.rs
+++ b/lapce-proxy/src/terminal.rs
@@ -1,3 +1,4 @@
+use std::time::Duration;
 use std::{
     borrow::Cow,
     collections::VecDeque,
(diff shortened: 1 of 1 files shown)


dzhou121:
maybe just drain the pty like this https://github.com/alacritty/alacritty/blob/master/alacritty_terminal/src/event_loop.rs#L268

jm-observer:
I've tried, but I still can't read all the data in this loop

dzhou121:
How can I replicate this issue?

jm-observer:
I'm doing a cargo check on a small project, and it either stops, doesn't show all, or there's no problem

dzhou121:
> I'm doing a cargo check on a small project, and it either stops, doesn't show all, or there's no problem

Is this "Run and Debug"?

jm-observer:
here：image

jm-observer:
?

jm-observer:
I encountered it in a windows environment

PR #3358, opened 2024-07-08
1. terminal: The default number of jump columns is 0

jm-observer:
No number of columns can also match, then jump to

### DIFF
--- a/lapce-app/src/terminal/view.rs
+++ b/lapce-app/src/terminal/view.rs
@@ -124,7 +124,7 @@ pub fn terminal_view(
     });
 
     // for rust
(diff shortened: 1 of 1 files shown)


HaoliangXu:
Can you explain what Jump Column is? What should the default number be? And what's wrong with the picture?
Sorry, as noob, I don't have a clue about it.
TIA

jm-observer:
This function gets the information of the file, the number of rows, and the number of columns from the terminal, and then jumps to the specific location. For things like cargo clippy, it gives a specific file, number of rows, and number of columns, so you can jump around normally. But in the case of log (which is the image above), it only gives the file and the number of lines, so it can't jump. In this pull request, the default number of columns is 0, so log can be jumped

PR #3367, opened 2024-07-12
Terminal ctrl + click to jump: Support path contains '-' and toml file

jm-observer:
- [ ] ctrl + click a similar text (_Cargo.toml:1:1_, examples/one-two.rs:2:5) in the terminal to jump to the corresponding file

The effect is as follows:

https://github.com/user-attachments/assets/285aeb43-5462-43d0-957f-7812ccccb740


jm-observer:
Please let me know if it doesn't fit to merge

PR #3387, opened 2024-07-26
show code lens

jm-observer:
Show code lens, but the click event has not been implemented

the effect as follow:

https://github.com/user-attachments/assets/9c6f7d98-1861-4715-b2bb-528df42392fe

### DIFF
--- a/lapce-app/src/window_tab.rs
+++ b/lapce-app/src/window_tab.rs
@@ -1926,7 +1926,8 @@ impl WindowTabData {
                     // todo filter by language
                     self.main_split.docs.with_untracked(|x| {
                         for doc in x.values() {
-                            doc.on_update(None);
+                            doc.get_code_lens();
+                            doc.get_semantic_styles();
                         }
                     });
                 }
--- a/lapce-app/src/window_tab.rs
+++ b/lapce-app/src/window_tab.rs
@@ -10,7 +10,7 @@ use std::{
 
 use crossbeam_channel::Sender;
 use floem::{
-    action::{open_file, TimerToken},
+    action::{open_file, remove_overlay, TimerToken},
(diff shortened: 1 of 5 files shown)


dzhou121:
I'm wondering if it makes sense to reuse the "code action" popup for the codelens popup?

jm-observer:
I refer to the implementation of the code action. Compared with code lens, it does not use overlay, has the function of turning pages, and seems to have the function of shortcut keys, so the implementation is more complicated for me.

jm-observer:
Now there's a problem with code lens. In addition to clicking to cancel the display, what scenarios should I cancel the display? How to achieve more appropriate?

jm-observer:
In addition, the code action can only display one icon at a time, and the icon does not always exist. but code lens, on the other hand, shows more than one at a time and is always there. They are not similar

dzhou121:
I think when you click another codelens icon, it should cancel the previous one if there was one. That's also why I suggested to reuse the code actions popup, so that it's guaranteed to always have one popup display at all time and the style is consistent as well.

The codelens icon is shown all the time but the popup for it doesn't.

jm-observer:
Wait until i am more familiar with the code action

jm-observer:
the effect as follow:
https://github.com/user-attachments/assets/6cddaca9-75d3-4781-bd92-69e535b5e77b

lapce-app/src/window_tab.rs
```diff
@@ -1924,7 +1926,7 @@ impl WindowTabData {
                     // todo filter by language
                     self.main_split.docs.with_untracked(|x| {
                         for doc in x.values() {
-                            doc.get_semantic_styles();
+                            doc.on_update(None);
```
dzhou121, line 1929: not a good idea to call `on_update` here, and I would keep `on_update` private.

PR #3389, opened 2024-07-27
1. editor: Right-click add command "Show Call Hierarchy"

jm-observer:
1. editor: Right-click add command "Show Call Hierarchy".Currently, only the interaction with the lsp is complete, and the result is not rendered

2. fixed warnings in clippy report

### DIFF
--- a/lapce-app/src/command.rs
+++ b/lapce-app/src/command.rs
@@ -183,6 +183,10 @@ pub enum LapceWorkbenchCommand {
     #[strum(message = "Open File")]
     OpenFile,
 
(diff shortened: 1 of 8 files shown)


dzhou121:
Maybe we don't put it in the context menu for now?

jm-observer:
why not?

jm-observer:
No, it's getting the name of the mouse position function

dzhou121:
I mean from a user's perspective, it doesn't "work" when you click "Show Call Hierarchy" in the context menu. So let's hide it for the moment and expose it when it's fully implemented.

PR #3390, opened 2024-07-28
Gracefully handle locale_config panic on macOS(#2936)

### LINKED ISSUE #2936: Lapce not working at all on macOS
## Lapce Version

v0.3.1

## System information

macOS Sonoma 14.3 (23D56)

## Describe the bug

Opening Lapce brings up its window. After this point regardless of the user action, the application does not perform any responses. Terminal does not load (cursor stuck forever). Cannot open a file. Cannot change UI preferences. Log file is empty. Checking for an update does not give any response.

It feels like its just stuck at some point.

## Additional information

Re-installing did not help.

asemarafa:
### Summary
Fixes #2936

The function std::panic::catch_unwind is used to catch any panic that might occur when calling locale_config::Locale::global_default().

This ensures that if the locale_config crate encounters a null value and panics, it will not crash Lapce.

### DIFF
--- a/lapce-app/src/command.rs
+++ b/lapce-app/src/command.rs
@@ -183,6 +183,10 @@ pub enum LapceWorkbenchCommand {
     #[strum(message = "Open File")]
     OpenFile,
 
+    #[strum(serialize = "show_call_hierarchy")]
+    #[strum(message = "Show Call Hierarchy")]
+    ShowCallHierarchy,
+
     #[strum(serialize = "reveal_in_file_tree")]
     #[strum(message = "Reveal in File Tree")]
     RevealInFileTree,
--- a/lapce-app/src/editor.rs
+++ b/lapce-app/src/editor.rs
@@ -1267,6 +1267,49 @@ impl EditorData {
         );
     }
 
+    pub fn show_call_hierarchy(&self) {
+        let doc = self.doc();
+        let path = match if doc.loaded() {
+            doc.content.with_untracked(|c| c.path().cloned())
+        } else {
+            None
+        } {
+            Some(path) => path,
+            None => return,
+        };
+
+        let offset = self.cursor().with_untracked(|c| c.offset());
+        let (_start_position, position) = doc.buffer.with_untracked(|buffer| {
+            let start_offset = buffer.prev_code_boundary(offset);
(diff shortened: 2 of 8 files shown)


panekj:
There is no panic in the code you are changing

asemarafa:
> There is no panic in the code you are changing

Can you please check https://github.com/lapce/lapce/issues/2936#issuecomment-2053553781

panekj:
> > There is no panic in the code you are changing
> 
> Can you please check [#2936 (comment)](https://github.com/lapce/lapce/issues/2936#issuecomment-2053553781)

I checked but the proposed fix is going to override actual locale

asemarafa:
I changed my fix to handle the panic instead, i hope this is a more acceptable solution 


The original implementation did not handle the case where locale_config::Locale::global_default() might panic (e.g., due to a null value). This caused Lapce to crash on macOS when certain locale settings were used.

By using std::panic::catch_unwind, we ensure that the application does not crash and can handle the error gracefully.

asemarafa:
Someone reported the issue https://github.com/rust-locale/locale_config/issues/10 with locale_config two years ago

panekj:
Yes, I looked into `locale_config`, it will be better to fork the crate and use fixed version: https://github.com/lapce/locale_config (i will add patches from upstream later, unless someone else opens a PR earlier)

panekj:
https://github.com/lapce/lapce/pull/3395

panekj:
Could you please check if building from that works fine?

asemarafa:
Tested and no panic:

```bash
$ ./Lapce.app/Contents/MacOS/lapce -w
$ unset LANG
$ echo $LANG

$ ./Lapce.app/Contents/MacOS/lapce -w
```

But I want to note something, this fix to locale_config is not a total fix.

If you run, from the embedded terminal in Lapce, the command:

```bash
echo $LC_ALL
```

The result is:

```
.UTF-8
```

So yeah you can merge #3395.

But I was seeing how macOS Terminal and iTerm2 handle this scenario.

Both of them are able to set LANG (not LC_ALL) to en_IE.UTF-8 if:

```bash
defaults read -g AppleLocale
en_US@rg=iezzzz
```

If I have time to figure this out I will send a PR.

PR #3398, opened 2024-08-02
Change the style to avoid issues with displaying Chinese.

jm-observer:


dzhou121:
This breaks the scroll of the text area.

jm-observer:
It looks like I still need to check the source code of floem-cosmic-text.

PR #3403, opened 2024-08-05
show call hierarchy and jump to location

jm-observer:
1. show call hierarchy and jump to location
2. Some styles need further optimization.

the effect as follow:

https://github.com/user-attachments/assets/e69f2dbf-6166-40a3-8129-80652219dffe

### DIFF
--- a/lapce-app/src/command.rs
+++ b/lapce-app/src/command.rs
@@ -1,5 +1,6 @@
 use std::{path::PathBuf, rc::Rc, sync::Arc};
 
+use floem::reactive::RwSignal;
 pub use floem::views::editor::command::CommandExecuted;
 use floem::{
     keyboard::Modifiers, peniko::kurbo::Vec2, views::editor::command::Command,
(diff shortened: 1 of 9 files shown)


lapce-app/src/window_tab.rs
```diff
@@ -437,9 +439,18 @@ impl WindowTabData {
...
                     .unwrap_or_else(|_| i.panel.panels.clone());
+                if !panel_order
+                    .iter()
+                    .any(|x| x.1.iter().any(|x| *x == PanelKind::CallHierarchy))
+                {
+                    panel_order
+                        .entry(PanelPosition::BottomLeft)
+                        .or_default()
+                        .push_back(PanelKind::CallHierarchy);
+                }
```
dzhou121, line 453: `db.get_panel_orders()` should have taken care of this.

lapce-app/src/command.rs
```diff
@@ -746,6 +748,9 @@ pub enum InternalCommand {
         tab_index: usize,
         terminal_index: usize,
     },
+    CallHierarchyIncoming {
+        item: RwSignal<CallHierarchyItemData>,
```
dzhou121, line 752: It should pass raw data rather than signals in the commands. 

The reason for that is it's easy to mess up with the scope of the signal if doing this way.

lapce-app/src/window_tab.rs
jm-observer: For a newly upgraded lapce, there will be problems. At this time, the database does not have this type, which will cause the ui to fail to display.I had this problem when I was developing it.

jm-observer: Or don't even think about this scenario, you decide?

PR #3419, opened 2024-08-09
Implement feature "Run in Terminal"

jm-observer:
the effect as follow:

https://github.com/user-attachments/assets/04152756-4833-4c9c-b8aa-69969868dca1

### DIFF
--- a/lapce-app/src/window_tab.rs
+++ b/lapce-app/src/window_tab.rs
@@ -1394,6 +1394,39 @@ impl WindowTabData {
                     editor_data.call_hierarchy(self.clone());
                 }
             }
+            RunInTerminal => {
+                if let Some(editor_data) =
+                    self.main_split.active_editor.get_untracked()
+                {
+                    let name = editor_data.get_select_text().content;
+                    tracing::info!("{name}");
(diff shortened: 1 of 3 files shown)


lapce-app/src/window_tab.rs
```diff
@@ -1394,6 +1394,39 @@ impl WindowTabData {
                     editor_data.call_hierarchy(self.clone());
                 }
             }
+            RunInTerminal => {
+                if let Some(editor_data) =
+                    self.main_split.active_editor.get_untracked()
+                {
+                    let name = editor_data.get_select_text().content;
```
dzhou121, line 1401: You can use this https://github.com/lapce/lapce/blob/master/lapce-app/src/editor.rs#L2376
to get the selection text which is more reliable.

jm-observer: This method retrieves the word under the mouse even if no text is selected. But my intention is to require text selection.

dzhou121: There's no harm to retrieves the word under the mouse even if no text is selected.

PR #3424, opened 2024-08-10
implement auto-save

jm-observer:
1. I'm uncertain whether I missed a location where the auto-save function should be invoked
2. The focus functionality in the dependency seems to have an issue.


dzhou121:
auto save is already implemented https://github.com/lapce/lapce/blob/master/lapce-app/src/doc.rs#L666

jm-observer:
ok, it seems sufficient.

PR #3436, opened 2024-08-14
Optimize shortcut key settings

jm-observer:
1.filter out alt/ctrl/shift
2. filter out duplicate
3. adjust some styles.


dzhou121:
What does this PR try to achieve?

jm-observer:
> 3. adjust some styles.

adjust the height


> 1.filter out alt/ctrl/shift
2. filter out duplicate

Avoid these scenarios:

PR #3440, opened 2024-08-15
implement "Open in GitHub" command

jm-observer:
the effect as follow:

https://github.com/user-attachments/assets/602405e9-bd63-4282-bd89-128a96a760b1

### DIFF
--- a/lapce-app/src/command.rs
+++ b/lapce-app/src/command.rs
@@ -191,9 +191,9 @@ pub enum LapceWorkbenchCommand {
     #[strum(message = "Reveal in Panel")]
     RevealInPanel,
 
-    #[strum(serialize = "open_in_git_bub")]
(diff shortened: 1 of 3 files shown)


lapce-app/src/command.rs
```diff
@@ -191,6 +191,10 @@ pub enum LapceWorkbenchCommand {
     #[strum(message = "Reveal in Panel")]
     RevealInPanel,
 
+    #[strum(serialize = "open_in_git_bub")]
+    #[strum(message = "Open In GitHub")]
+    OpenInGitHub,
```
dzhou121, line 196: Can you please rename it to  `SourceControlOpenActiveFileRemoteUrl`

similar to this:
https://github.com/lapce/lapce/blob/master/lapce-app/src/command.rs#L490

PR #3463, opened 2024-09-01
Set the Palette to default to the Help Palette.

jm-observer:
This makes it more convenient for the user.

help

### DIFF
--- a/lapce-app/src/palette.rs
+++ b/lapce-app/src/palette.rs
@@ -20,6 +20,7 @@ use floem::{
         SignalWith,
     },
 };
+use im::Vector;
 use itertools::Itertools;
 use lapce_core::{
     buffer::rope_text::RopeText, command::FocusCommand, language::LapceLanguage,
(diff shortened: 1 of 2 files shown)


dzhou121:
I probably would keep it as is because:

1) There's a search icon on the palette input bar before it was clicked so it kind of suggests it should be a file search

2) Maybe it would be better if it's a setting?

jm-observer:
reasons i commit this pr:
1. I often forget the symbols for various functions.I think many people have the same concern.
2. I noticed that VSCode now also sets the Help Palette as the default, which wasn't the case before.

https://github.com/user-attachments/assets/a6f8898e-9b54-4239-aaba-f7c3300f96ea

dzhou121:
Got you. 

It still have one issue, that when you do ctrl+p, it still shows the help.

jm-observer:
If the default is the Help item, we can refer to VSCode: when accessed via a shortcut key, it should show file search instead of the Help item.

jm-observer:
the effect as follow:

PR #3464, opened 2024-09-01
unwrap some results that significantly impact the program.

jm-observer:
Additionally, there are nearly 200 similar results that will be handled.


dzhou121:
I won't use unwrap for those because they were not supposed to crash Lapce.

jm-observer:
Alright, then I'll change them all to be logged into the logs.

PR #3485, opened 2024-09-10
1. Ensure that the panel can be displayed.

jm-observer:
2. Adjust some logs.


jm-observer:
Modifying just this part doesn't necessarily ensure that it will be displayed, for example, when the user is upgrading the version.

PR #3488, opened 2024-09-10
1. add unsafe

jm-observer:


jm-observer:
67a9806df4061ce0f51aba08aa14c928

dzhou121:
It should still be "safe" for 2021 edition I think.

jm-observer:
I regret it now; I think I should have insisted, because this is indeed an error message.

PR #3494, opened 2024-09-12
1. Add the editor setting disable_single_click: Disable single-click …

jm-observer:
Add the editor setting disable_single_click: Disable single-click to open files.


defaults/settings.toml
```diff
@@ -5,6 +5,7 @@ modal = false
 color-theme = "Lapce Dark"
 icon-theme = "Lapce Codicons"
 custom-titlebar = true
+disable-single-click = false
```
dzhou121, line 8: we probably want to rename this to something like `file-exploerer-double-click`

PR #3496, opened 2024-09-12
Right click

jm-observer:
Add a right-click command "Add Run Debug Config" to quickly add a configuration to the run.toml file

the effect is as shown in following video:

https://github.com/user-attachments/assets/043a28c6-cf18-4382-957d-34b8b16d927a

### DIFF
--- a/lapce-app/src/editor.rs
+++ b/lapce-app/src/editor.rs
@@ -2841,9 +2841,19 @@ impl EditorData {
                     Some(CommandKind::Workbench(
                         LapceWorkbenchCommand::RevealInFileExplorer,
                     )),
+                    Some(CommandKind::Workbench(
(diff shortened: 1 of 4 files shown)


lapce-app/src/editor.rs
```diff
@@ -2818,40 +2818,71 @@ impl EditorData {
...
+                    )),
+                    None,
+                    Some(CommandKind::Edit(EditCommand::ClipboardCut)),
+                    Some(CommandKind::Edit(EditCommand::ClipboardCopy)),
+                    Some(CommandKind::Edit(EditCommand::ClipboardPaste)),
+                    None,
+                    Some(CommandKind::Workbench(
+                        LapceWorkbenchCommand::PaletteCommand,
+                    )),
+                ]
```
dzhou121, line 2884: We probably want these for run.toml as well

PR #3520, opened 2024-09-19
Move the ID to the front to facilitate initiating cancel requests later.

jm-observer:
1. Move the ID to the front to facilitate initiating cancel requests later.
2. maybe conflict with previous prs

### DIFF
--- a/lapce-app/src/app.rs
+++ b/lapce-app/src/app.rs
@@ -3817,6 +3817,7 @@ pub fn launch() {
         let notification = create_signal_from_channel(rx);
         create_effect(move |_| {
             if notification.get().is_some() {
+                tracing::debug!("notification reload_config");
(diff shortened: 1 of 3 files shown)


dzhou121:
I guess this is for send cancel to lsp server? 

I think it should use the plugin id to send the notification to, and binding a LSP id a a window doesn't make sense to me.

jm-observer:
> I guess this is for send cancel to lsp server?
yes
> I think it should use the plugin id to send the notification to
i don't understand. lapce send request to all plugin, how use plugin id? Perhaps maintain a global ID, but still need to pass the ID to the plugin from the main thread?

dzhou121:
You're right. Then probably we should use the request id, and then the plugins route them based on the request id.

PR #3527, opened 2024-09-20
Format the document before auto-saving.

jm-observer:

### DIFF
--- a/lapce-app/src/doc.rs
+++ b/lapce-app/src/doc.rs
@@ -703,6 +703,7 @@ impl Doc {
             let doc = self.clone();
             let scope = self.scope;
             let proxy = self.common.proxy.clone();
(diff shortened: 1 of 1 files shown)


lapce-app/src/doc.rs
```diff
@@ -696,8 +718,25 @@ impl Doc {
...
+                            if let Ok(ProxyResponse::GetDocumentFormatting {
+                                edits,
+                            }) = result
+                            {
+                                let current_rev = doc.rev();
+                                if current_rev == rev {
+                                    doc.do_text_edit(&edits);
+                                }
+                            }
+                            doc.save(|| {});
```
dzhou121, line 732: You need to check doc rev and return early if not matching. You can't run save if the doc has changed.

PR #3531, opened 2024-09-20
1. enhance the panel view

jm-observer:
2. display the folder name of workspace


dzhou121:
I think it's better to display it as the first line in file explorer

jm-observer:
like this?

dzhou121:
> like this? 

Yes!

PR #3533, opened 2024-09-21
1. Add the teminal setting default-program: Set the default program w…

jm-observer:
Add the teminal setting default-program: Set the default program when creating a new terminal.


dzhou121:
This is already implemented I think https://github.com/lapce/lapce/blob/e69729efa1a56df20a27bd6f27af66d277e06276/lapce-app/src/config/terminal.rs#L24

jm-observer:
Yes, this feature is indeed quite comprehensive, although it's not very obvious or clear.

PR #3539, opened 2024-09-24
change GoToLocation to JumpToLocation

jm-observer:


dzhou121:
ignore_unconfirmed should be kept as false I think

PR #3546, opened 2024-09-28
`enable_error_lens` is configured to display diagnostic messages but …

jm-observer:
`enable_error_lens` is configured to display diagnostic messages but does not include the styling for diagnostics.


dzhou121:
We probably need a different config for this.

PR #3573, opened 2024-10-14
read terminal even if it already exit status

dzhou121:


dzhou121:
@jm-observer Can you please test if this PR fixes the windows terminal not getting all output for you?

This is another approach for https://github.com/lapce/lapce/pull/3477

jm-observer:
I'll merge it into my branch for observation, but since the issue occurs probabilistically, I won't be able to confirm whether it meets the requirements in the short term.

PR #3590, opened 2024-11-14
Fix git & rust warnings on linux

gtsiam:
- No idea how you managed to get CRLF line endings on `.cargo/config.toml` without git throwing a tantrum. I'm definitely curious. But let's not keep that around.
- Fixed an unused const warning when compiling on linux.
- Removed nonsense toml git diff driver from `.gitattributes`

Also, minor pet peave of mine: Kindly don't squash my commits.
Or do, if you want worse commit history. I don't care that much.


gtsiam:
The project seems abandoned, so...

PR #3610, opened 2025-01-02
Improve support for `musl` targets

ljgermain:
It is necessary to dynamically link `libc` on `musl` targets in order for a window to spawn when launching `lapce` since `dlopen` won't work without it, which is required by `winit` to load X11 and Wayland libraries. This can be achieved by appending `-C target-feature=-crt-static` to the rust flags.

IIRC, Alpine Linux packages a Rust toolchain that does this by default, but Alpine isn't the only `musl` distro, so it's worth specifying this explicitly.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,11 @@
 # Link runtime statically so Visual C++ Redist is not required
 [target.'cfg(all(windows, target_env = "msvc"))']
 rustflags = ["-C", "target-feature=+crt-static"]
+
+# `musl` must be linked dynamically instead of statically in order for `dlopen` to work,
+# which is required for `winit` to open X11 or Wayland libraries and spawn a window.
+#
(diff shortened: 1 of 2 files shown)


panekj:
> IIRC, Alpine Linux packages a Rust toolchain that does this by default, but Alpine isn't the only `musl` distro, so it's worth specifying this explicitly.

Not really, every `musl` based distro knows of Rust build targets shenanigans and has specifically patched `rust` to use dynamic linking by default.
I don't see a point of forcing such config when it's already enforced in system toolchains.

ljgermain:
I prefer to use `rustup` instead of a system toolchain (as do a lot of other Rust developers), and if someone were on a `musl` system and wanted to contribute to `lapce` while using `rustup` then the lack of this could prevent them from testing their changes. So the purpose is really compatibility with `rustup` on `musl`, rather than system toolchains.

I suppose this isn't important enough to warrant a change, though.

PR #3615, opened 2025-01-17
make lapce use proxy installed in PATH

panekj:


dzhou121:
I feel like this complicates things and I can't see the benefits?

panekj:
> I feel like this complicates things and I can't see the benefits?

This should allow running proxy on hosts that don't have it built by us and could also run proxy installed via system package manager

panekj:
The target audience for that change are *BSD and Linux distros

lapce-app/src/proxy/remote.rs
```diff
@@ -116,20 +116,37 @@ pub fn start_remote(
         _ => format!("{remote_proxy_path}/lapce"),
     };
 
-    if !remote
+    let mut remote_version = remote
         .command_builder()
-        .args([&remote_proxy_file, "--version"])
+        .args(["lapce-proxy", "--version"])
```
dzhou121, line 121: This will cause the Lapce managed proxies fail to break out quickly, which is the majority.

panekj: I don't understand what you mean

dzhou121: So for remotes that Lapce self installs, it won't be in PATH, which means this will always fail for them.

panekj: Yes, that is expected

dzhou121: I don't think that's a good idea, since most will be Lapce installed ones.

panekj: I don't think it's that big of a deal since it will always fallback to script

dzhou121: So I connect to my remote server every day, and I don't want it to fallback to script which is much faster.

PR #3622, opened 2025-01-24
Upstream winit

dzhou121:


Cargo.lock
```diff
@@ -975,14 +1055,23 @@ dependencies = [
...
+]
+
 [[package]]
 name = "cosmic-text"
 version = "0.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "59fd57d82eb4bfe7ffa9b1cec0c05e2fd378155b47f255a67983cb4afe0e80c2"
 dependencies = [
  "bitflags 2.6.0",
- "fontdb",
```
panekj, line 985: would be nice to bring fontdb in cosmic-text up2date as well

Cargo.lock
```diff
@@ -2315,6 +2569,58 @@ version = "0.13.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d196ffc1627db18a531359249b2bf8416178d84b729f3cebeb278f285fb9b58c"
 
+[[package]]
+name = "gtk"
+version = "0.18.2"
```
panekj, line 2574: can we not pull in gtk if we don't use it? It seems that would require forking muda and hiding linux backend behind feature

PR #3633, opened 2025-02-09
Improve dependency installation by creating setup files for major distributions

guymarshall:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Features/Changes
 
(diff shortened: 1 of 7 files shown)


panekj:
It would be better if they were added to Makefile than separate scripts

guymarshall:
Ok I have no experience with Makefiles. Could it be done with a single setup.sh file that checks for the existence of apt, dnf, pacman in if statements and uses the correct one, else gives an error saying something like "Your distribution is not currently supported. Please create a new issue"?

PR #3637, opened 2025-02-21
Fix compilation with Rust 2024 Edition

hbina:

### DIFF
--- a/lapce-app/src/app.rs
+++ b/lapce-app/src/app.rs
@@ -7,58 +7,59 @@ use std::{
     process::Stdio,
     rc::Rc,
     sync::{
-        atomic::AtomicU64,
-        mpsc::{channel, sync_channel, SyncSender},
         Arc,
+        atomic::AtomicU64,
+        mpsc::{SyncSender, channel, sync_channel},
     },
 };
 
-use anyhow::{anyhow, Result};
+use anyhow::{Result, anyhow};
 use clap::Parser;
 use floem::{
+    IntoView, View,
     action::show_context_menu,
     event::{Event, EventListener, EventPropagation},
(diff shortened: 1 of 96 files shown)


LunaStev:
I'm curious if there's a particular reason to switch to the Rust 2024 Edition at this point.
Unless there's a strong motivation, sticking with the 2021 Edition might be better for project stability and consistency.

panekj:
> Unless there's a strong motivation, sticking with the 2021 Edition might be better for project stability and consistency.

Lapce historically supports only latest Rust version, so it's not like stability is a concern. MSRV has been used to prevent older compilers from trying to compile Lapce and for clippy to figure out supported lints.

LunaStev:
> > Unless there's a strong motivation, sticking with the 2021 Edition might be better for project stability and consistency.
> 
> Lapce historically supports only latest Rust version, so it's not like stability is a concern. MSRV has been used to prevent older compilers from trying to compile Lapce and for clippy to figure out supported lints.

Got it, that makes sense. Then I guess there's nothing to be concerned about.

lapce-app/src/app.rs
```diff
@@ -2219,7 +2220,7 @@ fn palette_item(
     palette_item_height: f64,
     config: ReadSignal<Arc<LapceConfig>>,
     keymap: Option<&KeyMap>,
-) -> impl View {
+) -> impl View + use<> {
```
hbina, line 2223: This use of `use<>` is the only syntax/semantic change of this PR. The rest of it is just cargo fmt reordering imports using case-sensitive order.

hbina:
I forgot I still have this up. I am bumped it up to 1.87.0.

PR #3643, opened 2025-02-27
Focus window when open file in a existing window

yan4rust:

### DIFF
--- a/lapce-app/src/app.rs
+++ b/lapce-app/src/app.rs
@@ -3932,7 +3932,6 @@ pub fn launch() {
                 if let Some(window_tab) = app_data.active_window_tab() {
                     window_tab.open_paths(&paths);
                     // focus window after open doc
-                    floem::action::set_window_maximized(true);
                     floem::action::focus_window();
                 }
             }
--- a/lapce-app/src/app.rs
+++ b/lapce-app/src/app.rs
@@ -3931,6 +3931,9 @@ pub fn launch() {
             if let Some(CoreNotification::OpenPaths { paths }) = notification.get() {
                 if let Some(window_tab) = app_data.active_window_tab() {
                     window_tab.open_paths(&paths);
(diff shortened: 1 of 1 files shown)


lapce-app/src/app.rs
```diff
@@ -3931,6 +3931,9 @@ pub fn launch() {
             if let Some(CoreNotification::OpenPaths { paths }) = notification.get() {
                 if let Some(window_tab) = app_data.active_window_tab() {
                     window_tab.open_paths(&paths);
+                    // focus window after open doc
+                    floem::action::set_window_maximized(true);
```
dzhou121, line 3935: What's the reason to set window maximized?

yan4rust: If not set maximized first ,the window will not show as expected.  [focus_window doc](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.focus_window)

panekj: > Has no effect if the window is already in focus, minimized, or not visible.  
  
This is correct behaviour, and there is no need for window to be maximized.

yan4rust: When the VS Code window is minimized, right-clicking 'Open with Code' will maximize and bring the VS Code window to the front.  I believe Lapce should exhibit the same behavior.

dzhou121: I see. The correct one to use is `set_minimized(false)` but it's not exposed by Floem yet.

dzhou121: @yan4rust shall we remove `floem::action::set_window_maximized(true);` for now and we can merge?

yan4rust: Okay, I have deleted it.  @dzhou121

dzhou121: Thanks!

Issue #3646, opened 2025-03-18
PATH in terminal mutated inconveniently


cag:
## Lapce Version

Version: 0.4.2

## System information

Arch Linux

## Describe the bug

The path to the executable is prepended onto the PATH in the lapce terminal. This is inconvenient for me, as I have a `rustc` in my `/usr/bin` now taking precedence over the one in my .cargo directory. I want both so my package manager won't yell at me while I'm having a standard rust dev installation.

## Additional information

https://github.com/lapce/lapce/blob/c6959764c6c8da3156c2990b71bfc8135b825b57/lapce-proxy/src/lib.rs#L140

I think this is a side effect of the referenced line, as alacritty tries to preserve the spawning process PATH. Right now I'm working around it by manually sourcing a thing that dedups the first /usr/bin out of the path in every terminal session:

```
export PATH=`python -c 'import os;print(":".join(reversed(dict.fromkeys(reversed(os.environ["PATH"].split(":"))))))'`
```

What if the registration function doesn't just auto-prepend /usr/bin or whatever the install directory of lapce is to PATH, but glances over all the paths and compares each path to the lapce exe path, and only mutates PATH if the exe path is not one of the paths. Or maybe mutating PATH can be avoided altogether?

![Difficulty finding dev tools using system `rustc`](https://github.com/user-attachments/assets/09136ba2-e4d4-4045-be68-8f3b0c67bea9)

I think this also causes this to happen...

EDIT: just compiled from source, have more evidence that this is happening, as my .cargo/bin is now in the terminal PATH twice and the Rust for Lapce plugin doesn't complain anymore

PR #3648, opened 2025-03-20
only register lapce dir in PATH if not already present

cag:
Opinionated way of resolving #3646 by only mutating PATH if exe dir not already present.

### DIFF
--- a/lapce-proxy/src/lib.rs
+++ b/lapce-proxy/src/lib.rs
@@ -9,7 +9,6 @@ pub mod watcher;
 
 use std::{
     io::{stdin, stdout, BufReader},
-    path::PathBuf,
     process::exit,
     sync::Arc,
     thread,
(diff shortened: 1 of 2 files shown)


lapce-proxy/src/lib.rs
```diff
@@ -130,27 +129,31 @@ pub fn mainloop() {
...
-    if let Some(path) = path.parent() {
-        if let Some(path) = path.to_str() {
-            if let Ok(current_path) = std::env::var("PATH") {
-                let mut paths = vec![PathBuf::from(path)];
-                paths.append(
-                    &mut std::env::split_paths(&current_path).collect::<Vec<_>>(),
-                );
-                std::env::set_var("PATH", std::env::join_paths(paths)?);
-            }
+pub fn require_lapce_path() -> Result<()> {
```
panekj, line 139:
```suggestion
pub fn register_lapce_path() -> Result<()> {
```

panekj (changes requested):

lapce-proxy/src/lib.rs
```diff
@@ -138,19 +137,23 @@ pub fn mainloop() {
...
+        .canonicalize()?;
+
+    let current_path = std::env::var("PATH")?;
+    let paths = std::env::split_paths(&current_path);
+    for path in paths {
+        if exedir == path.canonicalize()? {
+            return Ok(());
         }
     }
+    let paths = std::env::split_paths(&current_path);
```
dzhou121, line 152: duplicate with line 146

cag: It intentionally shadows the previous declaration with a regeneration of the iterator

PR #3770, opened 2025-06-06
Use Diff syntax for files with .diff/.patch extensions by default.

sT331h0rs3:
- [X] Added an entry to `CHANGELOG.md` if this change could be valuable to users


panekj:
Thank you

REDMEE1 (commented):
- [ ] [_@I am not _]()

PR #3774, opened 2025-06-20
feat: update file dialog titles for better clarity

Integral-Tech:

Update titles of file dialogs for better clarity.

Before: 
Screenshot_20250620_233204

After: 
Screenshot_20250620_233121


dzhou121:
Thanks!

PR #3820, opened 2025-10-20
Improve title bar usability outside of macOS

ArthurCose:
- Increased the size of primary title bar buttons to take the full height of the bar, just like the "Connect to Remote" button.
- The right side "Settings" button is removed outside of macOS, with options moved to the "Menu" button
- Added "New Window" button above "Close Window"

The update indicator is visible just for this clip:

https://github.com/user-attachments/assets/ab10435f-4e35-4f9a-a536-6e7bebe91f78

### DIFF
--- a/lapce-app/src/title.rs
+++ b/lapce-app/src/title.rs
@@ -324,10 +324,8 @@ fn right(
         drag_window_area(empty())
             .style(|s| s.height_pct(100.0).flex_basis(0.0).flex_grow(1.0)),
         container(stack((
-            not_clickable_icon(
-                || LapceIcons::SETTINGS,
-                || false,
-                || false,
+            title_button(
+                lapce_icon(|| LapceIcons::SETTINGS, config),
                 || "Settings",
                 config,
             )
@@ -374,8 +372,7 @@ fn right(
                     }))
             }),
             update_indicator(config, latest_release, false),
-        )))
-        .style(move |s| s.margin_horiz(6.0).justify_center().items_center()),
+        ))),
         window_controls_view(
             window_command,
             true,


ryanabx:
> Increased the size of primary title bar buttons to take the full height of the bar, just like the "Connect to Remote" button.

I feel like all the buttons could use a face-lift in this regard:


https://github.com/user-attachments/assets/481bdf4a-eb6c-4d35-b949-5199d03901b7

ArthurCose:
Yeah these are difficult for me to click as well, but I'd like to tackle that as a separate PR and focus just on the title bar for this.

For those buttons, I'm not sure if it would be best to: change the style to flat + fill the space, keep the rounded corners + fill the space, or fully preserve the look and make the empty space surrounding the buttons selectable.

ryanabx:
> Yeah these are difficult for me to click as well, but I'd like to tackle that as a separate PR and focus just on the title bar for this.
> 
> For those buttons, I'm not sure if it would be best to: change the style to flat + fill the space, keep the rounded corners + fill the space, or fully preserve the look and make the empty space surrounding the buttons selectable.

No worries! I agree it's good for a separate pr :)

Without making any judgements on style, making the empty space surrounding it selectable would certainly help with usability. Haven't worked with Floem so idk if it's possible to have separate click geometry from the selection highlight.

panekj:
I can merge this if you restore Settings button

ArthurCose:
I can split that into a separate PR tomorrow, but I do find the settings button makes Lapce harder to learn + navigate by moving options to the other side of the editor in an unusual place. I'd like to find a path to remove it or justify it as the options it adds are usually found in the first menu option in other editors (See VS Code's File button)

ArthurCose:
Just brought back the settings button to non macOS platforms and reverted the changes to the popout menus to avoid duplicate options:

image
image

The settings menu doesn't use the larger button size along with the back + forth and run button, not sure if it should.

ryanabx:
> The settings menu doesn't use the larger button size along with the back + forth and run button, not sure if it should.

I personally think it should, if the other buttons are also square.

PR #3822, opened 2025-10-29
Diff view fixes

### LINKED ISSUE #3821: Diff viewer not scrollable (Windows + Linux)
## Lapce Version

0.4.5

## System information

Windows 11

## Describe the bug

The diff viewer in lapce cannot be scrolled with my laptop's touchpad. I can move the view down by using the arrow keys.

## Additional information

This commonly includes:

While a video isn't super helpful, I'm attempting to scroll here:

https://github.com/user-attachments/assets/9e87d56e-f33f-43c8-becd-bff069213b12

ArthurCose:
Fixes #3821

Not sure if it would be best to split this up. I'm looking into fixing more issues with the diff view, such as the editor skipping incorrect lines. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 ### Bug Fixes
 
 - Fix mouse wheel scrolling when viewing diff (<https://github.com/lapce/lapce/issues/3821>)
+- Fix editor tabs not selectable while appearing selectable
+- Fix flickering when reordering editor tabs
 
 ## 0.4.5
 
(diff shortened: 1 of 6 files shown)


ArthurCose:
There are more performance improvements and fixes for wrapped lines to be made, but I'm leaving that for a separate PR if I get around to it.

ryanabx (changes requested):

ryanabx:
Other than that one thing, seems good to me! Merge conflicts need to be resolved

PR #3838, opened 2025-11-19
Theme: Add fallback colors, and add `PANEL_BACKGROUND_SIDE` and `PANEL_BACKGROUND_BOTTOM`

ryanabx:
This commit series adds the ability to set fallback colors in `color.rs`, by switching each tag to an array of tags. While this looks like a big change, most of the changed lines is passing the const string arrays by slice to avoid copying anything.

The second commit adds the first multi tag colors `PANEL_BACKGROUND_SIDE` and `PANEL_BACKGROUND_BOTTOM`, which have fallbacks to `PANEL_BACKGROUND` and `EDITOR_BACKGROUND` respectively.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features/Changes
 
+- Added `PANEL_BACKGROUND_SIDE` and `PANEL_BACKGROUND_BOTTOM` theme keys
+
 ### Bug Fixes
 
 - Fix editor tabs not selectable while appearing selectable


panekj:
The bottom panel is supposed to match editor colour

ryanabx:
> The bottom panel is supposed to match editor colour

I see. Why is that? I find it looks a bit awkward, compared to the other panels.

On my custom theme, I have transparent borders, and unfortunately it makes my terminal look like this:

image

Maybe we could add a `bottompanel.background` or something like that, and if unset, the panel defaults to `editor.background`? If not, that's fine, I'll just deal with it :)

panekj:
Not my choice: https://github.com/lapce/lapce/pull/2786
I don't really have any preference what it looks like, so maybe if you refactor it to `LEFT|RIGHT|BOTTOM` + `_PANEL_BACKGROUND` and make it current behaviour as default, then you can style it however you want

panekj:
or `SIDE_PANEL_BACKGROUND` and `BOTTOM_PANEL_BACKGROUND`

ryanabx:
Sure, I can look at that tomorrow after work. Thanks!

ryanabx:
Here's my idea. To avoid problems in the future with theming changes causing backwards incompatibility (like I've seen in the past with the [Lapce VSCode theme](https://github.com/panekj/lapce-vscode-theme), we can make each theme color contain fallback keys, which can be used when the more granular theme color isn't available.

@panekj what do you think?

PR #3840, opened 2025-11-22
Update to a newer version of floem

ArthurCose:
Draft currently, as it's missing a way to get `key_without_modifiers` and has some issues with scrolling (at least for me).

Also could use more testing, at least after fixes for the above are made. I've been focused on getting it to compile.


ryanabx:
Oof, I'm ready to have so many rebasing conflicts after this one :smile: 

Looking forward to this one :)

Issue #3650, opened 2025-03-27
Open a remote workspace using build in terminal - files/folders not listed in the file explorer


branchus:
## Lapce Version
Version: 0.4.2+Nightly.115bc73

## System information
host: windows 11 pro
remote: ubuntu 24.04

## Describe the bug
after connected to the remote host, I typed in the below command in the build in terminal
$ lapce <PATH TO THE ROOT FOLDER OF WORKSPACE>
Lapce will geneally create a new window, with <FOLDER NAME> [Remote host name]
but in this windows, nothing is shown in the file explorer.

What I expect to see is the workspace folder structure shown in the file explorer

Abhijeet213:
Can i Please Have Screenshot

branchus:
> Can i Please Have Screenshot

<img width="683" alt="Image" src="https://github.com/user-attachments/assets/1cab9bdc-881a-4226-a892-ecdb137e87f5" />

I already used the lapce Notes to open a folder on remote server through ssh, but as you can see nothing showing in the file explorer.

ywllyht:
I meet the same problem.  from the document "After it's connected, it will show a green connected status on the remote development icon. Then you can start to Open folder as if you are opening a local folder."   In fact, after I connect to my Ubuntu 24.04 server succesfully, there is no response when clicking "open folder" button.   
my lapce version 0.4.5

ryanabx:
Same issue with the WSL host:

https://github.com/user-attachments/assets/e7ec0161-c941-4f04-9117-315c9c75c7cb

Issue #3472, opened 2024-09-05
WSL Can not Open Folder


mztlive:
Nothing happens when clicking Open Folder

jm-observer:
This feature needs to be revalidated and tested.

Anivie:
Nearly three months have passed, and this issue is still not fixed; connecting to WSL is completely for show?

dzhou121:
Currently opening workspaces via Open folder on WSL is not possible due to recent UI rewrite. But you can use built-in terminal to run `lapce <PATH>`.

vector-space:
I got the same error, emm...

PR #3841, opened 2025-11-22
feat: Implement folder choosing in remotes

ryanabx:
Fixes: https://github.com/lapce/lapce/issues/3650
Fixes: https://github.com/lapce/lapce/issues/3472
Fixes (part of): https://github.com/lapce/lapce/issues/2417

Editors like Visual Studio Code have a pallete-based file picker for remotes. This commit implements that, as the current code just ignores `Open Folder` and `Open File` requests.

[Screencast_20251122_131738.webm](https://github.com/user-attachments/assets/b9be6bb7-0964-41bc-a631-226c38c7d7b6)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Features/Changes
 
+- Implemented Folder/File choosing in remotes
+
 ### Bug Fixes
 
 - Fix editor tabs not selectable while appearing selectable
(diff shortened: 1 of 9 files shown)


ryanabx:
Note: In order to test this, you have to build lapce-proxy from source and manually push it to your remote, since usually the proxy script will try to download the latest nightly proxy which is incompatible with this change.

panekj:
Previously remote picker was a popup

Issue #3906, opened 2026-05-23
Runtime system theme watcher in app.rs


shivamnarkar47:
## Parent

#3903

## What to build

Wire up the runtime system theme change watcher in `lapce-app/src/app.rs` (or nearby) so that when the OS appearance changes while Lapce is running, the theme switches automatically.

Specifically:
- Initialize `system_theme::watch_system_theme()` during app startup, alongside the existing `ConfigWatcher`
- When the callback fires with a new `SystemTheme` value:
  - Check if `core.color_theme == "auto"` (to avoid unnecessary reloads when user has a fixed theme)
  - If yes, call `app_data.reload_config()` to trigger the full config reload pipeline
- The reload pipeline already propagates to all windows reactively — no additional UI changes needed

## Acceptance criteria

- [ ] System theme watcher is initialized during app startup
- [ ] OS appearance change triggers `reload_config()` when in auto mode
- [ ] OS appearance change does NOT trigger reload when a fixed theme is selected
- [ ] Watcher thread/mechanism is properly cleaned up on shutdown
- [ ] Compiles cleanly on all three platforms

## Blocked by

- Blocked by #3904, #3905

Issue #3905, opened 2026-05-23
Auto-mode theme resolution + 'auto' in theme picker


shivamnarkar47:
## Parent

#3903

## What to build

Modify `LapceConfig::resolve_theme()` in `lapce-app/src/config.rs` to handle `core.color_theme == "auto"`. When in auto mode:

1. Call `system_theme::current_system_theme()` to determine OS preference
2. Map `SystemTheme::Dark` → `"Lapce Dark"`, `SystemTheme::Light` → `"Lapce Light"`
3. Resolve the mapped theme from `available_color_themes` as normal

Also:
- Add `"auto"` to the color theme list in the settings UI (`settings.rs` / `palette.rs`)
- Ensure the theme picker shows `"auto"` as a selectable entry
- When the user selects `"auto"` from the picker, `SetColorTheme` saves `"auto"` to `settings.toml`

## Acceptance criteria

- [ ] `LapceConfig::resolve_theme()` dispatches to the correct theme when `color_theme = "auto"`
- [ ] `"auto"` appears in the settings theme dropdown
- [ ] `"auto"` appears in the command palette theme picker
- [ ] Selecting `"auto"` saves the value to `settings.toml`
- [ ] Unit tests verify auto resolution with mocked system theme
- [ ] Non-auto themes continue to work exactly as before

## Blocked by

- Blocked by #3904

Issue #3904, opened 2026-05-23
System theme detection module


shivamnarkar47:
## Parent

#3903

## What to build

Create a new module `lapce-app/src/system_theme.rs` that detects the OS-level color scheme preference (light/dark) and can watch for runtime changes.

The module exposes:
- `SystemTheme` enum (`Light`, `Dark`)
- `current_system_theme() -> SystemTheme` — synchronous query of current OS preference
- `watch_system_theme(callback: Box<dyn Fn(SystemTheme) + Send + 'static>)` — register a callback fired on OS appearance changes

Platform implementations:
- **macOS**: Query `NSApp.effectiveAppearance` (via `objc2` or floem interop). Subscribe to appearance changes via distributed notifications or KVO.
- **Windows**: Read `HKCU\\Software\\Microsoft\\Windows\\CurrentVersion\\Themes\\Personalize\\AppsUseLightTheme` (0 = dark). Watch for changes via registry notification or polling.
- **Linux**: Support GNOME (`gsettings org.gnome.desktop.interface color-scheme` via DBus or subprocess) and KDE (`kreadconfig5` or `kconfig`). Use polling with `inotify` or a simple timer for watching.
- **Fallback**: Return `Dark` if detection fails.

The module should be designed with a trait/abstraction so that unit tests can inject a mock system theme result.

## Acceptance criteria

- [ ] `system_theme::current_system_theme()` returns the correct value on each platform
- [ ] `system_theme::watch_system_theme()` fires callbacks when the OS appearance changes
- [ ] A mock/trait abstraction exists for testing
- [ ] Unit tests verify the abstraction layer works
- [ ] Compiles cleanly on all three platforms (no cfg-gated compilation errors)

## Blocked by

None - can start immediately

Issue #3903, opened 2026-05-23
PRD: Follow system theme (auto mode)


shivamnarkar47:
## Problem Statement

Lapce users who follow their OS-level light/dark mode switching (e.g., automatic timer on macOS/Linux/Windows) currently have to manually switch their editor theme whenever the system changes. There is no way to tell Lapce "just follow whatever the OS says."

## Solution

Add an `"auto"` theme mode. When the user sets `color-theme = "auto"` in their settings (or selects it from the theme picker), Lapce will:
1. Detect the current OS-level color scheme preference (light or dark)
2. Automatically resolve to the corresponding built-in theme ("Lapce Light" or "Lapce Dark")
3. Watch for OS-level appearance changes at runtime and switch themes reactively

## User Stories

1. As a Lapce user, I want to set my theme to `"auto"` so that Lapce automatically matches my OS light/dark preference.
2. As a Lapce user, I want the theme to switch live when my OS changes from light to dark (or vice versa) without restarting Lapce.
3. As a Linux/GNOME user, I want Lapce to detect my system color scheme preference via `gsettings` / DBus.
4. As a Linux/KDE user, I want Lapce to detect my system color scheme preference via `kconfig` / DBus.
5. As a macOS user, I want Lapce to detect my system appearance via `NSApp.effectiveAppearance`.
6. As a Windows user, I want Lapce to detect my system theme via the `AppsUseLightTheme` registry key.
7. As a developer, I want the auto-mode logic to be testable in isolation from the UI framework.

### Config model

- A theme name of `"auto"` is treated as a reserved keyword that triggers follow-system mode.
- No new config field needed — the existing `core.color_theme` string is overloaded: `"auto"` is handled specially, everything else works as before.
- When `core.color_theme == "auto"`, the effective theme is determined at runtime by querying the OS.
- The mapping is: system dark → "Lapce Dark", system light → "Lapce Light". In the future, users could configure which themes map to each mode.

### Platform detection module

A new module `system_theme` (in `lapce-app/src`) with a clean platform-abstracted API:

```rust
pub enum SystemTheme {
    Light,
    Dark,
}

pub fn current_system_theme() -> SystemTheme;
pub fn watch_system_theme(callback: Box<dyn Fn(SystemTheme) + Send + 'static>);
```

Platform implementations:
- **macOS**: Use `floem`'s existing platform integration or raw `objc2` calls to query `NSApp.effectiveAppearance` and subscribe to appearance changes via distributed notifications or KVO.
[...]

PR #3907, opened 2026-05-23
feat: implement follow system theme with Wayland support

### LINKED ISSUE #1324: Feature request: match system theme
It would be nice if Lapce had an option match the system theme without needing to change it manually. E.g. Lapce light when in light mode and Lapce dark when in dark mode.

It would probably be a good idea to use a third party crate for picking up the system theme. This would ease the strain on making this work cross-platform. In addition, because Lapce allows for third party themes via the use of plugins, it is important to allow the user to choose which theme is selected. A drop down for which theme is used when in light mode and which in dark mode should be straight forward enough. Plugins can even offer to configure this automatically. 

This behaviour could be implemented through a plugin but I don't believe that's necessary.

### LINKED ISSUE #3900: Feature request: Follow system theme

I like a uniform system light/dark mode, and I have an automatic timer set to change the theme. It would be nice if lapce had an option to follow the system theme instead of either light or dark

shivamnarkar47:
## Summary

Adds `follow-system-theme` setting that automatically switches Lapce's color theme based on OS dark/light mode preference. Users can configure which theme to use for each mode via `color-theme-dark` and `color-theme-light`.

### New settings (CoreConfig)
- `follow_system_theme` (bool, default `false`) — enables system theme following
- `color_theme_dark` (string, default `"Lapce Dark"`) — theme for dark mode
- `color_theme_light` (string, default `"Lapce Light"`) — theme for light mode

### Theme detection
- **X11/macOS/Windows**: listens for `Event::ThemeChanged` from winit (via floem)
- **Wayland (Linux)**: polls GTK settings every 5s (`gsettings` or `~/.config/gtk-{3,4}.0/settings.ini`)
- Initial detection at startup via Linux GTK config parsing
- Atomic static `SYSTEM_THEME` stores detected state; all `LapceConfig::load()` calls automatically apply override

### Config plumbing
- `override_color_theme()` — in-memory theme switch without persisting to settings file
- Modified `LapceConfig::load()` to apply system theme override after normal resolution
- `AppData::handle_system_theme_change()` — handles theme change events
- `AppData::os_theme` signal tracks current OS theme

## Related Issues

Closes #3900 — Feature request: Follow system theme
Closes #1324 — Feature request: match system theme (duplicate)
Refs #3903 — PRD: Follow system theme (auto mode)
Refs #3904 — System theme detection module
Refs #3905 — Auto-mode theme resolution + 'auto' in theme picker
Refs #3906 — Runtime system theme watcher in app.rs
Related: #502 — Window border is white in dark mode on Fedora 36 (Wayland)
Related: #2386 — titlebar color not changing
Related: #3626 — Font colors in Lapce Light mode

## Testing
1. Set `follow-system-theme = true` in settings, optionally `color-theme-dark`/`color-theme-light`
2. Change system dark/light mode — Lapce follows
3. Tested on Wayland via GTK config polling
4. Works alongside existing theme picker (follow mode overrides in-memory)

### DIFF
--- a/defaults/settings.toml
+++ b/defaults/settings.toml
@@ -7,6 +7,9 @@ icon-theme = "Lapce Codicons"
 custom-titlebar = true
 file-explorer-double-click = false
 auto-reload-plugin = false
+follow-system-theme = false
+color-theme-dark = "Lapce Dark"
+color-theme-light = "Lapce Light"
 
 [editor]
 font-family = "monospace"
--- a/extra/schemas/settings.json
+++ b/extra/schemas/settings.json
@@ -176,6 +176,18 @@
                 },
                 "custom-titlebar": {
                     "type": "boolean"
+                },
+                "follow-system-theme": {
+                    "type": "boolean",
+                    "description": "Automatically switch color theme based on system appearance"
+                },
+                "color-theme-dark": {
+                    "type": "string",
+                    "description": "Color theme to use in dark mode (when follow_system_theme is enabled)"
+                },
+                "color-theme-light": {
+                    "type": "string",
+                    "description": "Color theme to use in light mode (when follow_system_theme is enabled)"
                 }
             },
             "required": [],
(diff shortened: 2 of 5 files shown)


shivamnarkar47:
Btw I do read #1324  , but based on my current experience, I think we do need auto theme, and I have added a custom theme addition option too, so that a user can put his preference light and dark theme on it.

Screenshot attached : 
image

